PR #1, opened 2018-10-25
Added user login and new user login pages

drewlearnsabout:
Hi all, made some simple pages that could be used for the login page and new user login page.  Let me know your thoughts!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,44 @@
+<html>
+    <head>
+            <link rel="stylesheet" href="style.css">
+    </head>
+    <body>
+        <main id="main">
+            <form id="new_user_form" action="/submit-info-here">
+                <section id="personal_information">
+                    <p id="personal_information_label">New User Login</p>
+                    <section id="first_name_section">    
+                        <label id="first-name-label" class="labels">First Name</label>
+                        <input id="first-name" type="text" placeholder="First Name" required>
+                    </section>
+                    <section id="last_name_section">                    
+                        <label id="last-name-label" class="labels">Last Name</label>
+                        <input id="last-name" type="text" placeholder="Last Name" required>
+                    </section>
+                    <section id="email_section">
+                        <label id="email-label" class="labels">Email</label>
+                        <input id="email" type="email" placeholder="Email" required>
+                    </section>
+                    <section id="admin_permissions_section">            
+                        <label id="dropdown-label">Admin permissions</label>
+                        <select id="dropdown" name="admin_permissions" class="dropdown">
+                            <option disabled value>Please select an option</option>
+                            <option value="yes">Yes</option>
+                            <option value="no">No</option>
+                            <option value="Default to no">Default to no</option>
+                        </select>
+                    </section>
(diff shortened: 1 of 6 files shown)


index.html
```diff
@@ -0,0 +1,44 @@
...
+                        <label id="email-label" class="labels">Email</label>
+                        <input id="email" type="email" placeholder="Email" required>
+                    </section>
+                    <section id="admin_permissions_section">            
+                        <label id="dropdown-label">Admin permissions</label>
+                        <select id="dropdown" name="admin_permissions" class="dropdown">
+                            <option disabled value>Please select an option</option>
+                            <option value="yes">Yes</option>
+                            <option value="no">No</option>
+                            <option value="Default to no">Default to no</option>
```
NickSchimek: Default to no should not be an option. When the form displays the "no" value should be the default value.

NickSchimek:
Great work! I like it. There is just one change that needs to be done for now.

We'll need to hold off on merging this until we get the front end structure built. Then we'll figure out where this needs to go, what name to give the files, etc...

The indentation seems off in a couple of places, but that's not really a big deal. I will try to get a style robot added so that it can automatically inform us of these little issues so that they're easily caught and fixed.

oh yeah, it seems there are duplicate files `index.html` & `index_new_user.html` is there a reason for both of files?

index_new_user.html
```diff
@@ -0,0 +1,44 @@
...
+                        <label id="email-label" class="labels">Email</label>
+                        <input id="email" type="email" placeholder="Email" required>
+                    </section>
+                    <section id="admin_permissions_section">            
+                        <label id="dropdown-label">Admin permissions</label>
+                        <select id="dropdown" name="admin_permissions" class="dropdown">
+                            <option disabled value>Please select an option</option>
+                            <option value="yes">Yes</option>
+                            <option value="no">No</option>
+                            <option value="Default to no">Default to no</option>
```
NickSchimek: Default to no should not be an option. When the form displays the "no" value should be the default value.

index.html
```diff
@@ -0,0 +1,44 @@
...
+            <link rel="stylesheet" href="style.css">
+    </head>
+    <body>
+        <main id="main">
+            <form id="new_user_form" action="/submit-info-here">
+                <section id="personal_information">
+                    <p id="personal_information_label">New User Login</p>
+                    <section id="first_name_section">    
+                        <label id="first-name-label" class="labels">First Name</label>
+                        <input id="first-name" type="text" placeholder="First Name" required>
```
NickSchimek: These validations are great work. Good job!

index.html
drewlearnsabout: Hey Nick, sorry about the super late response, I'll work on it this week!

NickSchimek: No worries! That is perfect

NickSchimek:
Thanks so much for this PR @drewlearnsabout - I'm looking forward to your react PR's. I'm going to go ahead and close this since @hmarcks has a design for us to convert to react.

PR #3, opened 2018-10-31
src/expunge: add core objects for expunge analysis

a-ilango:
I've added some classes that would be used for expunge analysis.

### DIFF
--- a/src/backend/expungeservice/__init__.py
+++ b/src/backend/expungeservice/__init__.py
@@ -1,3 +1,4 @@
 from . import app
 from .app import create_app
 from . import endpoints
+from . import expunger
--- a/src/backend/expungeservice/expunger/__init__.py
+++ b/src/backend/expungeservice/expunger/__init__.py
@@ -1,0 +1,1 @@
+from .analyze import *
--- a/src/backend/expungeservice/expunger/analyze.py
+++ b/src/backend/expungeservice/expunger/analyze.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python3
+
+import collections
+import enum
+
+class Level(object):
+    """ Crime Level.
+
+    Describes crime level. e.g. Felony Class A.
+
+    Attributes:
+        type_: A string describing the type of crime.
+        class_: A string of length 1 specifying the class.
+    """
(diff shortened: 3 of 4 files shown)


src/expunge/expunge.py
```diff
@@ -0,0 +1,131 @@
...
+        fine: A float that tells how much money is owed to the court.
+    """
+    def __init__(
+            self,
+            name,
+            statutes,
+            level,
+            date,
+            disposition,
+            fine=0):
```
NickSchimek: I think the fine should be moved to the `Case` class and call it `balance_due` ? The disposition page has an easy to parse balance due and that is what the client is most interested in. It will be difficult to extract the fine per charge as its not clearly stated. If there are payments to be made for the charges it's grouped together for each case at the bottom of the disposition page in the financial section.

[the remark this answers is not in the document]
a-ilango: I see. I remember Michael telling tracking money owed is needed when analyzing but I don't see it in the chart that he sent. I can move it to `Case` if tracking it per charge is not needed.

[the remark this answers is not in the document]
a-ilango: That makes sense. I'll update the code. Thanks!

a-ilango:
I've added the updates in separate commits so that it's easy to see what changed. I'll squash them before merge.

NickSchimek:
Thanks, I will review these changes soon.

src/expunge/expunge.py
```diff
@@ -0,0 +1,144 @@
...
+        name: A string that specifies client's name.
+        dob: A datetime.date that specifies date of birth.
+        cases: A list of Case(s).
+    """
+    def __init__(self, name, dob, cases):
+        self.name = name
+        self.dob = dob
+        self.cases = cases
+
+# Tests
```
NickSchimek: Can we update these to automated tests and put them in a test directory?

a-ilango: Sure. We need to merge #5 soon so that the directory structure is set.

a-ilango: Added tests to test directory.

PR #4, opened 2018-11-03
Working minimal Flask application

erikjasso:
Pushed a working `Pipenv` setup and minimal `Flask` application accessible at http://localhost:5000/hello.

See `README.md` for instructions to run the application.

My changes caused a merge conflict; let me know if you need help resolving it.


a-ilango:
> My changes caused a merge conflict; let me know if you need help resolving it.

@erikjasso you would have to sync with upstream to resolve the conflict.

doc/design.md
```diff
@@ -0,0 +1,229 @@
...
+**`POST`** `/api/users/`
+
+Creates a user
+
+`POST` body:
+
+- format: `JSON`
+- fields:
+    * username
+	* encrypted password
```
a-ilango: Should this still be encrypted as the session is over HTTPS?

doc/design.md
```diff
@@ -0,0 +1,229 @@
...
+
+    users (uuid, username, email, password, password_salt, admin), uuid primary key
+
+    expunged_records (uuid, first_name, last_name, dob), uuid primary key
+
+Other tables TBD
+
+Notes:
+
+- Store passwords encrypted
```
a-ilango: Isn't it recommended to store only the salted passwords and not even encrypted ones?

src/models/__init__.py
```diff
@@ -1,11 +0,0 @@
...
-
-# initialize our db
-db = SQLAlchemy()
-
-
-from flask_bcrypt import Bcrypt
-#######
-# existing code remains #
-#######
-bcrypt = Bcrypt()
```
a-ilango: What's the second recordexpunge directory for?

doc/design.md
NickSchimek: I suggest for security reasons we use [bcrypt](https://github.com/pyca/bcrypt/) to hash the passwords, and then store the hashed passwords.

erikjasso: Yes, the table will store the salt and the hashed password.

doc/design.md
erikjasso: You're right, the password sent from the application to the server will not be encrypted. It will be checked against the hashed password stored in the database.

doc/design.md
erikjasso: When a password is sent to the server it will be compared against the hashed password.

src/models/__init__.py
erikjasso: I removed the redudant dir in `src/backend/` to leave`src/backend/expungeservice/` as introduced in commit 234a6c0and.

erikjasso:
Closing in favor of newer PR w/o conflicts.

PR #14, opened 2018-11-15
Create record parser

### LINKED ISSUE #6: Create a record parser
Build a parser that collects all the data from the  search results page on OJCIN that builds a list of cases. Each row in the results page would be a case.

NickSchimek:
Created the record parser
This parses the search results (a record) and creates a list of cases from the record.

Fixes #6


src/backend/tests/parser/test_case_parser.py
```diff
@@ -0,0 +1,92 @@
...
+        assert self.parser.cases[1].charges[0] == 'Poss Controlled Sub 2 (Reduced - to A Misdemeanor)'
+
+
+
+class TestEmptyRecord(unittest.TestCase):
+
+    def test_empty_record(self):
+        """Tests it can parse a blank record."""
+        parser = CaseParser()
+        parser.cases = []
```
NickSchimek: whoops how did this get in there. That can be removed.

src/backend/expungeservice/parsers/case_parser.py
```diff
@@ -0,0 +1,119 @@
...
+                1: self.__column_1,
+                2: self.__column_2,
+                3: self.__column_3,
+                4: self.__column_4,
+                5: self.__column_5,
+            }
+            switcher.get(self.column, self.__column_6_plus)(data)
+
+        elif 'Charge(s)' == data:
+            self.collect_data = True
```
a-ilango: It seems this should always be true once we finish parsing the header (<th>) which would occur once per page. So can we just start collecting data for the table rows?

NickSchimek: yeah that's what its doing. There are five `<th>` tags in the document. `Charge(s)` is the very last `<th>` tag data. Once it passes that point it starts collecting data for the rest of the page.

src/backend/expungeservice/parsers/case_parser.py
```diff
@@ -0,0 +1,119 @@
...
+            self.__reset_flags()
+
+    def handle_data(self, data):
+        if self.__within_valid_table_row():
+            switcher = {
+                1: self.__column_1,
+                2: self.__column_2,
+                3: self.__column_3,
+                4: self.__column_4,
+                5: self.__column_5,
```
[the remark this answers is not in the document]
NickSchimek: Yes, good suggestion. I will update.

src/backend/tests/parser/test_case_parser.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class TestCaseParser(unittest.TestCase):
+
+    def setUp(self):
+        self.parser = CaseParser()
+        self.parser.feed(JohnDoe.RECORD)
+        self.case1_date = datetime.date(datetime.strptime('03, 23, 1963', '%m, %d, %Y'))
+        self.case2_date = datetime.date(datetime.strptime('04, 11, 1963', '%m, %d, %Y'))
+        self.case3_date = datetime.date(datetime.strptime('04, 01, 2012', '%m, %d, %Y'))
```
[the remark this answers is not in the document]
NickSchimek: It would probably be better to do something like this instead.
`datetime.date(1963, 3, 23)`

src/backend/expungeservice/parsers/case_parser.py
```diff
@@ -0,0 +1,119 @@
...
+        self.citation_number = []  # column 2
+        self.info = []             # column 3
+        self.date_location = []    # column 4
+        self.type_status = []      # column 5
+        self.charges = []          # column 6+
+        self.case_detail_link = ''
+
+    def handle_starttag(self, tag, attrs):
+        self.__increase_column_count(tag)
+        self.__assign_case_link(tag, attrs)
```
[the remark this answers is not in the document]
NickSchimek: Sorry I think I see what your saying. There are no other anchor tags once data collection is turned on, but I think it does make sense to add the extra conditional.

src/backend/expungeservice/parsers/case_parser.py
NickSchimek: There are actually four separate tables on the page. The table we collect data from is the last table, hence the reason to check for `Charge(s)` table data. Once we pass that point data the data collection is turned on.

a-ilango: Ah, I see. I assumed we had only one table on the page. Thanks for the clarification!

src/backend/expungeservice/parsers/case_parser.py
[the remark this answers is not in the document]
a-ilango: Checking for the anchor tag would be good. It would be nice if that check is moved out of the function to here, so it clarifies when we assign case link.

src/backend/expungeservice/parsers/case_parser.py
```diff
@@ -22,7 +22,8 @@ def __init__(self):
 
     def handle_starttag(self, tag, attrs):
         self.__increase_column_count(tag)
-        self.__assign_case_link(tag, attrs)
+        if self.__case_column():
+            self.__assign_case_link(tag, attrs)
```
a-ilango: I think this should work as well

src/backend/tests/parser/test_case_parser.py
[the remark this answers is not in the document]
a-ilango: Got it. Thanks!

NickSchimek:
Thanks @a-ilango I see the importance for doing the small commits for the code review process. I should've taken the time to do that. Next time I will. Was there anything else that needs to be addressed or can this be merged?

a-ilango:
Smaller commits, PRs are easier to review but this wasn't long. The PR looks fine and I think we can merge it.

a-ilango:
Also please squash together the trivial commits.

PR #18, opened 2018-11-24
Create params parser for search page

### LINKED ISSUE #17: Create parser to pull valid params from search page
There are a few params that are sent back in the response when visiting navigating to the search. Need to retrieve those params in order to perform a search.
child of #11

NickSchimek:
This collects the dynamic data when navigating to the search page in order to validate a session.

Fixes #17

### DIFF
--- a/src/backend/expungeservice/parsers/param_parser.py
+++ b/src/backend/expungeservice/parsers/param_parser.py
@@ -1,0 +1,19 @@
+from html.parser import HTMLParser
+
+
+class ParamParser(HTMLParser):
+
+    def __init__(self):
+        HTMLParser.__init__(self)
+
+    def handle_starttag(self, tag, attrs):
+        pass
+
+    def handle_endtag(self, tag):
+        pass
+
+    def handle_data(self, data):
+        pass
+
+    def error(self, message):
+        pass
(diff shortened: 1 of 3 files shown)


src/backend/expungeservice/parsers/param_parser.py
```diff
@@ -0,0 +1,51 @@
...
+        if tag == 'input':
+            switcher = {
+                '__EVENTTARGET': self.__event_target,
+                '__EVENTARGUMENT': self.__event_argument,
+                '__VIEWSTATE': self.__view_state,
+                '__VIEWSTATEGENERATOR': self.__view_state_generator,
+                '__EVENTVALIDATION': self.__event_validation,
+                'NodeID': self.__node_id
+            }
+            switcher.get(dict(attrs)['name'], self.__default)(dict(attrs).get('value'))
```
a-ilango: What will these parameters be used for? It would be nice to add some documentation.

NickSchimek: Good point, I should definitely add some documentation. They're actually explained what they're for in issue #17 and a little bit in the git commit. I will update with documentation.

a-ilango: Sorry, I didn't see the commit message. So these parameters needed when sending the search query?

NickSchimek: Exactly. Need to navigate to the page and grab these parameters in order to post a valid search request. It seems ASP.NET does that to verify the request originates from controls rendered by their server.

Issue #10, opened 2018-11-10
Create database objects


NickSchimek:

NickSchimek:
This was completed in PR #19

PR #19, opened 2018-11-26
Db setup

erikjasso:
This PR has the database schema we talked about in `config/postgres/initdb/scripts/create-tables.sql` and runs the Postgres database as a service in a Docker container. Container is managed by targets in `Makefile`. Instructions for installing Docker on Mac are in `doc/docker_mac.md` (which I just realized should be linked from the main README but isn't). Linux instructions are similar (the difference is installing the Docker engine, which I think is in apt). Once Docker is installed on either platform all interaction is via Make.

I plan to put our Flask application and Nginx web server in containers, too.

This will make setting up the dev environment easier (e.g., run the database container vs configuring and running Postgres manually) and will simplify deployment down the line.


config/postgres/initdb/scripts/create-tables.sql
```diff
@@ -0,0 +1,58 @@
...
+
+CREATE EXTENSION "uuid-ossp";
+CREATE EXTENSION citext;
+
+CREATE TABLE users (
+    user_id         UUID NOT NULL,
+    username        CITEXT UNIQUE NOT NULL,
+    email           CITEXT UNIQUE NOT NULL,
+    hashed_password TEXT NOT NULL,
+    password_salt   TEXT NOT NULL,
```
NickSchimek: If using `bcrypt` to hash the passwords the salt is stored in the hash. Are we planning to use a different hashing function?

config/postgres/initdb/scripts/create-tables.sql
```diff
@@ -0,0 +1,58 @@
...
+    statute        TEXT NOT NULL,
+    date_eligible  TIMESTAMP WITH TIME ZONE,
+    rule_ids       UUID[] NOT NULL,
+    date_created   TIMESTAMP WITH TIME ZONE DEFAULT now(),
+    date_modified  TIMESTAMP WITH TIME ZONE DEFAULT now(),
+    expunged       BOOLEAN,
+    date_expunged  TIMESTAMP WITH TIME ZONE,
+    user_id        UUID NOT NULL REFERENCES users(user_id) ON DELETE CASCADE,
+    PRIMARY KEY (client_id, case_id)
+);
```
a-ilango: Speaking with Michael it seems like we may need to create separate table for "time eligibility analysis" if it makes sense to do so. The info that would need to go on this would be
- if "10 year rule" passed - means all charges are expungeable
- if most recent conviction is eligible, etc.
- date of eligibility for charges other than MRC.
...

These would be listed on the UI page at the top. @hmarcks 

So we either need to add those info in this table or create a new one.

NickSchimek:
Fixes issue #10

erikjasso:
Fixes will follow to address issues brought up from discussions since this PR was opened.

PR #20, opened 2018-11-27
expunger: add record analyzer

a-ilango:
This PR adds the abstractions needed for running an analysis on client's records and a few checks. More checks would be added soon.

### DIFF
--- a/src/backend/expungeservice/expunger/analyze.py
+++ b/src/backend/expungeservice/expunger/analyze.py
@@ -103,6 +103,15 @@ class Charge(object):
         self.level = level
         self.date = date
         self.disposition = disposition
+        self._result = None
+
+    @property
+    def type_elig_result(self):
+        return self._result
+
+    @type_elig_result.setter
+    def type_elig_result(self, result):
+        self._result = result
 
 CaseState = enum.Enum('CaseState', 'OPEN CLOSED')
 
@@ -140,3 +149,159 @@ class Client(object):
         for case in self.cases:
             num += case.num_charges()
         return num
+
+class ResultCode(enum.Enum):
+    INELIGIBLE = 'Ineligible'
+    ELIGIBLE = 'Eligible'
+    FURTHER_ANALYSIS = 'Further analysis needed'
+    EXAMINE = 'Examine'
+    NO_ACTION = 'Take no action'
+    OPEN_CASE = 'Open Case'
+
+"""
+A class for storing individual analysis steps
+
+Attributes:
+    result: A boolean that's the result of the check
+    check: A string that breifly describes the check
+    check_desc: A string that elaborates the check (optional)
+"""
+CheckResult = collections.namedtuple('CheckResult', 'result check check_desc',
+                                     defaults=[None])
+
+class Result(object):
+    """
+    A class for storing analysis results
+
+    Attributes:
+        code: A ResultCode instance.
+        statute: A Statute instance that's associated with arrived result.
+        date: A datetime.date instance that specifies eligibility date.
+        analysis: A list of CheckResult instances that describes the analysis done.
+    """
+    def __init__(self, code=None, analysis=None,
+                 statute=None, date=None):
+        self.code = code
+        self.analysis = analysis
+        self.statute = statute
+        self.date = date
+
+    def __str__(self):
+        return ' '.join([str(self.code), str(self.analysis), str(self.statute), str(self.date)])
+
+class ResultInElig_137_225_5(Result):
+    def __init__(self, **kwargs):
+        Result.__init__(self, code=ResultCode.INELIGIBLE,
+                        statute=Statute(137, 225, 5), **kwargs)
+
+class RecordAnalyzer(object):
+    """
+    A class for analyzing client's records
+
+    Attributes:
+        client: A Client instance
+    """
+
+    def __init__(self, client):
+        self.client = client
+
+    def _is_charge_level(charge, type_, class_):
+        check = 'Is the charge a {}'.format(type_)
+        if class_:
+            check += ' class {}'.format(class_)
+
+        result = (charge.level.type_ ==  type_ and
+                  (not class_ or charge.level.class_ == class_))
+        return CheckResult(check=check, result=result)
+
+    def _is_charge_statute(charge, statute):
+        check = 'Does the charge fall under statute: ' + str(statute)
+        return CheckResult(check=check, result=charge.statute == statute)
+
+    def _is_charge_in_statute_list(charge, statutes, desc):
+        check = ('Does the charge fall under any of these statutes: ' +
+                 ','.join(str(statute) for statute in statutes))
+        # TODO implement this
+        return CheckResult(check=check, result=False, check_desc=desc)
+
+    def _is_charge_sex_crime(charge):
+        # TODO update
+        _statutes_sex_crimes = []
+        return RecordAnalyzer._is_charge_in_statute_list(
+            charge, _statutes_sex_crimes, 'Is the charge a sex crime')
+
+    def _is_charge_traffic_crime(charge):
+        # TODO update
+        _statutes_traffic_crimes = []
+        return RecordAnalyzer._is_charge_in_statute_list(
+            charge, _statutes_traffic_crimes, 'Is the charge a traffic crime')
+
+    def _have_open_case(self):
+        check = 'Is there a open case for the client'
+        result = any([case.state == CaseState.OPEN for case in self.client.cases])
+        return CheckResult(check=check, result=result)
+
+    """
+    Run Time Eligibility analysis on the client and their records
+
+    TODO: make it return analysis for each charge as well (which is supposed
+    to be an update on eligibility date and relevan statutes)
+
+    Returns:
+        An Result instance
+    """
+    def time_eligibility(self):
+        analysis = []
+
+        analysis.append(self._have_open_case())
+        if analysis[-1].result:
+            return Result(ResultCode.OPEN_CASE, analysis)
+
+        # TODO implement the rest
+
+        return Result(ResultCode.NO_ACTION)
+
+    """
+    Run Type Eligibility analysis on a charge
+
+    Args:
+        charge: A Charge instance
+
+    Returns:
+        A Result instance
+    """
+    def type_eligibility(self, charge):
+        analysis = []
+
+        analysis.append(RecordAnalyzer._is_charge_level(charge, 'Felony', 'A'))
+        if analysis[-1].result:
+            return ResultInElig_137_225_5(analysis=analysis)
+
+        analysis.append(RecordAnalyzer._is_charge_sex_crime(charge))
+        if analysis[-1].result:
+            return ResultInElig_137_225_5(analysis=analysis)
+
+        analysis.append(RecordAnalyzer._is_charge_traffic_crime(charge))
+        if analysis[-1].result:
+            return ResultInElig_137_225_5(analysis=analysis)
+
+        # TODO add remaining analysis
+
+        return Result(ResultCode.FURTHER_ANALYSIS, analysis)
+
+    """
+    Analyze which records are expungeable
+
+    The method sets the Charge instance's result attribute with type eligibility
+    analysis result.
+
+    Returns:
+        A tuple consisting of:
+            - A Result instance that describes the Time Eligibility analysis.
+    """
+    def analyze(self):
+        for case in self.client.cases:
+            for charge in case.charges:
+                charge.type_elig_result = self.type_eligibility(charge)
+
+        return self.time_eligibility()
--- a/src/backend/expungeservice/expunger/analyze.py
+++ b/src/backend/expungeservice/expunger/analyze.py
@@ -63,6 +63,14 @@ class Statute(object):
         self.subsection = subsection
         # TODO we may need to add components beyond subsection
 
+    def __eq__(self, other):
+        return (self.chapter == other.chapter and
+                self.subchapter == other.subchapter and
+                ((not self.section and not other.section) or
+                 self.section == other.section) and
+                ((not self.subsection and not other.subsection) or
+                 self.subsection == other.subsection))
+
     def __str__(self):
         # TODO do these need to have leading zeros?
         statute = '{}'.format(self.chapter)
--- a/src/backend/expungeservice/expunger/analyze.py
+++ b/src/backend/expungeservice/expunger/analyze.py
@@ -106,6 +106,9 @@ class Case(object):
         self.state = state
         self.balance_due = balance_due
 
+    def num_charges(self):
+        return len(self.charges)
+
 class Client(object):
     """ Client - An individual who wants to expunge charges from their record.
 
@@ -118,3 +121,9 @@ class Client(object):
         self.name = name
         self.dob = dob
         self.cases = cases
+
+    def num_charges(self):
+        num = 0
+        for case in self.cases:
+            num += case.num_charges()
+        return num
--- a/src/backend/expungeservice/expunger/analyze.py
+++ b/src/backend/expungeservice/expunger/analyze.py
@@ -3,7 +3,8 @@
 import collections
 import enum
 
-class Level(object):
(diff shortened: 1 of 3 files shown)


src/backend/expungeservice/expunger/analyze.py
```diff
@@ -118,3 +118,182 @@ def __init__(self, name, dob, cases):
...
+    """
+    def analyze(self):
+        type_elig_analysis_list = []
+        for case in self.client.cases:
+            for charge in case.charges:
+                result = self.type_eligibility(charge)
+                type_elig_analysis_list.append(ChargeAnalysis(charge, case, result))
+
+        time_elig_analysis = self.time_eligibility()
+        return type_elig_analysis_list, time_elig_analysis
```
NickSchimek: It may be better to add the expunge-ability of a charge onto the charge class itself as an attribute. This will help with presenting the information to the user. And then when we test we can do something like `charge.result.code`

src/backend/expungeservice/expunger/analyze.py
```diff
@@ -118,3 +118,182 @@ def __init__(self, name, dob, cases):
...
+        if checks[-1].result:
+            return ResultInElig137(checks=checks)
+
+        checks.append(RecordAnalyzer._is_charge_traffic_crime(charge))
+        if checks[-1].result:
+            return ResultInElig137(checks=checks)
+
+        # TODO add remaining checks
+
+        return Result(ResultCode.FURTHER_ANALYSIS, checks)
```
NickSchimek: Let's discuss this method next time we meet so that I can get a better understanding of the approach being taken.

src/backend/expungeservice/expunger/analyze.py
a-ilango: I thought about that and went back and forth between modifying the charge object itself vs returning a new object (ChargeAnalysis) that has the result and reference to that charge. It seemed the latter is cleaner as it doesn't have side effects. What do you think?

src/backend/expungeservice/expunger/analyze.py
a-ilango: Sure but it would be in about a month. This function basically runs through different checks and appends them to a list which would be returned as part of the type eligibility analysis.

src/backend/expungeservice/expunger/analyze.py
NickSchimek: I see. I'm not against side effects as some side effects are good. I think the UI will ultimately decide on how we proceed here and if this data ever plans to go in a database will also be a factor.

a-ilango:
@NickSchimek updated the PR based on the discussion we had. The new changes are mainly modifying the Charge object directly with the type eligibility analysis result. Also fixed a few issues in existing objects.

src/backend/expungeservice/test/test_expunger.py
```diff
@@ -20,16 +20,51 @@ def test_statute():
 
 def get_disp():
     date = datetime.date(1996, 1, 1)
-    disp_type = DispositionType.NO_CONVICTION
+    disp_type = DispositionType.CONVICTED
```
NickSchimek: One thing to do would be to name the functions better. get_disp() doesn't clearly state what it's doing which is creating a Ten plus year convicted disposition. If arguments are passed to the function to create the dispo than it makes sense to have a broader name.

src/backend/expungeservice/test/test_expunger.py
```diff
@@ -20,16 +20,51 @@ def test_statute():
...
+    ]
+    cases = [Case(charges, CaseState.OPEN, 100.50)]
     client = Client('John Doe', datetime.date(1970, 1, 1), cases)
+    record_analyzer = RecordAnalyzer(client)
+    time_elig_result = record_analyzer.analyze()
+
+    # TODO add individual test cases that test different paths in the decision tree
+    assert(charges[0].type_elig_result.code == ResultCode.FURTHER_ANALYSIS)
+    assert(charges[1].type_elig_result.code == ResultCode.INELIGIBLE)
+    assert(time_elig_result.code == ResultCode.OPEN_CASE)
```
NickSchimek: If the case is open it shouldn't stop us from analyzing the expungeability of a charge if the charge has a disposition associated to it.

src/backend/expungeservice/expunger/analyze.py
```diff
@@ -118,3 +134,166 @@ def __init__(self, name, dob, cases):
...
+        statute: A Statute instance that's associated with arrived result.
+        date: A datetime.date instance that specifies eligibility date.
+        checks: A list of Check instances that describes the analysis done.
+    """
+    def __init__(self, code=None, checks=None,
+                 statute=None, date=None):
+        self.code = code
+        self.checks = checks
+        self.statute = statute
+        self.date = date
```
NickSchimek: Would this not be the same for all charges except for the most recent?

src/backend/expungeservice/expunger/analyze.py
```diff
@@ -118,3 +134,166 @@ def __init__(self, name, dob, cases):
...
+        code: A ResultCode instance.
+        statute: A Statute instance that's associated with arrived result.
+        date: A datetime.date instance that specifies eligibility date.
+        checks: A list of Check instances that describes the analysis done.
+    """
+    def __init__(self, code=None, checks=None,
+                 statute=None, date=None):
+        self.code = code
+        self.checks = checks
+        self.statute = statute
```
NickSchimek: What is a result statute? Would the result statute not be the same as a charge statute?

src/backend/expungeservice/expunger/analyze.py
```diff
@@ -118,3 +134,166 @@ def __init__(self, name, dob, cases):
...
+    Attributes:
+        code: A ResultCode instance.
+        statute: A Statute instance that's associated with arrived result.
+        date: A datetime.date instance that specifies eligibility date.
+        checks: A list of Check instances that describes the analysis done.
+    """
+    def __init__(self, code=None, checks=None,
+                 statute=None, date=None):
+        self.code = code
+        self.checks = checks
```
NickSchimek: Can you add some tests for these checks? If not we can remove this and add it as a feature to do be done at a later time.

src/backend/expungeservice/expunger/analyze.py
```diff
@@ -91,6 +96,14 @@ def __init__(
         self.date = date
         self.disposition = disposition
 
+    @property
+    def type_elig_result(self):
+        return self._result
+
+    @type_elig_result.setter
+    def type_elig_result(self, result):
+        self._result = result
```
NickSchimek: This should probably be instantiated within the constructor, even if it's just being instantiated to None.

src/backend/expungeservice/expunger/analyze.py
```diff
@@ -118,3 +134,166 @@ def __init__(self, name, dob, cases):
...
+    analysis result.
+
+    Returns:
+        A tuple consisting of:
+            - A Result instance that describes the Time Eligibility analyis.
+    """
+    def analyze(self):
+        for case in self.client.cases:
+            for charge in case.charges:
+                charge.type_elig_result = self.type_eligibility(charge)
```
NickSchimek: Do you think it would be better to pass one charge to the expunger and evaluate that individually?, and then have a separate analyzer if necessary (dependent on whether or not we can use a db) to check if there are any open cases and the time eligibility for the group of charges?

src/backend/expungeservice/test/test_expunger.py
```diff
@@ -20,16 +20,51 @@ def test_statute():
...
+        get_unqualifying_charge(),
+    ]
+    cases = [Case(charges, CaseState.OPEN, 100.50)]
     client = Client('John Doe', datetime.date(1970, 1, 1), cases)
+    record_analyzer = RecordAnalyzer(client)
+    time_elig_result = record_analyzer.analyze()
+
+    # TODO add individual test cases that test different paths in the decision tree
+    assert(charges[0].type_elig_result.code == ResultCode.FURTHER_ANALYSIS)
+    assert(charges[1].type_elig_result.code == ResultCode.INELIGIBLE)
```
NickSchimek: This is testing the crime level and not the statutue? I thought it was the statute that is used to determine expungeability?

src/backend/expungeservice/expunger/analyze.py
```diff
@@ -118,3 +134,166 @@ def __init__(self, name, dob, cases):
...
+    Args:
+        charge: A Charge instance
+
+    Returns:
+        A Result instance
+    """
+    def type_eligibility(self, charge):
+        checks = []
+
+        checks.append(RecordAnalyzer._is_charge_level(charge, 'Felony', 'A'))
```
NickSchimek: Do you think there is a better way to implement this then to go down a list and test for each charge in a given order?

src/backend/expungeservice/test/test_expunger.py
NickSchimek: I see you have it to check the level and then statutes... Does it only matter if the level is a felony class A? It seems statutes may have different levels, is that a correct assumption?

NickSchimek:
Thanks for the PR @a-ilango. I see there are a lot of functions that need to be implemented, and we should probably make testing a priority as we're implementing. The requirements for the UI is to know if a charge can be expunged regardless of date rules so that it can be grouped into lists of charges that are eligible now, charges that are eligible in the future, and charges that are simply not eligible. Am I missing anything @hmarcks ? The main thing is to determine if a charge is eligible, all other features are nice to haves, and can be implemented later. Do you think it would be better to allow the frontend to analyze the time eligibility portion of the analysis? Then this analyzer can focus on analyzing a single charge?

src/backend/expungeservice/expunger/analyze.py
a-ilango: Do you mean the individual checks? If so, yes it would be nice to have tests that exercise all the paths in the decision tree. I can add tests for the ones that can be added easily and defer other checks.

src/backend/expungeservice/expunger/analyze.py
a-ilango: Hmm. I'm not sure about that. Let me check with Michael.

src/backend/expungeservice/expunger/analyze.py
a-ilango: I believe it may be different for different charges. There's a path in type eligibility flow chart that indicates that the charge would be eligible in one year for no-complaint charges while a similar path in time eligibility chart applies only to most recent charge.

src/backend/expungeservice/expunger/analyze.py
a-ilango: I wonder if there's a bit of a misunderstanding here. To clarify, checks list is actually results of the checks. I should probably rename it. So we check a thing and add the result to a list which would be returned to the caller.

Do you still have the same question?

src/backend/expungeservice/expunger/analyze.py
a-ilango: We can do that but I'm not sure if that's better. The analyze function assumes the caller would want both type and time-eligibility analysis for a client.

> have a separate analyzer if necessary (dependent on whether or not we can use a db)

How does that relate to the db?

src/backend/expungeservice/test/test_expunger.py
```diff
@@ -20,16 +20,51 @@ def test_statute():
...
+    level = CrimeLevel('Infraction', 'A')
+    return Charge('Qualifiying Charge', statute, level,
+                  datetime.date(1995, 1, 1), disp)
+
+def get_unqualifying_charge():
+    # Fix me with correct charge
+    disp = get_disp()
+    statute = Statute(113, 45, 5, 'd')
+    level = CrimeLevel('Felony', 'A')
+    return Charge('Unqualifying Charge', statute, level,
```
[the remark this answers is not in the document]
a-ilango: The Statute specified here is just a dummy and doesn't correspond to any real ones. It's there just to construct the charge to test a particular check in the analyzer.

src/backend/expungeservice/test/test_expunger.py
a-ilango: Good point. Will update.

a-ilango:
> I see there are a lot of functions that need to be implemented, and we should probably make testing a priority as we're implementing

Yes, having a thoroughly tested analyzer would be good to have.

> Do you think it would be better to allow the frontend to analyze the time eligibility portion of the analysis?

It would be good to have all the analysis in one place. So I'd prefer it being here unless there's a strong reason to move it to front-end.

src/backend/expungeservice/test/test_expunger.py
a-ilango: Yes. This is just testing if the time eligibility analyzer returns the result of "Open Case" instead of running its analysis (as per the flowchart). We need to add other tests to test expunge-ability of a single charge.

src/backend/expungeservice/test/test_expunger.py
a-ilango: > Does it only matter if the level is a felony class A? 

Yes in this context. The tests are just following the checks in the flowchart.

> It seems statutes may have different levels, is that a correct assumption?

I'm not sure about that. We'd have to check with Michael.

hmarcks:
For each record, we want to show: 

**Time-eligibility** - tricky because it has to consider all the records to know whether an individual record is eligible.

**Type-eligibility** - only has to consider it's own record.

**Combined eligibility** - if the record is ultimately eligible based on Time + Type.

Here's is a mockup with more details about the UI data, let me know if you have any questions or ideas (it's not final, but I think it's getting pretty solid):

record expunge search ui data

src/backend/expungeservice/expunger/analyze.py
NickSchimek: We could ask the db to return the most recent charge and decide time elig on that or is it more involved than just looking at the most recent charge?

src/backend/expungeservice/test/test_expunger.py
NickSchimek: @michaelatcommunitylaw Can there be different level's associated to a charge?

src/backend/expungeservice/expunger/analyze.py
a-ilango: So @michaelatcommunitylaw says the "expunge criteria statute" would be the same as the charge statute. I can remove this and probably add an assert / warning if they're not the same. Good catch!

a-ilango: So we would have to maintain the statute after all as confirmed by Michael on the slack discussion.

src/backend/expungeservice/expunger/analyze.py
a-ilango: For time eligibility analysis, we would have to look at all the charges on the client. Time Eligibility of most recent charge also depends on other convictions in the past.

src/backend/expungeservice/expunger/analyze.py
a-ilango: I agree with you but I think @michaelatcommunitylaw has to clarify this. From UI perspective, it makes sense to put any eligibility that involves time under "time eligibility".

hmarcks: Yeah, I'll ask him to clarify tomorrow.

src/backend/expungeservice/test/test_expunger.py
NickSchimek: Ah I see! Thanks

src/backend/expungeservice/expunger/analyze.py
NickSchimek: Cool, glad we clarified that!

src/backend/expungeservice/expunger/analyze.py
NickSchimek: OK, thanks for clarifying that.

src/backend/expungeservice/test/test_expunger.py
NickSchimek: Michael responded in slack and said that there are not different levels associated.

a-ilango:
@NickSchimek I've updated the PR based on your comments and re-organized the tests a bit. Please take a look. Thanks!

PR #35, opened 2018-12-07
Create case parser

### LINKED ISSUE #7: Create a case parser
Build a parser that returns a list of charges for a case from the case detail page.

NickSchimek:
## Description
This parser is needed to extract data from the case detail page. There are two tables on the page that are the most important. The first table it parses is the Charge Information table. The second is the Events & Orders of the Court table. The last table if it exists is the Financial Information table, if this table is here the parser simply extracts the balance due.

The main part of the parser has three parts. 
1.  The first part of the parser adds all relevant data from the Charge Information and Events table into lists.
2. The second part of the parser cleans up the list from the Events table by filtering out any information that is not relevant (only dispositions are relevant) and then adding the dispositions to a hash of hashes using the charge_id as the key for each dispo hash.
3. The third part of the parser adds the data collected from the charge table to a hash of hashes by using the charge_id as the key for each charge hash.

These two hashes can now easily be combined to create charge objects and associate it to the relevant case along with the balance due for the case.

## Testing
I added quite a few test fixtures to capture all the various forms the case detail page can take. The various information found in these tests should also be useful for creating unit tests for the expungement analysis tool and creating system tests.

### Reviewing
I created many small commits with detailed explanations for what is being added and why so that a reviewer can simply click on the first link and iterate through them with a story of what is being done and why. Please let me know if you have any questions or see anything that can be done better. I will squash these before merging.

Fixes #7


src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -0,0 +1,24 @@
+from html.parser import HTMLParser
+
+from expungeservice.crawler.models.charge import Charge
+
+
+class CaseParser(HTMLParser):
+
+    def __init__(self):
+        HTMLParser.__init__(self)
+        self.charges = []
```
NickSchimek: I think this line isn't used. I was originally going to use it, and decided I agree with what @a-ilango was suggesting and create the charge classes from outside of the parser. I will remove it now.

PR #61, opened 2019-01-24
Normalize charge statute at initialization

### LINKED ISSUE #60: statute should be normalized before persisting
## Description
Currently the charge class is initialized with a statute with whatever string value is passed to it. The Oregon e-courts website displays statutes in a variety of representations. A statute may be 6, 7, or 8 characters long, and may or may not contain decimal points, parenthesis, and other separators. 

Have the charge initializer call a method to normalize the statute so that only alphanumeric characters are being persisted.

### This task is done when...
 - [x] A charge class is initialized with a statute containing only alphanumeric characters.

NickSchimek:
## Description

Statutes are 8 characters long and include one '.' and two parentheses,
however they're persisted to the API's database inconsistently. Created
a method to strip the statute of all characters that are not
alphanumeric. This should enable us to consistently extract the
different parts of the statute for analysis.

Fixes #60

### DIFF
--- a/src/backend/expungeservice/crawler/models/charge.py
+++ b/src/backend/expungeservice/crawler/models/charge.py
@@ -14,4 +14,4 @@ class Charge:
 
     @staticmethod
     def __strip_non_alphanumeric_chars(statute):
-        return re.sub(r'[^a-zA-Z0-9*]', '', statute)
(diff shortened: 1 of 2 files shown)


NickSchimek:
hmmm should also uppercase all characters as well

PR #66, opened 2019-02-11
Add authentication endpoints

a-ilango:
This PR adds register, login endpoints to the backend. The login endpoint uses JWT for stateless authentication. The endpoints don't talk to the database yet.

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -119,9 +119,10 @@ Global headers:
 - `Accept: application/json`
 
 
-**`POST`** `/api/login`
+**`POST`** `/api/<version>/auth_token`
 
-Logs user into system
+Post username, password to create an auth token (JWT) that can be used when
+accessing protected APIs
 
 `POST` body:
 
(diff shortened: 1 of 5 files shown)


a-ilango:
I've updated the API routes to reflect usage of JWT for authentication. Btw, do we need both username and email address when registering / logging in?

hmarcks:
Let's do just email please

doc/design.md
```diff
@@ -201,7 +207,7 @@ Returns: Search results
     * tbd
 
 
-**`GET`** `/stats`
+**`GET`** `/api/<version>/stats`
```
maxkwallace: Hey Arun! @a-ilango please let me know if this comment isn't helpful, definitely not trying to tell you what to do, but I thought I'd just share my thoughts. I'm taking a look at your PR ahead of tonight's Meetup to get a sense of where the project is and how auth is going to fit in with the frontend code.

I'd recommend not versioning the API unless we're planning to support multiple and/or 3rd party clients (I don't think we are). Versioning is great in those circumstances since you can make changes without breaking things. But since, AFAIK, we just have the one React JS client, I think it'll be better and simpler to not version the API and instead make changes to the client and the server at the same time. It will save time and be less complex. Just my $0.02.

erikjasso: @a-ilango Sending data to the server should be POST instead of GET, and POST should return 201 on success. I agree with the English wording about getting a token but this kind of operation is done w/POST. Returning 401 UNAUTHORIZED on bad username or password seems reasonable

a-ilango: Okay. I can change it back to POST. Why do we want to use "201 created"? Is it because we create a token based on user supplied data?

a-ilango: @maxkwallace, @erikjasso and I had an offline discussion regarding this and we decided to keep the versioning. In case there's a plan to create a public API in the future, the clients would be ready to handle the version and it should probably not be hard to handle it now.

a-ilango:
@hmarcks I'll remove the username in a follow-up PR. Merging this now.

doc/design.md
erikjasso: Yes, exactly. Any time we are sending data to the server for processing or for storing in the database it should be in a POST or PUT body. In this case we should use POST

PR #69, opened 2019-02-14
Proxy from frontend to backend in dev

maxkwallace:
Configures the frontend package.json to proxy unhandled requests to the backend server (which seems to run on port 5000) in development. This is the first step for connecting the frontend to the backend.

When getting set up, my repo picked up some changes to the package.json lockfile and the Pipfile. I didn't intentionally do anything to cause those changes. If you'd prefer to leave them our of this PR, let me know, I can remove those changes and resubmit.

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -2,6 +2,7 @@
   "name": "record-expunge-pdx",
   "version": "0.1.0",
   "private": true,
+  "proxy": "http://localhost:5000",
   "authors": [
     "rex@codeforpdx.org"
   ],
--- a/Pipfile
+++ b/Pipfile
@@ -10,8 +10,8 @@ flask-script = "*"
 flask-bcrypt = "*"
 pyjwt = "*"
 python-dotenv = "*"
-recordexpunge = {path = ".",editable = true}
-expungeservice = {path = ".",editable = true}
+recordexpunge = {editable = true,path = "."}
+expungeservice = {editable = true, path = "."}
 pytest = "*"
 requests-mock = "*"
 requests = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -10,8 +10,8 @@ flask-script = "*"
 flask-bcrypt = "*"
 pyjwt = "*"
 python-dotenv = "*"
-recordexpunge = {editable = true,path = "."}
-expungeservice = {editable = true, path = "."}
+recordexpunge = {path = ".",editable = true}
+expungeservice = {path = ".",editable = true}
 pytest = "*"
 requests-mock = "*"
 requests = "*"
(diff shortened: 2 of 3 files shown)


NickSchimek:
Thanks @maxkwallace Let's leave the updated changes in `pipfile` and `pipfile.lock` out of the PR for now. Everyones system seems to create these files differently and I'm not sure why the app name in the `pipfile.lock` is being updated and what the optional keys are for so let's disregard those changes for now. do you know or does @erikjasso know if the name in the `pipfile.lock` file matter?

src/frontend/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "record-expunge-pdx",
   "version": "0.1.0",
   "private": true,
+  "proxy": "http://localhost:5000",
```
NickSchimek: I have question. Having separate frontend and backend frameworks is new to me, and so connecting the two is completely new to me. Hypothetically if this was being deployed we would also need to set a production proxy too, I assume? if so where would we set this?

maxkwallace: Here's how this typically works in prod:

When deploying the frontend code, it gets transpiled, minified, etc. (in our case, by webpack) to a set of static assets, at minimum, just an empty HTML page with a script tag, where the script tag links to a big JS file that's our React app.

When someone visits the home page in prod, the "home" route in Flask serves that HTML page, the client downloads the JS file, and then the React app boots, and most of the remaining requests to the server are generated by the React app.

The simplest way to do this is: when deploying the backend, have a task that runs on the filesystem to build the frontend code. If the built frontend assets are in the filesystem that gets deployed, then the "home" route in Flask would look like:

    with open('home.html', 'r') as home_file:
      html_text = home_file.read()
      return html_text

There is no proxy in production. We have a proxy in dev so that we can run an additional server for the frontend code, which massively speeds up the dev experience by incorporating new changes to the React code when you edit a file (without recompiling the whole JS asset) and autoreloading the page so you don't have to manually refresh to see the changes.

maxkwallace:
@NickSchimek thanks for the review! I removed the changes in the `Pipfile` and `package-lock.json`.

src/frontend/package.json
NickSchimek: I see. Thanks for the detailed explanation!

NickSchimek:
Looks good. Feel free to merge. 
I wonder if we should explicitly state the host address `http://127.0.0.1:5000/` as some machines (Ubuntu) may not have the `localhost` name configured. Could also provide instructions on how to configure host names too, I believe for Ubuntu the `/etc/hosts` file needs to be updated. What do you think?

maxkwallace:
@NickSchimek Thanks! I'm going to go ahead and merge.

I've always used Ubuntu for development and AFAIK I've never needed to configure `localhost` and I've never personally updated the `/etc/hosts` file, so this isn't something I know a lot about. Happy to make those changes though, I'd just need to know more about the circumstances under which they could help someone. I.e. "this is when you need to configure host names" so I'd understand the context for explaining that to someone and what they might be confused about.

PR #71, opened 2019-02-22
Refactor case parser

### LINKED ISSUE #70: refactor case parser
The case parser could be refactored to a state design pattern.

NickSchimek:
Implemented the state design pattern on the case parser. This cleans up some of the logic and will make it easier to extend in the future.

Fixes #70


NickSchimek:
Thanks @a-ilango  - No, the tests don't need to be updated in this case. Usually when a state design refactoring is done the tests need to be updated because the calling method is different due to the different states that can be invoked, however in this case the parser automatically invokes the state not the user/tests.

Also the observable behavior has not changed only the implementation, and therefore no "new" tests are needed as the pre-existing behavior is already under test.

PR #79, opened 2019-03-15
Issue 73: Added info on setting up frontend to README.md

ash-tamraz:
[For issue 73](https://github.com/CodeForPortland/recordexpungPDX/issues/73), added information on how to set up the frontend to README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,65 +4,131 @@ A project to automate expunging qualifying criminal records.  This project is do
 [![Waffle.io - Columns and their card count](https://badge.waffle.io/CodeForPortland/recordexpungPDX.svg?columns=all)](https://waffle.io/CodeForPortland/recordexpungPDX)
 
 [![Build Status](https://travis-ci.com/CodeForPortland/recordexpungPDX.svg?branch=master)](https://travis-ci.com/CodeForPortland/recordexpungPDX)
+
+
+## Table of Contents
+
+- [Installation](#installation)
+- [Running Components](#running-components)
+- [Testing](#testing)
+- [Project Layout](#project-layout)
+- [Contributing](#contributing)
+- [License](#license)
+
 ## Installation
 
-Fork, and clone the repo.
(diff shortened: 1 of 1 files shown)


maxkwallace:
@ash-tamraz thanks for the PR! I'll take a look tomorrow.

README.md
```diff
@@ -62,6 +62,38 @@ $ make clean
 
 will remove build artifacts.
 
+### Frontend
```
maxkwallace: I think if we are going to have a "Frontend" heading for the frontend part of the README, we should equivalently have a "Backend" heading for the backend part, otherwise it may be inconsistent and less clear. So it would be good to update the existing text to add that.

README.md
```diff
@@ -62,6 +62,38 @@ $ make clean
...
+
+#### Set Up & Running Development Server
+
+1. **Follow all the installation and set up instructions above**
```
maxkwallace: The existing instructions in the README use bolding this way:

```
 6. **Commit** changes to your branch (never to master)
```

which is inconsistent with the style you have here. I think the style needs to be consistent, otherwise the README will be harder for people to read and less clear. I think there are two options:

1. Change the existing style to match what you have here
2. Change the style you use here to match the existing style

I'm okay with either-- just go with the one that you think makes it easiest for people to read and understand the README.

README.md
ash-tamraz: Thanks for the great feedback! I was actually going to open an issue for fixing the formatting on the backend set up section, I think it's a little confusing, but I'll fix it in this PR if that works.

README.md
ash-tamraz: I'll edit my added text to match the conventions followed by the existing style (e.g. the 'Contributing' section).

README.md
maxkwallace: Yup, go for it!

maxkwallace:
@ash-tamraz looks great! Go ahead and merge :)

Issue #44, opened 2019-01-03
Containerize Flask app


erikjasso:
Build a Docker container for our Flask application. This container will run the Python/Flask code.

This involves:
- adding a Docker file to `src/backend/expungeservice`
- adding a stanza to `docker-compose.yml`
- adding targets to `Makefile` for building the image, starting and stopping the container (see how the database container is done)
- flesh out `src/backend/expungeservice/wsgi.py`
- pick a base image, such as [library/ubuntu](https://hub.docker.com/_/ubuntu/) to reference in `src/backend/expungeservice/Dockerfile`

erikjasso:
Working on this

erikjasso:
Opened pull request. https://github.com/CodeForPortland/recordexpungPDX/pull/81

wittejm:
@erikjasso  I think the recent update from psycopg version 2.7 to 2.8 by dependabot broke this image build. 

According to the doc: http://initd.org/psycopg/docs/faq.html#problems-compiling-and-installing-psycopg2

Version 2.7 included automatic install of a postgres dependency but 2.8 takes that out and the dependency needs to be installed manually to build the package from source. 

Alternatively the pre-compiled package  psycopg2-binary can be used instead (according to the install doc), but is not recommended. 

Adding the prerequisite libpq-dev to the Dockerfile fixes the issue. Is that the best solution?

Tagging @a-ilango too because we were talking about this bug earlier.

NickSchimek:
>Adding the prerequisite libpq-dev to the Dockerfile fixes the issue. Is that the best solution?

Yes, I would say that it is the best solution

erikjasso:
Thanks for digging into this, Jordan. Sounds like you hit on the correct solution.

erikjasso:
@wittejm Is this ready to be closed?

wittejm:
@erikjasso It sure is!  Thanks for spotting it, will close now

PR #81, opened 2019-03-25
#44: Flask app running in Docker container.

erikjasso:
Addresses #44. Adds image w/our basic Flask app and /hello endpoint and Make target for building and running container:

    make dev_stop
    make dev
    docker ps

Output of `docker ps` should look like this:

    CONTAINER ID        IMAGE                            COMMAND                  CREATED             STATUS              PORTS               NAMES
    492eb756d544        recordexpungpdx:database         "docker-entrypoint.s…"   4 minutes ago       Up 4 minutes        5432/tcp            recordexpungpdx_db.1.fffzr80l3jrfo7whmnw6omzzn
    19206f5a054b        recordexpungpdx:expungeservice   "flask run --host=0.…"   5 minutes ago       Up 4 minutes        5000/tcp            recordexpungpdx_expungeservice.1.548jfy94ws71vav75gxt4ssuv

### DIFF
--- a/src/backend/expungeservice/Dockerfile
+++ b/src/backend/expungeservice/Dockerfile
@@ -1,0 +1,27 @@
+FROM library/ubuntu
+RUN apt-get update &&\
+    apt-get install -y python3.7 python3.7-dev python3-pip curl &&\
+    python3.7 -m pip install pip --upgrade &&\
+    python3.7 -m pip install wheel
+
+ #&&\
+ #curl https://raw.githubusercontent.com/kennethreitz/pipenv/master/get-pipenv.py | python3.7
+
+ADD requirements.txt /tmp
+WORKDIR /tmp
+RUN python3.7 -m pip install -r requirements.txt && rm requirements.txt
+
+RUN mkdir -p /var/www/expungeservice
+COPY . /var/www/expungeservice
+WORKDIR /var/www/expungeservice
+RUN python3.7 -m setup install 
+
+# This exposes a port; docker-compose file sets mapping.
+EXPOSE 5000
+
+WORKDIR /var/www
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+ENV FLASK_APP="expungeservice:create_app('development')"
+ENTRYPOINT ["flask"]
+CMD ["run", "--host=0.0.0.0", "--port=5000"]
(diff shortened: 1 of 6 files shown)


erikjasso:
Just realized I hadn't merged w/master, and leapfrogged a number of changes. Fixing...

erikjasso:
Okay. Fixed and ready for review

NickSchimek:
Thanks @erikjasso !

src/backend/expungeservice/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    apt-get install -y python3.7 python3.7-dev python3-pip curl &&\
+    python3.7 -m pip install pip --upgrade &&\
+    python3.7 -m pip install wheel
+
+ #&&\
+ #curl https://raw.githubusercontent.com/kennethreitz/pipenv/master/get-pipenv.py | python3.7
+
+ADD requirements.txt /tmp
+WORKDIR /tmp
+RUN python3.7 -m pip install -r requirements.txt && rm requirements.txt
```
a-ilango: Can't we invoke pipenv to install the dependencies inside docker?

erikjasso: I didn't think it made sense to use pipenv inside a docker container since the container is an isolated environment, but maybe you mean why maintain two dependency files? That's a good point

a-ilango: > maybe you mean why maintain two dependency files? 

Yes, I was of the idea that pipenv is supposed to replace the requirements.txt method of tracking dependencies.

It seems requirements.txt may be needed if we want to package / install the application according to https://medium.com/@greut/building-a-python-package-a-docker-image-using-pipenv-233d8793b6cc

It seems also possible to use pipenv within docker and there are some practices to follow to avoid docker rebuilding the layer repeatedly according to https://jasonfavrod.com/developing-with-docker/

I suppose we can go with what's known or simple and easy to maintain.

erikjasso: I compromise and use pipenv to generate requirements.txt in the makefile, then use pip to install the generated requirements.txt in the container. This way we maintain one file of requirements. One trade-off is that we install more into the container than is probably required

erikjasso:
Merged.

PR #97, opened 2019-04-25
Backend/remove username field

wittejm:
This includes changes to endpoint signatures to no longer have a 'username' field. All instances of 'email address' as a json field were replaced with 'email' to sync with other endpoints. 

Changes are reflected in the design doc and test cases. requesting review from front-end folks @hmarcks and @maxkwallace because of API changes.

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -19,7 +19,7 @@ User Flow
 
 User visits login page in browser
 
-User types username and password to log in
+User types email and password to log in
 
 User is directed to main page
 
@@ -70,7 +70,7 @@ These routes are set up in the front-end application for navigating between the 
 
 Login page
 
-- Form for entering `username` and `password`
+- Form for entering `email` and `password`
 - Application sends credentials to server
 - Server validates credentials
 - If credentials are good server returns session cookie (contents tbd)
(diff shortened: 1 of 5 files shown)


a-ilango:
@wittejm thanks for the PR!

I think some of these commits can be squashed together. Can you please squash those that change the same file and/or make sense to be together in a single commit?

wittejm:
@a-ilango Yea I think squashing it all into one makes sense, so I'll do that. I haven't done much group work on git before so wasn't sure that the PR would include all this clutter.

doc/design.md
```diff
@@ -152,17 +151,15 @@ Creates a user
 
 - format: `JSON`
 - fields:
-    * username
-	* encrypted password
-	* email address
+	* email
+    * encrypted password
```
maxkwallace: I don't mean to suggest that this is wrong, honestly just wondering: is there a reason why the indentation is different here for email and encrypted password?

a-ilango: Do you mean before this change? I wonder if spaces and tabs are getting mixed up.

wittejm: It was a tabs and spaces thing. I just fixed those

Issue #45, opened 2019-01-03
Build React app into webserver container


erikjasso:
Build the webapp into a Docker volume used by the webserver container. Write a Make target to control that process

See issue #43

erikjasso:
Ask the front-end devs what they are using to build the React app

maxkwallace:
@erikjasso right now the frontend is set up to use the create-react-app defaults (Webpack) to do builds and I don't see a need to change that in the future. You can do a build in that directory with the command `npm run build` and it looks like this:

```
mkw@M4600:~/Documents/repos/recordexpungPDX/recordexpungPDX/src/frontend$ npm run build

> record-expunge-pdx@0.1.0 build /home/mkw/Documents/repos/recordexpungPDX/recordexpungPDX/src/frontend
> react-scripts-ts build

Creating an optimized production build...
Starting type checking and linting service...
Using 1 worker with 2048MB memory limit
(node:8496) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
Compiled successfully.

File sizes after gzip:

  34.18 KB  build/static/js/1.be79822f.chunk.js
  3.56 KB   build/static/js/main.7dcb98c8.chunk.js
  1.02 KB   build/static/css/main.60428d27.chunk.css
  763 B     build/static/js/runtime~main.229c360f.js

The project was built assuming it is hosted at the server root.
You can control this with the homepage field in your package.json.
For example, add this to build it for GitHub Pages:

  "homepage" : "http://myname.github.io/myapp",

The build folder is ready to be deployed.
You may serve it with a static server:

  npm install -g serve
  serve -s build

Find out more about deployment here:

  http://bit.ly/CRA-deploy
```

maxkwallace:
If we decide to use Hugh's server-rendered HTML (Flask, Jinja) code then we can close this issue since there will be no separate frontend app.

erikjasso:
@maxkwallace I've never used Webpack. What does it do?

maxkwallace:
@erikjasso Webpack is a tool that's primarily used to transform Javascript code files on the filesystem (i.e. what we edit in a repo) into static assets, typically and most importantly, one big Javascript file that's the result of bundling together and minifying all the code you need.

Issue #43, opened 2019-01-03
Containerize Nginx


erikjasso:
Make  a Docker container for the webserver. The weberver will:

- [x] - serve the React application
- [x] - route requests to the Flask service
Involves:
- [x] - writing an Nginx config file `config/nginx/nginx.conf` with static route to the web app and routes to the Flask services
- [ ] - configuring Nginx to use HTTPS
- [x] - adding a stanza for the web server to `docker-compose.yml`
- [x] - adding a target to `Makefile` to build the web server image
- [ ] - adding a target to `Makefile` to build the web application as a docker volume that the web server will use to serve static html/css/images/etc

We also need to pick a Docker image to use for this task. Something like this: [nginx - Docker Hub](https://hub.docker.com/_/nginx) seems like a good place to start

wittejm:
Given @maxkwallace’s comment in issue #45, is it correct that the nginx server would not need to serve any static content? Meaning there’s no need to set up the additional docker volume?

wittejm:
I realized that because of the reliance on react knowledge (which I don't have) and that it sounds like some react design questions are still unresolved (like issue #83 ), I should not actually be the one working on this.

erikjasso:
@wittejm Much of this task of configuring the web server and running it in a Docker container can be done while we figure out how the front-end code and assets are built. In fact this should be orthogonal to React. Independent of which front-end framework we use and how it is built we will still use a web server to serve the static content (as well as route requests of dynamic content to uWSGI). Does that make sense?

wittejm:
Ok, that partially makes sense. Initially I was misunderstanding the role of the nginx web server and thought it should only reroute http requests directly to the uWSGI server, which doesn't seem right. So, it sounds like instead the nginx configuration needs to expose uWSGI only for requests made by the frontend, which will still behave as though it is accessing a local socket. Is that correct?

The question of static content pathing is still unclear to me (even though it seems like it should be simple). It sounds like, for nginx to serve static content it only needs to specify the path from the server’s root directory to the public/ directory. Then the block in the nginx config file would be:

location / { 
   root /path/to/build/public/; 
}
and that path is the destination of the react build.

erikjasso:
Right, we'll have a `location` block like that for the static web site and one for the Python API (as well as one or more `server` blocks listening on some port(s)).

Incoming requests will have the base form:

    https://server_name/path

where

    server_name is our domain name when deployed, or localhost/127.0.0.1 during development
[...]

PR #105, opened 2019-05-01
webserver container with static files and api-call forwarding

wittejm:
This is a first pass at the webserver container. The included readme details the changes and my concerns. The main feature this is missing is the configuration to use https but I'd appreciate any comments on what's here so far.


wittejm:
Addresses issues #43 and #45 .

config/nginx/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+### STAGE 1: Static web files ###
+
+FROM node as build
+RUN mkdir /usr/src/app
+ENV PATH /usr/src/app/node_modules/.bin:$PATH
+COPY frontend/package.json /usr/src/app/package.json
+
+WORKDIR /usr/src/app 
+RUN npm install --silent
+# RUN npm install react-scripts -g --silent # from tutorial but missing in project docs.
```
maxkwallace: By "project docs" do you mean the README for /frontend or other docs for this project? Just want to make sure I understand what this is referring to.

config/nginx/README.md
```diff
@@ -0,0 +1,28 @@
...
+* In the make file, added the webserver_image target to the dev_deploy target’s dependencies.
+* Missing feature: https configuration (and cert).
+* The Nginx docker image already redirects the server access and error logs to std_out and std_err, so docker logging captures both. 
+* The destination of static files in the image is just nginx's default directory for static files: /usr/share/nginx/html/ . 
+* Building the image requires that the frontend code has already been built in src/frontend/build, as described in src/frontend/README.md (just cd to src/frontend and run "npm run build"). I don't know the best workflow to put these steps together so it's just manual for now.
+
+## Behavior issue to be aware of:
+* After visiting the webpage, nginx will forward requests to the backend only if you do a hard refresh (shift-F5) on the endpoint url. Otherwise, that url first hits a react service which redirects to the site’s login page instead of reaching the server (because react doesn't recognize the endpoint uri). This stackoverflow post describes the issue:
+https://stackoverflow.com/questions/51120222/how-to-serve-react-app-and-django-blog-from-nginx
```
maxkwallace: Is this something that you are seeing in our app on the current `master` branch?

The SO post you linked says that the solution is probably to turn off Service Workers in the React app. I already did this https://github.com/CodeForPortland/recordexpungPDX/pull/104 so it's possible that merging that PR will fix the issue.

I don't totally understand the behavior you're describing; we should probably discuss this verbally at the next meeting.

config/nginx/Dockerfile
wittejm: I was looking at the frontend install instructions in the project's top-level README. But after looking around more I just realized that react-scripts is included in the dependencies in package.json . 

I should take out that commented line from the Dockerfile if you agree it doesn't belong there.

config/nginx/README.md
wittejm: I find that with the app running on master (using npm start instead of nginx), every path on the website redirects to the login page rather than returning 404. I don’t know if that’s the intended behavior. But without running it on the nginx server which is supposed to redirect paths matching api/*, the comparison doesn't mean much.

I agree, let’s check in about this in person.

config/nginx/Dockerfile
maxkwallace: Let's chat about this at the next meeting. Just wanted to make sure I understood your comment. 

Having something in the `package.json` might not be enough for documentation purposes. Main thing is I'm not sure if `react-scripts` is only needed when running create-react-app the first time or every time the project is built.

config/nginx/README.md
maxkwallace: I don't think that's intended behavior but it's probably expected behavior given the way the app is set up now.

wittejm:
I'm rejecting this because it's super old and I'll write a new and updated version.

PR #106, opened 2019-05-02
Added lib dependency for psycopg2

wittejm:
I updated the Dockerfile and the readme, but only with an instruction for Linux and not for Mac or Windows, because I don't know those environments..

Also fixed the (typo) numbering on other install instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,24 @@ A project to automate expunging qualifying criminal records.  This project is do
 	package manager which also automatically creates and manages virtual 
 	environments.
 
+4. Install additional libraries needed for running the backend natively or with pipenv:
+
+    * **Mac**
+
+      TBD
+
+    * **Linux**
+      ```
+      sudo apt-get install libpq-dev -y
+      ```
+    * **Windows**
+
+      TBD
+
 5. Install NPM if you don't already have it installed. [This link provides 
 	instructions on how to install Node.js and NPM](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
 
-4. Install backend dependencies:
+6. Install backend dependencies:
 
 	A [Makefile](https://www.gnu.org/software/make/) controls installing 
   dependencies, running the Flask app, and removing build artifacts.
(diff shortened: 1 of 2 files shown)


NickSchimek:
Thanks @wittejm. This looks good. I believe for the mac the `libpq` headers are installed if `postgresql` is installed using `homebrew`. If your on a mac you may or may not need to run `export LDFLAGS="-L/usr/local/opt/openssl/lib”` after installing `postgresql` via `brew install postgresql`.

NickSchimek:
For devs wanting to use windows I would suggest that they use the docker image or partition ubuntu onto their hard drive. There's been to many issues trying to get Windows users on-boarded, and I don't think it's worth the headache at this point.

README.md
```diff
@@ -56,10 +56,29 @@ A project to automate expunging qualifying criminal records.  This project is do
 	package manager which also automatically creates and manages virtual 
 	environments.
 
+4. Install additional libraries needed for running the backend natively or with pipenv:
+
+    * **Mac**
+
+      ```
+      brew install postgres
```
NickSchimek: Brew command for postgres is actually `brew install postgresql` <- this is only needed if the user doesn't have postgres installed already. More information regarding the install can be found here https://wiki.postgresql.org/wiki/Homebrew

erikjasso: I just want to point out that for Windows users who want to use Ubuntu there is an Ubuntu app on the Microsoft Store that runs Ubuntu natively in Windows, no need to partition your disk and install yourself.

Installing Ubuntu app:
https://docs.microsoft.com/en-us/windows/wsl/install-win10

Installing Docker Desktop on Windows:
https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly

It does take some PowerShell-fu that was new to me coming from Linux but once Ubuntu and Docker Desktop are configured, you can run the Ubuntu app from the Windows menu or type 'bash' from PowerShell to get a bash prompt (or use a bash-friendly console like ConEmu)

wittejm: Woops I will fix that line add that link for more information re the homebrew install. 

Given @erikjasso 's comment, should we just point that out at the top of the README and remove instructions for installing components on Windows?

PR #110, opened 2019-05-04
Assign balance due to Case object

### LINKED ISSUE #101: Assign balance due to case object
## Description
Currently the balance due is being parsed by the `case_parser`, and is assigned to the case parser object here: https://github.com/CodeForPortland/recordexpungPDX/blob/master/src/backend/expungeservice/crawler/parsers/case_parser/case_parser.py#L20, however that information is lost as it's not being assigned to the case object by the crawler.

### This task is done when:

- [ ] Create a test for assigning the `balance_due` attribute to the case object.
- [ ] Add `balance_due` as an attribute of the Case object.
- [ ] Implement the balance_due assignment.

#### Additional information:
- The test can be added to `test_crawler.py` file here: https://github.com/CodeForPortland/recordexpungPDX/blob/master/src/backend/tests/crawler/test_crawler.py#L71 : currently here at line 71 it is parsing the `CASEJD1` file, however that file has a balance due of $0.  Add test to this test method that checks it assigns `0.00` balance due, and create a new test method that parses the case `CASE_X1`. Case X1 has a balance due of $1,516.80 as shown here: https://github.com/CodeForPortland/recordexpungPDX/blob/master/src/backend/tests/fixtures/case_details.py#L70 It is important to test that it assigns a non-zero value as the default value is zero.
- Once the test is built the attribute can be added in the `Case` class here: https://github.com/CodeForPortland/recordexpungPDX/blob/master/src/backend/expungeservice/crawler/models/case.py#L14
- Assignment can be done after parsing the case file here: https://github.com/CodeForPortland/recordexpungPDX/blob/master/src/backend/expungeservice/crawler/crawler.py#L36

wittejm:
closes #101

### DIFF
--- a/src/backend/expungeservice/crawler/models/case.py
+++ b/src/backend/expungeservice/crawler/models/case.py
@@ -12,4 +12,5 @@ class Case:
         self.date = datetime.date(datetime.strptime(date, '%m/%d/%Y'))
         self.violation_type, self.current_status = type_status
         self.charges = charges
+        self.balance_due = 0.00
         self.case_detail_link = case_detail_link
--- a/src/backend/expungeservice/crawler/models/case.py
+++ b/src/backend/expungeservice/crawler/models/case.py
@@ -15,10 +15,10 @@ class Case:
         self.balance_due_in_cents = 0
         self.case_detail_link = case_detail_link
 
-    def set_balance_due_dollar_amount(self, balance_due_dollar_amount):
+    def set_balance_due(self, balance_due_dollar_amount):
         if type(balance_due_dollar_amount) == str:
             balance_due_dollar_amount = float(balance_due_dollar_amount.replace(',',''))
         self.balance_due_in_cents = int(balance_due_dollar_amount * 100)
 
-    def get_balance_due_dollar_amount(self):
+    def get_balance_due(self):
         return self.balance_due_in_cents / 100
--- a/src/backend/expungeservice/crawler/models/case.py
+++ b/src/backend/expungeservice/crawler/models/case.py
@@ -12,5 +12,13 @@ class Case:
         self.date = datetime.date(datetime.strptime(date, '%m/%d/%Y'))
         self.violation_type, self.current_status = type_status
         self.charges = charges
-        self.balance_due = 0.00
+        self.balance_due_in_cents = 0
         self.case_detail_link = case_detail_link
+
+    def set_balance_due_dollar_amount(self, balance_due_dollar_amount):
+        if type(balance_due_dollar_amount) == str:
+            balance_due_dollar_amount = float(balance_due_dollar_amount.replace(',',''))
+        self.balance_due_in_cents = int(balance_due_dollar_amount * 100)
+
+    def get_balance_due_dollar_amount(self):
+        return self.balance_due_in_cents / 100
--- a/src/backend/expungeservice/crawler/crawler.py
+++ b/src/backend/expungeservice/crawler/crawler.py
@@ -34,7 +34,7 @@ class Crawler:
         # Parse search results (case detail pages)
         for case in self.result.cases:
             case_parser = self.__parse_case(case)
-            case.set_balance_due_dollar_amount(case_parser.balance_due)
+            case.set_balance_due(case_parser.balance_due)
             for charge_id, charge in case_parser.hashed_charge_data.items():
                 new_charge = Crawler.__build_charge(charge_id, charge, case_parser)
                 case.charges.append(new_charge)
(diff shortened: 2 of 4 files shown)


NickSchimek:
Thanks @wittejm This is perfect. The only thing I have to suggest, which I overlooked when writing the issue, is that the balance due should be persisted as an integer since we're dealing with currency. This requires to be done due to the way computers store floating point integers (It is not possible to accurately represent base 10 floating point numbers in base 2). If we're dealing with one floating point currency representation it may be OK, but each case may have a balance due which will all need to be added together, and once arithmetic is performed on the floating point numbers is when the inaccuracy is always seen.

There are a few different ways to achieve this. I will let you decide which way is best or reach out if you need a suggestion. 

Depending how the total due algorithm is implemented this may or may not actually be an issue, but its always good practice to store currency as an integer.

wittejm:
I'm inclined to save the value in cents and then put getter and setter methods on the Case object for dollar values. Though I notice none of the existing model objects have getters and setters so I'm not sure if this is a preferred approach. 

It does mean that summing balances as integer in a forthcoming feature wouldn't use these methods. And since sums would operate directly on cent values then it makes sense to not format the internal balance due field as private.

Here's the code change to show what I mean (will squash before merge).

wittejm:
(tag @NickSchimek ; Idk if you get a similar notification just from the new comment on an issue if you're a participant?

src/backend/expungeservice/crawler/models/case.py
```diff
@@ -12,4 +12,13 @@ def __init__(self, info, case_number, citation_number, date_location, type_statu
         self.date = datetime.date(datetime.strptime(date, '%m/%d/%Y'))
         self.violation_type, self.current_status = type_status
         self.charges = charges
+        self.balance_due_in_cents = 0
         self.case_detail_link = case_detail_link
+
+    def set_balance_due_dollar_amount(self, balance_due_dollar_amount):
+        if type(balance_due_dollar_amount) == str:
+            balance_due_dollar_amount = float(balance_due_dollar_amount.replace(',',''))
+        self.balance_due_in_cents = int(balance_due_dollar_amount * 100)
```
NickSchimek: Question: If line 19 and line 21 are deleted do all the tests still pass?

src/backend/expungeservice/crawler/models/case.py
```diff
@@ -12,4 +12,13 @@ def __init__(self, info, case_number, citation_number, date_location, type_statu
         self.date = datetime.date(datetime.strptime(date, '%m/%d/%Y'))
         self.violation_type, self.current_status = type_status
         self.charges = charges
+        self.balance_due_in_cents = 0
```
NickSchimek: This attribute name was a bit confusing at first as I'm used to seeing the getter/setters the same name as the attribute, but after a bit I understand the thought process here.

NickSchimek:
Thanks @wittejm - This is looking good. 

#### Regarding testing:
I generally wouldn't unit test getters/setters, but as soon as they start manipulating the value then I would create a couple unit tests, however they are under functional tests so the case could be made that it's not needed, and of course the case could be made that unit tests have the benefit of isolating the behavior under test so that in the future if there are issues you can more quickly find the culprit. I'm fine either way in this scenario just wanted to throw that out there.

#### Possible alternatives:
I have a couple suggestions: You can take them or leave them as you see fit.

1. Update your existing code to name the attribute `balance_due`, and set the setters/getters to `set_balance_due`, and `get_balance_due`. My reasoning behind this is the implementation of the setters/getters should be invisible to the user.
2. Store the balance due as a string, leaving the original string intact. When this value is passed to the frontend it doesn't care the value is a string as it will just interpolate that into a view to display on the screen. And then when the PR is done for the total due functionality one getter can be added that transforms the string into an integer that is returned for the calculations.

src/backend/expungeservice/crawler/models/case.py
wittejm: Not sure, but I think you're referring to lines 19 and 20? The test case for setting nonzero balance happens to be over 1000 and has a comma in the string value, so without checking for this then the math operation (and test) does fail.

NickSchimek: haha I shouldn't review PR's late at night. This is perfect!

NickSchimek:
@wittejm Feel free to merge or make updates as you see fit.

wittejm:
I agree that having a getter and setter that do extra work is inelegant, but:
- I think storing the value as a numeric type is more intuitive than storing as a string (and a string that might still have frustrating commas in it), and it could be weird for internal operations to expect a string attached to the crawler data object while expecting a numeric type in the database (which doesn't exist yet in our table but would be preferable). 
- Renaming the getter and setter to `get_balance_due` and `set_balance_due` sounds fine since dollar amount is the more likely assumed behavior. But given the inelegance of a mismatched internal type to its getter and setter, I think the more explicit internal variable name is nice to have to keep clear what's going on.
- Adding the unit tests for getter and setter sounds good given the added complexity here.

wittejm:
Great. I appreciate all the feedback along the way @NickSchimek. Merging now.

PR #109, opened 2019-05-02
Docker readme instructions

wittejm:
Added a new README section with Docker usage basics and links.

PR #111, opened 2019-05-04
moved / combined docker documentation

wittejm:
Moved my notes from the main readme into the pre-existing docker doc file.

closes #109 (again)

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,47 +157,6 @@ Run all tests with the following command:
 $ make test
 ```
 
-## Using Docker
-
-The above instructions are for running components locally, but the app will be deployed as a docker stack in production. Documentation links and basic docker operations for development are provided here:
-
-### Docs
-[Installation](https://docs.docker.com/install/)
-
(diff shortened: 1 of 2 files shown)


NickSchimek:
Looks good. Can you add a navigation link from the README to the `docker.md` file, and also add docker as a possible alternative for installation?

wittejm:
@NickSchimek  Yep I'll add the link to the wiki. I don't know what the approach is for installing the app with Docker for Windows. As a dev environment, Docker setup still relies on running pipenv to generate the requirements.txt file for the expungeservice. To deploy the app purely in Docker containers I think we might want to pull source code from online while building the docker image, which will be a bit of a refactor for all three images. I'm definitely out of my depth here but I'd be interested in talking about this on Wednesday.

NickSchimek:
yeah docker is very handy, but is quite the complex beast from my understanding. I've never dove into it myself, and you definitely know more about it than me at this point. Let's talk more about it Wednesday!.

PR #112, opened 2019-05-06
Create expunger

### LINKED ISSUE #8: Epic: Search if a person has eligible expungeable records
As a user
I would like to enter in a search name and/or date of birth.
so that I can see a list of eligible and non-eligible expungeable records.

Parent of #6, Parent of #7, Parent of #9, Parent of #10

### LINKED ISSUE #9: Automate expunge decision
Build a tool that will analyze each charge and decide if it can be expunged

NickSchimek:
## Description
This PR adds expungement functionality. There are 3 main files in this PR. `time_analyzer.py`, `type_analyzer.py`, and `expunger.py`. Also created the `expungement_result` class which is an attribute of the Charge class. The time and type analyzer evaluates a charge's time and type expunge-ability respectively, and pushes the result to the `expungement_result` object. The expunger class is basically a wrapper class that runs the type and time analyzer, along with gathering needed information that is passed to the time analyzer upon instantiation. Besides updates to the pipfile, charge, and disposition class the remaining files updated/created are test files. I created factory's for the case, charge, and crawler class which made it easier to use these objects while testing. Also created an initial implementation of a factory for a case details page that is located in `case_details.py` file under the method name `case_x` that is used in tests to insert arbitrary variables as needed when mocking the OJCIN website. This file can be further enhanced and used in future tests when needing to mock the OJCIN website.

I created a few factories to dry up the code in the test modules, which also has the added benefit of making the test cases more explicit, and just makes testing easier.

Things to do to finish this up:

- [x] Remove outdated expunger files
- [x] Run type analyzer even if open cases exist
- [x] Move models out into the open
- [x] Add remaining time-based logic for "B" class violations
- [x] Create test for drug free zone variance violation
- [x] Create unit test for non-traffic violation
- [x] Inactive case statuses should be handled the same way as closed cases

This can be used by instantiating the expunger class with a list of cases, and then calling the run method on the expunger object. 

This closes #9, and closes #8


NickSchimek:
@codeforpdx/team-record-expungment feel free to rip this PR apart. I know a few things I would comment on and will if no one else does. Sorry it's so big. A better project manager would've been able to break this apart into smaller issues that everyone could work on. I definitely learned a lot doing this and if it needed to be done again I know I could improve and break it apart into smaller issues/pr's.

wittejm:
@NickSchimek I'd definitely like to try digging into it tomorrow and I'll see how far I can get and/or if I have any useful thoughts. From a quick glance it doesn't look ... **too** overwhelming :)

NickSchimek:
Thanks, I updated the PR description. Hopefully that helps to understand how it's organized.
Rebasing with master due to conflict with `src/backend/tests/crawler/test_crawler.py` and force pushing.

wittejm:
I haven't starting looking at the main expunger logic yet but here are some comments so far mostly on crawler model structure: 

- Defunct files to be deleted, it looks like: expungeservice.expunger.analyzer.py and expungeservice.test.*
- Maybe Charge.level could be an enum? Its only values I see are 'Misdemeanor' ,'Felony Class A', 'Felony Class B', 'Felony Class C', 'Violation', 'Misdemeanor Class A', 'Misdemeanor Class C', 'Violation' , 'Class C Traffic Violation', 'Felony Class F', though I might be missing some.
 - Having ExpungementResult in the crawler.models module doesn’t feel right. And the objects in the models module are used in both crawler and expunger … maybe it makes sense to have all the model objects in an adjacent module e.g. at expungeservice.models? 
 - crawler.search() stores its result as a self field of type RecordParser, and that object attaches case data to itself when it runs its feed() method. Neither object is plain-old-data and has several data fields that are effectively private. Maybe it would be good to have the crawler.search() function return a data object that is actually plain-old-data? This might just mean defining a Record object which the RecordParser populates and provides to the crawler. This might be personal taste but it does seem like it'd be good to have a clean Record object to share between the crawler and expunger, which only contains data of interest to both. 

Overall, this looks really great to me so far and I'm really excited that the core logic of this project is nearing completion. I'll keep looking and let you know if I see anything else.

NickSchimek:
Thanks @wittejm ! Very keen eye to notice all of this. A couple of the issues you mentioned are concerns I also share, and am planning to add refactoring issues for them after this is completed. Such as the models definitely don't belong in the crawler directory, I/we should talk to @erikjasso to see where best they would fit. I would normally put them under the main app directory which would be `src/backend/expungeservice` for this app in order to follow the MVC architecture pattern. This would need to be a separate issue as moving these files don't belong with this PR. 
The `crawler.search()` does return a weird object (record parser) which I didn't really think about when building that and it got missed in the review cycle. When building the expunger I noticed this and I'm glad you mentioned it. We'll need to create a separate ticket to change this behavior. 
I've been delaying the removal of the other expunger files, but it may be time to remove them. What do you think @a-ilango ? 
And now regarding the last item you mentioned regarding the enums. I'm not familiar with using enums, and would like to get more information from you about that tonight.

wittejm:
Just one other question at the moment: The expunger quits early if there are open cases in a record. I saw that the flowchart prescribes this behavior. But would it be useful to still evaluate any other cases?

hmarcks:
I have this as an open question in the requirements doc, but I remember Michael saying it would be useful to see all the records evaluated. I also recall talking about this with @NickSchimek and he said we could evaluate all records...we can talk it over at the meeting, seeya in a few.

a-ilango:
> I've been delaying the removal of the other expunger files, but it may be time to remove them. What do you think @a-ilango ?

If this is replacing the old files, we should remove them.

> And now regarding the last item you mentioned regarding the enums. I'm not familiar with using enums, and would like to get more information from you about that tonight.

If there's a variable that would take only some pre-defined values, we can use enums for them.

NickSchimek:
Yes, We can run the Type eligibility and then quit when there are open cases as the time eligibility would be useless.

src/backend/expungeservice/expungement_analyzer/expunger.py
```diff
@@ -0,0 +1,89 @@
...
+    def _categorize_charges(self):
+        for charge in self._charges:
+            if charge.acquitted():
+                self._acquittals.append(charge)
+            else:
+                self._convictions.append(charge)
+
+    def _set_most_recent_dismissal(self):
+        self._acquittals.sort(key=lambda charge: charge.date)
+        if self._acquittals and self._acquittals[-1].recent_acquittal():
```
a-ilango: Shouldn't this be len(self.acquittals)?

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
```diff
@@ -0,0 +1,74 @@
...
+
+        elif charge.level == 'Felony Class C' or 'Misdemeanor' in charge.level:
+            charge.expungement_result.set_type_eligibility(True)
+            charge.expungement_result.set_reason('Eligible under 137.225(5)(b)')
+
+        elif charge.level == 'Felony Class B':
+            charge.expungement_result.set_reason('Further Analysis Needed')
+
+        else:
+            charge.expungement_result.set_reason('Examine')
```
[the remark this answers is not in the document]
NickSchimek: No, I see that as a separate feature/PR.

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @staticmethod
+    def __ineligible_under_137_225_5(charge):
+        return TypeAnalyzer.__crime_against_person(charge) or TypeAnalyzer.__traffic_crime(charge) or TypeAnalyzer.__felony_class_a(charge)
+
+    @staticmethod
+    def __crime_against_person(charge):
+        statute = int(charge.statute[0:6])
```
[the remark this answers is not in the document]
NickSchimek: Yes, that is a good suggestion especially since this is done more than once. Now just need to think up of a good name for the new attribute...

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
```diff
@@ -0,0 +1,74 @@
...
+        statute_ranges = (range(163305, 163480), range(163670, 163694), range(167008, 167108), range(167057, 167081))
+        return any(statute in statute_range for statute_range in statute_ranges)
```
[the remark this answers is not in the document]
NickSchimek: That may be true. I like the simplicity and readability that comes with using the `in` operator. If we do see performance issues later on we can run an analyzer to see where we need to optimize and go from there.

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @staticmethod
+    def __traffic_crime(charge):
+        statute = int(charge.statute[0:3])
+        statute_range = range(801, 826)
+        return statute in statute_range
+
+    @staticmethod
+    def __felony_class_a(charge):
+        return charge.level == 'Felony Class A'
```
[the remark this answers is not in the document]
NickSchimek: Will have to look into this.

src/backend/expungeservice/expungement_analyzer/expunger.py
NickSchimek: Can you expand on your question. Not sure what your asking here.

a-ilango: Yes, line 76. I was saying the check should be `if len(self._acquittals) and ...` or `if len(self._acquittals) > 0 and ...` like line 81. If `self._acquittals` is an empty list, `self._acquittals[-1]` would throw an exception.

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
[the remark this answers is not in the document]
NickSchimek: @a-ilango and @wittejm I looked into this and setting this as an enum would not be desirable as there are many more different levels than what need to explicitly check for.

NickSchimek:
@a-ilango regarding your efficiency concerns. The expunger doesn't have any code that is asymptotically greater than Big-O(n) where n is the number of charges in a record. I ran on our live test case that has 168 cases and a total of 258 charges and it returned right away. No visible delay.

src/backend/expungeservice/expungement_analyzer/expunger.py
NickSchimek: No, it doesn't throw an exception if `self._acquittals` is an empty list. An empty lists truth value is false. https://docs.python.org/3/library/stdtypes.html#truth-value-testing

NickSchimek: and so it would short circuit and not evaluate `self._acquittals[-1]`

a-ilango:
Yep, it's gonna be O(n) but my comments were on operations that are part of processing the O(n) inputs. Thanks for checking the numbers. If they're quick enough the optimizations are probably not needed at this time.

src/backend/expungeservice/expungement_analyzer/expunger.py
a-ilango: Cool. Then we are good here.

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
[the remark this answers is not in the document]
NickSchimek: Casting this to an int is going to cause issues. Portland's drug free zone variance statute has a letter in the third digit of the statute number. I'm thinking I will check with regex for any non numbers before casting; otherwise just return false. Will have to do that with the other slice [0:3] too.

src/backend/tests/expungement_analyzer/test_type_analyzer_convicted_charges.py
```diff
@@ -0,0 +1,570 @@
...
+        charge = self.create_recent_charge()
+        self.charges.append(charge)
+        self.type_analyzer.evaluate(self.charges)
+
+        assert charge.expungement_result.type_eligibility is None
+        assert charge.expungement_result.type_eligibility_reason == 'Examine'
+
+    # Test non-traffic violation
+
+    # TODO: Implement this test
```
NickSchimek: Will also need to implement this test case.

NickSchimek:
> Yes, We can run the Type eligibility and then quit when there are open cases as the time eligibility would be useless.

@hmarcks This is done.

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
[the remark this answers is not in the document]
NickSchimek: I've noticed some statutes are only four digits long. For example statute 2915.

I'm going to hold off on doing this right now until we get some feedback regarding the initial implementation from Michael. It's definitely something that could be done in a later refactoring.

src/backend/tests/expungement_analyzer/test_type_analyzer_convicted_charges.py
NickSchimek: This is done.

src/backend/expungeservice/expunger/analyzers/type_analyzer.py
```diff
@@ -0,0 +1,70 @@
...
+        for charge in charges:
+            self.__evaluate(charge)
+
+    def __evaluate(self, charge):
+        if charge.acquitted():
+            charge.expungement_result.set_type_eligibility(True)
+            charge.expungement_result.set_reason('Eligible under 137.225(1)(b)')
+
+        elif TypeAnalyzer.__marijuana_ineligible(charge):
+            charge.expungement_result.set_type_eligibility(False)
```
NickSchimek: There are 3 or 4 methods that take a `charge` object as an argument. These would be better placed in the charge class.

src/backend/expungeservice/expunger/analyzers/type_analyzer.py
```diff
@@ -0,0 +1,70 @@
...
+
+        elif charge.level == 'Felony Class B':
+            self.class_b_felonies.append(charge)
+            charge.expungement_result.set_reason('Further Analysis Needed')
+
+        else:
+            charge.expungement_result.set_reason('Examine')
+
+    @staticmethod
+    def __ineligible_under_137_225_5(charge):
```
NickSchimek: There are a few methods that take a `charge` object as a parameter. These would be better placed in the charge class.

src/backend/expungeservice/expungement_analyzer/analyzers/type_analyzer.py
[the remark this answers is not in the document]
NickSchimek: I will do some refactoring later this week. I may fix this at that time.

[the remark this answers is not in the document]
NickSchimek: This has been done. I like it much better now. Thank you for the feedback. Please let me know what you think of the initial implementation.

src/backend/expungeservice/models/charge.py
```diff
@@ -86,3 +87,11 @@ def _felony_class_a(self):
...
     def __strip_non_alphanumeric_chars(statute):
         return re.sub(r'[^a-zA-Z0-9*]', '', statute).upper()
+
+    @staticmethod
+    def __set_section(statute):
+        if len(statute) < 6:
+            return None
+        elif statute[3].isalpha():
+            return statute[0:7]
+        return statute[0:6]
```
a-ilango: The method can be renamed to `__get_section` as it returns the section. Alternatively, Line 20 can be moved inside this method.

NickSchimek: Yes, Line 20 should be inside the method. Thank you. I don't agree that it should be renamed `__get_section` because this not a public method that other classes use to retrieve the section. This method sets the `statute` attribute during initialization, and after that it is not used. However, we could add a getter method that returns the `statute` attribute.

NickSchimek: After getting some sleep I realize the dilemma of calling it set section as it doesn't actually set the section inside the method

NickSchimek: perhaps get section is more appropriate, however I feel the naming of methods as get/set should be reserved for accessor methods... Although `__get_section` sounds the most appropriate of the two. What do you think of naming it `__extract_section`? Anyone else have any thoughts on this?

a-ilango: I don't think there's anything wrong in naming internal methods get/set. We do have leading underscores to the method names that tells users of this class' instance that they're internal methods. get_section / extract_section: either should be fine. 👍 

Also, Python has the @property decorator to implement accessor methods succinctly.

NickSchimek: Thanks I created an issue #150 to fix the renaming along with changing the return type on the method. Also I read up on the property decorator and that is very nice. It's exactly what I could use in some places of the code. Thanks for pointing that out.

PR #113, opened 2019-05-08
add data model section in design doc

wittejm:
This just starts a section for the data objects used by the expungeservice. 

Record doesn't currently exist as an explicit class but implicitly as just a list of cases, so is included here.  

Section is subject to change, but this is a start.

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -215,6 +215,45 @@ Returns:
     * tbd
 
 
+Data Model
+----------
+
+#### Record:
+ - cases :: type list[Case]
+
+#### Case:
+ - name
+ - birth_year
+ - case_number
+ - citation_number
+ - date
+ - location
+ - violation_type
+ - current_status
+ - balance_due_in_cents
+ - case_detail_link
+ - charges # type list[Charge]
+
+#### Charge:
+ - name
+ - statute
+ - level
+ - date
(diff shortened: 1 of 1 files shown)


doc/design.md
```diff
@@ -215,6 +215,45 @@ Returns:
...
+#### Case:
+ - name
+ - birth_year
+ - case_number
+ - citation_number
+ - date
+ - location
+ - violation_type
+ - current_status
+ - balance_due_in_cents
```
NickSchimek: Did we change this to `balance_due`?

NickSchimek:
I think it would be useful to add a Record object that the crawler returns

doc/design.md
wittejm: I left it as balance_due_in_cents, because calling both that and the getter/setter balance_due seems confusing and bug-prone

NickSchimek: OK. Thanks

wittejm:
> I think it would be useful to add a Record object that the crawler returns

Should I write a new issue for this? Looks like it would be a fairly minor change to add the new file for the class, and a line or two in the crawler. 

Also possible would be to change the RecordParser to return the Record object, which then gets populated by the Crawler with the Charges information. Then the Crawler would not be reading/writing to a field of the RecordParser object, as it does now.

PR #116, opened 2019-05-09
Issue 67: Initial commit of database connection class.

erikjasso:
@a-ilango @NickSchimek @wittejm or anyone else interested please review.


src/backend/expungeservice/database/database.py
```diff
@@ -0,0 +1,96 @@
...
+password = os.environ['POSTGRES_PASSWORD']
+from expungeservice.database import database
+
+# Get a connection.
+db = database.Database(host='db', port='5432', name='record_expunge',
+                       username='postgres', password=password)
+
+# Run a query.
+query = 'SELECT * FROM users;'
+db.cursor.execute(query, ())
```
a-ilango: Why do we pass an empty tuple as second arg?

wittejm: @a-ilango If there are any arguments for the SQL string, they're passed in as a tuple instead of using conventional string-formatting, for security against SQL injection. Strictly speaking the tuple is optional if there are no arguments.

a-ilango: Thanks!

PR #120, opened 2019-05-09
Create dependabot config file

### LINKED ISSUE #102: Create config file to update FE dependencies
## Use Case
As a maintainer, I would like to have dependabot update the Front End dependencies automatically.

## Description
Dependabot is already installed and the default configuration is to look for python related dependencies. However, if a config file is created we can also use it to update the front end Node dependencies. More info can be found here: https://dependabot.com/docs/config-file/

### This task is done when:
 - Config file exists to update frontend dependencies.

NickSchimek:
This directs dependabot to look for front end javascript dependencies as
well as the backend python dependencies.

@erikjasso @wittejm Are there docker dependencies that need to be maintained as well? If so where are they located?

Fixes #102

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -14,3 +14,9 @@ update_configs:
     default_labels:
     - "frontend"
     - "dependencies"
+
+  - package_manager: "docker"
+    directory: "/src/backend/expungeservice"
+    update_schedule: "daily"
+    default_labels:
+    - "dependencies"
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,16 @@
+version: 1
+
+update_configs:
+  - package_manager: "python"
+    directory: "/"
+    update_schedule: "live"
+    default_labels:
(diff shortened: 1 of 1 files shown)


NickSchimek:
@wittejm I added the config to update the dockerfile. Here is an explanation of what it does. https://dependabot.com/blog/dependabot-now-supports-docker/ Let me know if this is useful or not.

NickSchimek:
Regarding line 10 in the dockerfile the command `RUN python3.7 -m pip install -r requirements.txt && rm requirements.txt && python3.7 -m setup install` seems to install dependencies from the requirements.txt file, however we don't have a requirements.txt file as we're using pipenv. I assume the docker image is working and there are no issues, hence I am wondering what I am missing. Pipenv can generate a requirements.txt file if we need to https://docs.pipenv.org/en/latest/advanced/#generating-a-requirements-txt

NickSchimek:
Will merge this. If changes/additions/deletions need to be done we can do that

Issue #86, opened 2019-03-28
Add "Record Search" box to search page


maxkwallace:
See the upper "Record Search" card on http://dev.huntermarcks.net/search/ 

This task is only to implement a Component that corresponds to the "Record Search" box at the top of the page under the header. The Component should contain the First, Last, and DOB fields as well as the search button. You can add HTML validation using the `<form>` tag. 

Do not implement the code to actually process the search request and send it to the backend when the user clicks the search button. (That will be in a separate task.)

If the /search URL doesn't exist in React you don't need to create it. We likely don't need actual URLs for the app, but we can add them later if needed.  

### This task is done when...
 - A Component exists in React that matches the "Record Search" box visible on http://dev.huntermarcks.net/search/

hmarcks:
Here's a good resource for implementing accessible required fields. I'll update the HTML to help support this.

https://developer.paciellogroup.com/blog/2019/02/required-attribute-requirements

NickSchimek:
@hmarcks Is this completed?

hmarcks:
@NickSchimek yeah it's done, I'll close it

PR #121, opened 2019-05-10
Record search box

cjbreaux:
Pulled the HTML from http://dev.huntermarcks.net/search/. Haven't added any additional form validations, assumed it would be a separate task?


hmarcks:
Thanks @cjbreaux ! A few things:
1) Your branch is behind master, please pull so it's up-to-date.
2) We don't need the styles.scss file you created as we already have a couple SCSS files in the /src/styles directory. You'll also see that the .visually-hidden class is already included once you get your branch up-to-date.
3) This issue includes validation: https://github.com/codeforpdx/recordexpungPDX/issues/86. We can pull that part out if you'd like and give it its own issue.
4) The wrapper is missing on the search view - see the comparison to the prototype:
missing-wrapper

NickSchimek:
@hmarcks @cjbreaux How is react usually structured? It seems to me that the styles in `src/components/RecordSearch/styles.scss` is being used to encapsulate specific styles for the RecordSearch container which would be preferred as it is over having a global style? Is this correct?

NickSchimek:
or are there no global styles in react?

hmarcks:
@NickSchimek You can do it either way. Since we're mostly using Tachyons and other custom immutable utility classes I don't see an advantage to giving a component it's own styles.

cjbreaux:
The wrapper is included now and I've removed the local stylesheet. A form validation check is included for missing entries and correctly formatted date(allows for M/D/YYYY and M-D-YYYY). Not yet accounting for downed services. Appreciate any feedback as I'm still new at this!

NickSchimek:
I see, thanks for explaining @hmarcks Should we create a ticket to remove the component specific styles in `src/components` as their presence is contradictory to the coding pattern that is being suggested/enforced?

NickSchimek:
FYI: Forward slashes are the only acceptable date format the website accepts that we're crawling.

Screen Shot 2019-05-11 at 6 50 17 PM

cjbreaux:
Got it, will update accordingly.

hmarcks:
Awesome, thanks for the updates and tackling validations @cjbreaux ! Some feedback:
• Can we switch the `aria-invalid` attributes on the inputs to `false` when they're invalid? Example below.
• Can we also show the Date of Birth validation message along with the required field validation? Example below.

search-validations

@NickSchimek thanks for catching the date format requirement. Regarding the component-specific styles, yeah I agree we should just get rid of those, I'll create a ticket.

NickSchimek:
@hmarcks @cjbreaux I see all fields are required, but just wanted to point out the requirements the backend has for the search parameters, although it doesn't mean the frontend can't have them all required. Did Michael want them all required?

- `first_name`: required field
- `last_name`: required field
- `middle_name`: optional field
- `birth_date`: optional field

Just realized the middle name field never made it on any of the initial wireframes. I can't remember if @michaelatcommunitylaw needed that parameter available or not. However the backend can support it if it's something that is needed.

NickSchimek:
Thanks for working on this @cjbreaux . It's much appreciated!

hmarcks:
@NickSchimek Only requiring first_name and last_name could return many results for multiple people so I assumed we wanted to require both. If we match first_name + last_name + birth_date for multiple people, can we just return multiple sets of results?

Regarding middle_name, I remember asking Michael about what fields he wanted, but I'll confirm.

NickSchimek:
I had spoke to Michael at the beginning regarding the issue of results being not discrete. He said that he would only be inputting search arguments that would return results for one person.

A little background information: The system allows for fuzzy search so even if we require all three fields a user can still get results for multiple people. Therefore it is up to the user using the system to ensure proper search parameters. 

No, returning multiple sets of results is currently not possible. If we were querying an API or the results on the page were displayed differently then that may be possible. 
Perhaps we could assume everyone with a different name and birthdate are not the same people, but then issues arise due to aliases. However, Michael said we don't have to worry about the search results returning multiple people so that is what I'm going off of.

maxkwallace:
@hmarcks @NickSchimek my $0.02 on component-specific styles:

As Nick pointed out, because of encapsulation, if there are styles that are only used in one Component, it's better to put them in a `styles.css` files in `components/ThatComponent`.

But to Hunter's point, if a Component can be styled only with Tachonys and utility classes, that's better still. If this is the case, then I think there should be no component-specific `styles.css` file. But I don't think we should forbid those files; I suspect there will be some styles that are only used in one Component.

@hmarcks if there are existing component-specific styles that can be refactored into or replaced by global utility classes, I'm in favor of that, from looking at the existing `styles.css` files I'm not sure if that's the case.

Addendum: 

With component-specific you can get real encapsulation with [CSS Modules](https://facebook.github.io/create-react-app/docs/adding-a-css-modules-stylesheet) that guarantees the styles are not being used anywhere else. This helps a lot anytime you need to change existing CSS, since you don't have to search the whole project to make sure you're not accidentally breaking something.

(I thought CSS Modules was turned on; it's not now, but all we have to do to turn it on is rename `x.scss` to `x.module.scss`, which I'm happy to do. I just tested this and it works.)

hmarcks:
Thanks @maxkwallace , the prototype was created using only utility classes, so I'm confident we don't need any of the existing component style sheets and won't in the future.

When new styles are needed I'd like to continue using utility classes or BEM: http://dev.huntermarcks.net/design-system/#css-framework

I think this approach keeps things simple for a small app like this. We only have to manage a couple style sheets. Open to other approaches though!

maxkwallace:
@hmarcks great, let's move ahead with that approach then, and get rid of the component-specific styles!

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -2,14 +2,104 @@ import React from 'react';
...
+  state = {
+    firstName: null,
+    lastName: null,
+    dateOfBirth: '',
+    error: null,
+    validForm: false
+  }
+
+  handleChange = (e: React.BaseSyntheticEvent ) => {
+    console.log(e)
```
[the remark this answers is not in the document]
cjbreaux: Oops! Is there a linter set up for this project. If so, could you tell me how to run it? I'd like to avoid these kinds of errors in the future. Thanks!

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -2,14 +2,104 @@ import React from 'react';
...
+      //dispatch an action with local state data
+    }
+  }
+
+    //still need a case to handle downed services
+  validateForm = () => {
+    if (!this.state.firstName || !this.state.lastName || !this.state.dateOfBirth) {
+      this.setState({error: <p className="bg-washed-red mv4 pa3 br3 fw6 w-100">All search fields are required.</p>})
+    } else if (moment(this.state.dateOfBirth,'MM/DD/YYYY',true).isValid()===false) {
+      this.setState({error: <p id="dob_msg" className="bg-washed-red mv4 pa3 br3 fw6 w-100">The date format must be MM/DD/YYYY.</p>})
```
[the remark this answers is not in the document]
cjbreaux: I've refactored the validation method to separate the concerns (I think!)

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -2,14 +2,104 @@ import React from 'react';
...
+    if (this.state.validForm === true) {
+      //dispatch an action with local state data
+    }
+  }
+
+    //still need a case to handle downed services
+  validateForm = () => {
+    if (!this.state.firstName || !this.state.lastName || !this.state.dateOfBirth) {
+      this.setState({error: <p className="bg-washed-red mv4 pa3 br3 fw6 w-100">All search fields are required.</p>})
+    } else if (moment(this.state.dateOfBirth,'MM/DD/YYYY',true).isValid()===false) {
```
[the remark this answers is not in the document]
cjbreaux: I've refactored this and attempted to use proper spacing.

src/frontend/src/components/RecordSearch/index.tsx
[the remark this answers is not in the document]
maxkwallace: We don't have a linter set up yet! PRs to add a linter are welcome :) 

I just added an Issue for this: https://github.com/codeforpdx/recordexpungPDX/issues/141

maxkwallace:
@cjbreaux did you push your most recent changes? I don't see and updates in the diff.

cjbreaux:
@maxkwallace  The most recent changes should be pushed up now. Thanks for reviewing the code!

@hmarcks In this update, I'm not rendering the 'aria-invalid' attributes until form submission. Is that the correct way to handle this?

hmarcks:
@cjbreaux thanks for these updates!

We'll want the aria-invalid attributes rendered and set to false right away. This is to stop form controls from announcing as invalid by default. Here's an article with more details if you're interested: https://developer.paciellogroup.com/blog/2019/02/required-attribute-requirements

And the Date of Birth validation message functionality looks good, thank you.

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -2,14 +2,103 @@ import React from 'react';
...
+    e.preventDefault();
+    this.validateForm();
+    if (this.state.missingInputs === false && this.state.invalidDate === false) {
+      // Dispatch an action.
+    };
+  }
+
+  validateForm = () => {
+    !this.state.firstName ? this.setState({missingFirstName: true}) : this.setState({missingFirstName: false});
+    !this.state.lastName ? this.setState({missingLastName: true}) : this.setState({missingLastName: false});
```
maxkwallace: `!' '` is `false`. Is this intended behavior for when there is only whitespace in the form field?

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -2,14 +2,103 @@ import React from 'react';
...
+
+    !this.state.firstName || !this.state.lastName || !this.state.dateOfBirth ?
+    this.setState({missingInputs: true}) : this.setState({missingInputs: false});
+
+    moment(this.state.dateOfBirth,'MM/DD/YYYY',true).isValid() === false ?
```
maxkwallace: Should be

    moment(this.state.dateOfBirth, 'MM/DD/YYYY', true)

(spaces in between arguments in a list of positional arguments)

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -2,14 +2,103 @@ import React from 'react';
...
+    this.setState({invalidDate: true}) : this.setState({invalidDate: false});
+  }
```
maxkwallace: All of the statements in this method are of the form:

    [boolean expression] ? do_foo(true) : do_foo(false)

You can rewrite (and thus shorten/simplify them) as:

    do_foo([boolean expression])

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -2,14 +2,103 @@ import React from 'react';
...
+              </div>
+              <div className="w-100 w-10-ns mb3">
+                <button className="br2 bg-blue white bg-animate hover-bg-dark-blue db w-100 tc pv3 btn--search" type="submit">
+                  <span className="visually-hidden">Search Records</span>
+                  <i aria-hidden="true" className="fas fa-search"></i>
+                </button>
+              </div>
+              <div role="alert" className="w-100">
+                {this.state.missingInputs === true ? <p className="bg-washed-red mv4 pa3 br3 fw6">All search fields are required.</p> : null}
+                {this.state.invalidDate === true ? <p className="bg-washed-red mv4 pa3 br3 fw6">The date format must be MM/DD/YYYY.</p> : null}
```
maxkwallace: This looks way better-- awesome!

src/frontend/src/components/RecordSearch/index.tsx
cjbreaux: Well, the behavior was intended but my variable naming was unclear.  
`missingFirstName` -> `firstNameHasInput`
So, an empty form field  should set `firstNameHasInput: false`

src/frontend/src/components/RecordSearch/index.tsx
cjbreaux: I see! Thank you!

src/frontend/src/components/RecordSearch/index.tsx
cjbreaux: Good catch! I've rewritten them as suggested.

cjbreaux:
@hmarcks  That article was indeed helpful, thank you! I've tried to provide additional context with the aria-describedby attribute as well.

src/frontend/src/components/RecordSearch/index.tsx
maxkwallace: That's fine and you don't have to change the names! Just wanted to make sure since otherwise the validations would consider whitespace as valid, wasn't sure if that was intended behavior.

cjbreaux: Ahhhh, I see what you mean now.  This was not intended. I'll try to account for this case.

maxkwallace:
Is this good to merge? Would be ideal make sure everyone has these changes for the meeting tonight.

hmarcks:
Good from my POV, but I'll let @cjbreaux press the button

cjbreaux:
I'm in class now but will be at the meeting tonight, can it wait until
then?

hmarcks:
Certainly, seeya then

PR #123, opened 2019-05-13
Design doc section for database calls

wittejm:
Nothing complicated. Includes the functions needed for user and search endpoints, which I went over with @ash-tamraz and @a-ilango .

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -294,6 +294,10 @@ Return the password_hash associated with an email. The password hash can then be
 
 Check the Users table to see if a user with the given email already exists. Returns boolean. If query fails, throws an error.
 
+    check_is_admin(database, email)
+
+Check the Users table to see if a user with the given email is admin. Returns boolean. If query fails e.g. because user does not exist, throws an error.
+
     save_stats(database, email, record)
 
 Takes a Record object which has processed by the expungement analyzer. Saves only a limited amount of information. Details tbd, but here are a few examples for tracking app usage and app impact:
--- a/doc/design.md
+++ b/doc/design.md
@@ -274,3 +274,30 @@ Notes:
 
 - Store passwords encrypted
 - Use Postgres features for exporting JSON
+
+Database Functions
+--------------
+Functions that query or update the database will be organized into a single module in the expungeservice.
+
+    create_user(database, email, admin, hashed_password)
+
+Insert a new user into the Users table with the given email string and admin flag, and inserts the password hash into the Auth table and link it with the User uuid. The uuid of each, and the date_created and date_updated are generated within the database.
+
+This pair of inserts is an atomic operation, so if one fails then the other one is cancelled and has no effect.
+Returns None if insert is successful; otherwise throws an informative error.
+
+    get_password_hash(database, email)
+
+Return the password_hash associated with an email. The password hash can then be compared against a password in the calling code. If the user doesn't exist, an informative error is thrown. Otherwise, the function returns the password hash value.
+
+    check_user(database, email)
+
+Check the Users table to see if a user with the given email already exists. Returns boolean. If query fails, throws an error.
+
+    save_stats(database, email, record)
+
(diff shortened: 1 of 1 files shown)


doc/design.md
```diff
@@ -274,3 +274,32 @@ Notes:
...
+
+Check the Users table to see if a user with the given email already exists. Returns boolean. If query fails, throws an error.
+
+    get_password_hash(email)
+
+Return the password_hash associated with an email. The password hash can then be compared against a password in the calling code. Assumes the user exists, and throws an error otherwise.
+
+Performing login is two steps: check the user exists, and if so then compare the password.
+User creation is also two steps: check the user doesn’t exist yet, and if not then create the user.
+
```
NickSchimek: What is the reason for making the user login and creation a two step process?

wittejm: That's true, and the more I think about it the more I like the idea of each being a single database operation that uses error-handling to parse the different outcomes. Arun and I talked about this briefly and were thinking that it might be simpler to break each down into the two pieces of logic. I forget if there were other reasons for this. @a-ilango , your thoughts?

wittejm:
@NickSchimek Adjusted the spec a bit so that both user creation and password fetch use only one database operation. There is no need to check an email address is already in use before either operation because relevant error messages can be thrown instead. 
I also added database object as an argument for each of these functions. @erikjasso  had suggested we provide a reference to the database object using a feature of flask that imitates a global namespace. But since each of these functions will just get called in endpoint code, and each endpoint call will initialize the database objects, I think it's a more robust design to just pass in the database object as a reference.

doc/design.md
```diff
@@ -274,3 +274,34 @@ Notes:
...
+Database Functions
+--------------
+Functions that query or update the database will be organized into a single module in the expungeservice.
+
+    create_user(database, email, admin, hashed_password)
+
+Insert a new user into the Users table with the given email string and admin flag, and inserts the password hash into the Auth table and link it with the User uuid. The uuid of each, and the date_created and date_updated are generated within the database.
+
+This pair of inserts is an atomic operation, so if one fails then the other one is cancelled and has no effect.
+Returns None if insert is successful; otherwise throws an informative error.
```
NickSchimek: From working with other ORM's I'm used to seeing it return an object, in this case a user object (or True), if it succeeds; otherwise an error (or a falsey value). That way the endpoint code would be something like this:
```
if create_user(database, email, admin, hashed_password):
    process_request
else:
     return to user creation screen with failure notification
```
or do you have something else in mind?

NickSchimek:
I can't remember if there is a reason we’re NOT implementing all four basic crud operations for user and login (auth table)? I can’t recall if they’re part of the user stories or not. I'm thinking even if there is no immediate plan to have them as endpoints perhaps it would be good to have them available as part of the ORM.

doc/design.md
wittejm: @NickSchimek Yes, returning True makes intuitive sense, to make your use case syntax work. Though when I thought through possible sources of failure, it seemed like it never makes sense to return False; it should just throw the specific error (Since an error object isn't falsey). Making the "True" return value technically spurious. But I think it works equally well so there's no reason not to change this to returning True.

For returning a User object: again this would also be fine I think, but the caller already has the email, password_hash, and created/modified time values; the only thing they don't have is the UUID which I don't know is useful in an endpoint?

wittejm:
@NickSchimek  I talked with @erikjasso  about doing ORM/CRUD design and we agreed that it wasn't important enough to do within the current project specification, since the set of database ops required currently (and in the foreseeable future) is a very small subset of CRUD, so we wanted to avoid doing the engineering overhead though that may change at some point.

doc/design.md
NickSchimek: @wittejm Agreed!, I think it's not so much as what the caller knows already as to the flexibility of the ORM. However I'm fine either way: True/False, True/error, Object/False, Object/error. <- there are use cases for each one of the scenarios, but I would advocate to only build what we need while allowing it to be open for extensions in the future.

NickSchimek:
OK Thanks! please proceed 🙂

Issue #141, opened 2019-05-17
Add linting to frontend React app


maxkwallace:
Right now we don't have linting for our frontend / React app which means we end up spending time in PR reviews addressing style issues that the linter could handle automatically. Adding a linter will speed up development time for everyone.

create-react-app comes with eslint but since we are using Typescript, and we need a Typescript-compatible linter, this will be a bit more complicated. Looks like a good place to start is https://github.com/facebook/create-react-app/issues/5641

### This task is done when:
 - We have a linter set up in the app that and matches the existing code style rules set out in the wiki: https://github.com/codeforpdx/recordexpungPDX/wiki/Frontend

maxkwallace:
Completed with #147

PR #147, opened 2019-05-23
Add Prettier for code formatting (addresses style linting)

maxkwallace:
Addresses #141 

See:
https://facebook.github.io/create-react-app/docs/setting-up-your-editor#displaying-lint-output-in-the-editor

create-react-app includes ESLint by default which can't be customized unless we eject from the CRA defaults-- which is probably a bad idea.

So for now, let's follow the recommendations on that page and add Prettier and see if that helps speed up our PR review process.

If we decide to / need to customize ESLint directly in the future, see:
- https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project
- https://github.com/typescript-eslint/typescript-eslint

For more info on Prettier, see https://prettier.io/ -- the style rules it applies are intended to be non-configurable defaults that are already in use by a majority of the Javascript community.


ForrestLonganecker:
I use eslint and prettier in my projects and enjoy them both as time savers. These changes look good to me.

NickSchimek:
This will automatically work with the 8 editors listed on their homepage?

maxkwallace:
@NickSchimek not automatically AFAIK, you have to install the editor packages/extensions, but those are totally optional.

Even if you don't do anything different in your editor, this commit configures Prettier to run as a pre-commit hook that automatically reformats any staged js,jsx,ts,tsx,json,css,scss,md files when you commit (it's quite fast), so no one has to do anything differently. If the pre-commit hook is annoying to anyone I can look into configuring it differently.

PR #151, opened 2019-05-27
Database tables and doc update

wittejm:
Password salt is not stored separately in the database because it is encoded as part of the password hash. 
Also just a minor doc update to the database schema to match what is defined in create_tables.sql


wittejm:
Ugh some unintended changes got included here; will close and re-submit.

PR #152, opened 2019-05-27
remove password salt from auth table; update doc to match sql schema

wittejm:
Password salt is not stored separately in the database because it is encoded as part of the password hash.
Also just a minor doc update to the database schema to match what is defined in create_tables.sql


wittejm:
Such a tiny change that I'll just merge this later tonight if there are no comments

PR #153, opened 2019-05-28
update package requirements in Docker's setup.py

wittejm:
The docker build of the expungeservice was broken because this file listed the now-missing "test" package. Also updating the file to include our two new packages. 

Relatedly, I'm thinking we should add a script to check the docker build succeeds on a new pull request. I'm not clear on how to do this; can create a new issue for it though.

### DIFF
--- a/src/backend/expungeservice/setup.py
+++ b/src/backend/expungeservice/setup.py
@@ -6,9 +6,10 @@ setup(
     version='0.1',
     description='recordexpungPDX Expunge Service',
     packages=['crawler',
+              'database',
               'endpoints',
               'expunger',
-              'test'],
+              'models'],
     setup_requires=[],
     package_dir={}
 )


NickSchimek:
Adding the docker build to the CI so that the tests are ran inside the docker build would solve this. That's something we can work on next meeting. I've done it with CircleCI before... Will have to read up on how to do it with Travis

src/backend/expungeservice/setup.py
```diff
@@ -6,9 +6,10 @@
     version='0.1',
     description='recordexpungPDX Expunge Service',
     packages=['crawler',
+              'database',
               'endpoints',
               'expunger',
-              'test'],
+              'models'],
```
NickSchimek: Is this something we can do dynamically instead of hardcoding the values in?

wittejm: I'm not familiar enough with packaging to know if there is a conventional dynamic solution, but I'm happy to look into it. Meanwhile I'd like to just merge this for now.

Issue #118, opened 2019-05-09
Database authorization refactor


wittejm:
The database is currently with a postgres user-password that is saved in the git repo. This refactor will require the developer to instead define their own password, which will not be checked into the reop.
  

### This task is done when:
 - password in config/postgres/postgres.env  is replaced with something like "changeme"
 - postgres.env file added to .gitignore
 - existing postgres.env file renamed to something else that can be checked in. 
 - ensure that database build fails (e.g. by the missing env file) until user sets a password.

PR #160, opened 2019-06-02
Implementation of database Create and Read User operations, test cases

wittejm:
Design notes:

- I realized the functions I had describe in the design doc could just be rolled into a single Create and a single Read op. Starts to look like CRUD after all, nice. The project spec doesn’t currently have a need for Update or Delete, though those could clearly be useful to have at some point in the future.  

 - create_user now returns the User object that it created. Taking Nick’s suggestion, and it seemed like there was no reason not to do this.

 - There is no input-checking for email format, password length, etc. I think that stuff belongs in the endpoint code.

 - Error-handling: psycopg throws the database errors for duplicate emails, etc. I considered catching these and throwing a custom error object but don’t know if there is a good reason to do this, so I decided to just propagate the errors that psycopg generates. It all just goes to endpoint code. 

 - If the create_user op does throw an error, it calls connection.rollback() before propagating the error. This seems to be enough to keep the cursor object running smoothly after attempting a failed operation. I didn’t wrap the read-only function with this behavior.

Testing notes:

 - To make a database connection, the test class needs to use the password that has been defined in db configuration; currently it uses a relative path to the .env file in the config folder. I'm not sure if there is a nicer way to do this. 
EDIT: I decided to rely on pipenv to pick up a local .env file, and I added a .env.example file to copy to .env. This should be refactored later as part of issue #118 .
The database credentials in this file should be made to match those in config/postgres/postgres.env , except for the hostname being different ("db" in docker vs "localhost" in pipenv). That change again should be part of #118  .
I'm not totally sure that repeating (most of) the same database connection creds in two files is the best approach. I welcome opinions.

 - The test class also accesses the database at localhost. To make this work, I had to expose the database’s network port (in docker-compose.dev.yml), which, like the exposure of the expungeservices port, is only necessary in Dev. This seems fine to me. Having the port exposed is also nice for interactive testing. The other clean way to access the database with test cases would be from within another Docker container in the project stack. There might other reasons to create a dedicated “Test” Docker image at some point.

 - Note that when updating the project for testing you need to drop the database (using the dev_drop_database make target) before rebuilding the Docker stack, to let the database changes take effect. I totally forgot about this for a while.


wittejm:
Ah, build failed because the password didn't get read from the .env file. I'll look into this but meanwhile would appreciate feedback on the rest.

NickSchimek:
@wittejm I see your having issues using environment variables... Using the `.dotenv` library should work. Were they working for you locally??? We can specify environment variables in Travis CI so that it doesn't bomb out whenever the code refers to them. I will see if I can add you to admin the Travis CI so that you can add them; otherwise you can dm me the env variables.

wittejm:
@NickSchimek Looks like I can already set env variables on Travis. Nice to know :). So that issue is out of the way once the database can run on Travis. It's working for me locally just fine, I'm just uncertain what the best solution is for maintaining the dual configurations of pipenv (including for pytest), and the Docker environment. I think this can be a quick thing to work out at next meeting.

wittejm:
I added the Docker stack to Travis (starting issue #165 ) to get the database running and testable there, and now this build is passing all tests. 
Things I don't like about the current setup but don't have a preferred solution at the mo:
 - The .travis.yml copies .env.example to .env so that pipenv picks up the variables. I like this better than setting env variables in Travis because this way the variables for running pipenv are defined only once. But I'm looking into a more conventional solution.
 - After launching the docker stack, the script sleeps until the stack is fully spun up before running pytest. The in-line sleep loop in the travis file feels hacky? Is there some best practice for this? 
 - The build now takes ~4 minutes or so because the docker image builds take a while. I think there is a way to cache unchanged image layers in Travis to speed this up. I'm starting to look into this.

config/postgres/initdb/scripts/create-tables-and-functions.sql
```diff
@@ -60,3 +60,21 @@ CREATE TABLE analyses (
     user_id        UUID NOT NULL REFERENCES users(user_id) ON DELETE CASCADE,
     PRIMARY KEY (client_id, case_id)
 );
+
+CREATE FUNCTION CREATE_USER( email citext, admin boolean, hashed_password text)
```
NickSchimek: Should the function name be lower_case?

NickSchimek: Also what happens if the user of the function does not pass all three parameters? I'm not familiar with SQL, but it seems `admin` is a required parameter. Can we make it optional? It should automatically default to `FALSE` if it's not specified.

NickSchimek:
Other than two questions above ^^^ this looks good to me. Let's give @erikjasso a chance to look over this. Thanks @wittejm !

config/postgres/initdb/scripts/create-tables-and-functions.sql
wittejm: Yes, it looks like the convention is to match the format of function names and table names, which we're making lowercase where the table is created and uppercase where it is referenced, so I changed the create_user function declaration to be lowercase. 

Making the admin param optional sounds good too. Done.

Issue #165, opened 2019-06-04
Add docker stack to Travis-CI build


wittejm:
Testing the docker stack is a necessary add to the testing pipeline and has core support on Travic-CI:
https://docs.travis-ci.com/user/docker

The database test cases need to have the docker database service running in order to pass. Here's a resource on running tests cases from outside a running docker container: 
https://mike42.me/blog/how-to-set-up-docker-containers-in-travis-ci

I think also the set of backend tests should run and succeed within the expungeservice container. This will mean loading the test code into that docker image for Dev but not for Prod. 

### This task is done when:
 - Docker build and deploy steps added to .travis.yml
 - Database operation test cases in pipenv succeed on Travis-CI against a running database.
 - Implement the test code deployment to dev but not prod. (and those test cases succeed)

wittejm:
I'm looking into the possibility of caching Docker layers to speed up the CI build, since build time has gone from ~1 to ~4 minutes. It doesn't look like there is a great solution. 

This thread has useful discussion about the problem:
https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-195226941 

This guide addresses a related problem, but only applies to caching completed images: Namely, push your images to the Docker public hub and then pull them into the Travis-CI build.
http://atodorov.org/blog/2017/08/07/faster-travis-ci-tests-with-docker-cache/

That doesn't perform the kind of layer-wise caching that would actually speed things up for us. 

I guess we're stuck with the longer build time? ¯\\_(ツ)_/¯

wittejm:
This issue will be closed with PR #183 . The feature of adding test code to the dev expungeservice container has moved to #174

wittejm:
Closed with PR #183 merge.

PR #166, opened 2019-06-05
build and deploy docker stack in Travis-CI

wittejm:
Expands the .travis.yml to build and deploy the Docker stack before running the pytest battery. 

I'm making this a separate PR even though it's already feature of #160 . But this is pretty simple and standalone, and will be useful to have in the pipeline independent from finalizing the database features in PR #160. 

This is a WIP and partial contribution to issue #165.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,12 @@ addons:
       - docker-ce
 before_install:
 - docker swarm init
-- cp ./.env.example ./.env
 install:
 - pip install pipenv
 - pipenv install
 - make dev
 before_script:
-- counter=0 ; while ((counter < 10)) && ! pg_isready --host=$PGHOST --port=$PGPORT --dbname=$PGDATABASE ; do sleep 3; let counter=counter+1; done
+- counter=0 ; while ((counter < 20)) && ! pg_isready --host=$PGHOST --port=$PGPORT --dbname=$PGDATABASE ; do sleep 3; let counter=counter+1; done
 script:
 - pipenv run pytest
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,21 @@ language: python
 sudo: required
 dist: xenial
 python: 3.7.0
+services:
+- docker
+addons:
+  apt:
+    packages:
+      - docker-ce
+before_install:
+- docker swarm init
+- cp ./.env.example ./.env
 install:
 - pip install pipenv
 - pipenv install
+- make dev
+before_script:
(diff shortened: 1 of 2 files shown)


wittejm:
Things I don't like about the current setup but don't have a preferred solution at the mo:
 - The .travis.yml copies .env.example to .env so that pipenv picks up the variables. I like this better than setting env variables in Travis because this way the variables for running pipenv are defined only once. But I'm looking into a more conventional solution.
 - After launching the docker stack, the script sleeps until the stack is fully spun up before running pytest. The in-line sleep loop in the travis file feels hacky? Is there some best practice for this? 
 - The build now takes ~4 minutes or so because the docker image builds take a while. I think there is a way to cache unchanged image layers in Travis to speed this up. I'm starting to look into this.

.travis.yml
```diff
@@ -2,9 +2,21 @@ language: python
...
 python: 3.7.0
+services:
+- docker
+addons:
+  apt:
+    packages:
+      - docker-ce
+before_install:
+- docker swarm init
+- cp ./.env.example ./.env
```
NickSchimek: This shouldn't be necessary. Can we add the environment variables to Travis CI?

.travis.yml
```diff
@@ -2,9 +2,21 @@ language: python
...
 install:
 - pip install pipenv
 - pipenv install
+- make dev
+before_script:
+- counter=0 ; while ((counter < 10)) && ! pg_isready --host=$PGHOST --port=$PGPORT --dbname=$PGDATABASE ; do sleep 3; let counter=counter+1; done
```
NickSchimek: It seems this is a known issue in the travis community... https://github.com/travis-ci/travis-ci/issues/4842

NickSchimek:
Thanks for doing this @wittejm . Are you able to add the environment vars to Travis Ci https://travis-ci.com/codeforpdx/recordexpungPDX/settings ?

.travis.yml
wittejm: Sure we can do it that way

wittejm: Updated and committed. Added the variables in Travis and everything works fine.

.travis.yml
wittejm: Yep, look like some folks there used the same sort of idling loop as a workaround. Nice to feel like this solution is validated. (seems like a problem that Travis should fix though :) )

wittejm:
@NickSchimek Yep, I was able to add env variables to Travis and now the .env file is unnecessary.

.travis.yml
NickSchimek: yeah, agreed. We could always switch to Circle CI or another CI service if this causes issues in the future. I've used Circle CI with all my other projects, but thought I'd change things up a bit because I got a free sticker from them...

PR #175, opened 2019-06-06
Add default value to birth year

### LINKED ISSUE #167: Birth year should default to an empty string
## Description
Birth year is not present on all case files and we're seeing some failures here: https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/case.py#L7 

Need to update this so that Birth year defaults to an empty string when it is not present.

### This task is done when:
 - Create test case to expose bug (probably two tests: unit and functional)
 - Implement fix

NickSchimek:
Added test fixture with missing birth year, and created a system test to
recreate the bug being seen.

Created CaseFactory build option for manipulating distinct attributes
before initializing a case object which will allow explicit case tests.

Added the unit test to check the default birth year value.

Added default value for birth year

Fixes #167


wittejm:
Would it be worth it to add a new Record fixture that is only used to test for missing birth year? I liked the default RECORD fixture that had nothing wrong with it and I'm wary of overloading its purpose to include this test case. 

The new fixture could explicitly be called RECORD_MISSING_BIRTH_YEAR

Other than that, I think this looks good!

NickSchimek:
Thanks for reviewing @wittejm I was initially going to add a new record, but on a whim decided not to. I'd say technically the default record does not have anything wrong it, and it now better represents the real world. However, I'm fine doing it either way, and I think adding a new fixture `RECORD_MISSING_BIRTH_YEAR` would be more explicit.

NickSchimek:
@wittejm I added the new fixture. Squashed my commits and pushed up.

PR #185, opened 2019-06-08
Create a db connection upon incoming request

### LINKED ISSUE #173: Create a database connection for an incoming request
Using the flask "global" namespace:

`from flask import g` 

Use the @before_request decorator to instantiate a database connection, and attach it to:

`g.database`

Read connection info and credentials from environment variables to create the [Database](https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/database/database.py) object. 

Use @after_request to close the connection. 

Add the new functions to [app.py](https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/app.py)

### This task is done when:
 - g.database object accessible by endpoint code
 - the database connection is closed after a request is completed
 - relevant unit tests created

wittejm:
Created a new `request` submodule to store the `before` and `teardown` functions which are registered with the Flask app to run before and after each incoming request. I like this organization as opposed to decorating a get_connection() function with the @before_request decorator, because this explicitly keeps all of the before_request and teardown_requestion behavior in one function. There might be additional setup/teardown behavior needed later on.  

A drawback to triggering a db connection in this way is that a database connection is created even if the current request doesn't necessarily need one. In practice, all of our api endpoints do use the database, though this may change in the future. 

closes #173

### DIFF
--- a/src/backend/expungeservice/__init__.py
+++ b/src/backend/expungeservice/__init__.py
@@ -3,3 +3,4 @@ from .app import create_app
 from . import database
 from . import endpoints
 from . import expunger
+from . import request
--- a/src/backend/expungeservice/app.py
+++ b/src/backend/expungeservice/app.py
@@ -1,12 +1,10 @@
-import importlib
-import os
-
 from flask import Flask
 
 from .config import app_config
 
 # Add new endpoint imports here:
 from .endpoints import hello, auth, users, protected
+from .request import before, teardown
 
 def create_app(env_name):
     """
@@ -23,4 +21,6 @@ def create_app(env_name):
     users.register(app)
     protected.register(app)
 
+    app.before_request(before)
+    app.teardown_request(teardown)
     return app
--- a/src/backend/expungeservice/request/request.py
+++ b/src/backend/expungeservice/request/request.py
@@ -1,0 +1,22 @@
+from flask import g
+import os
+from expungeservice.database import Database
+
+
+def before():
+
+    host = os.environ['PGHOST']
+    port = os.environ['PGPORT']
+    name = os.environ['PGDATABASE']
+    username = os.environ['PGUSER']
+    password = os.environ['POSTGRES_PASSWORD']
+
+    g.database = Database(
+        host=host,
+        port=port,
+        name=name,
+        username=username,
+        password=password)
+
+def teardown(exception):
+    g.database.close_connection()
--- a/src/backend/tests/app/test_flask_app.py
+++ b/src/backend/tests/app/test_flask_app.py
@@ -1,0 +1,27 @@
+import unittest
+import expungeservice
+from flask import g
+
+
+class TestFlaskApp(unittest.TestCase):
+
+    def setUp(self):
+        self.app = expungeservice.app.create_app('development')
+        self.test_client = self.app.test_client()
+
+
+    def test_get_database_connection(self):
+
+        with self.app.app_context():
+
+            expungeservice.request.before()
+
+            assert g.database
+
+            query = 'SELECT * FROM users;'
+            g.database.cursor.execute(query, ())
+            rows = g.database.cursor.fetchall()
+            assert rows or rows == []
+
+            expungeservice.request.teardown(None)
+            assert g.database.cursor.closed
--- a/src/backend/expungeservice/request/__init__.py
+++ b/src/backend/expungeservice/request/__init__.py
@@ -1,0 +1,1 @@
+from .request import *


NickSchimek:
Thanks for weighing the pros and cons. This approach sounds good to me.

src/backend/expungeservice/__init__.py
```diff
@@ -3,3 +3,4 @@
 from . import database
 from . import endpoints
 from . import expunger
+from . import request
```
NickSchimek: It's generally good style to leave a newline at the end of the file. Not important, but thought I'd mention. You can setup your editor to automatically add a newline and remove trailing whitespace. You can also setup git to add newlines at the end of file and remove trailing whitespace. 

Screenshot 2019-06-08 19 22 32

src/backend/expungeservice/app.py
```diff
@@ -1,12 +1,12 @@
 import importlib
```
NickSchimek: This is currently an unused import... Perhaps we can remove it at this time?

src/backend/expungeservice/request/request.py
```diff
@@ -0,0 +1,26 @@
...
+        host=host, 
+        port=port, 
+        name=name, 
+        username=username, 
+        password=password)
+
+def teardown(exception):
+    try: 
+        g.database.close_connection()
+    except: pass
```
NickSchimek: I don't think this is a good idea as it will pass on all exceptions. Do you think we can get this under test, and catch the specific exception's?

src/backend/expungeservice/request/request.py
```diff
@@ -0,0 +1,26 @@
...
+
+    if exception: raise exception
```
NickSchimek: I'm not familiar with exceptions being passed to teardown, and what the flasky way is for dealing with them. Do you have a good reference I can read regarding this?

src/backend/tests/app/test_flask_app.py
```diff
@@ -0,0 +1,35 @@
...
+from flask import g
+
+
+class TestFlaskApp(unittest.TestCase):
+
+    def setUp(self):
+        self.app = expungeservice.app.create_app('development')
+        self.test_client = self.app.test_client()
+
+    def tearDown(self):
```
NickSchimek: This isn't necessary if we're not using it. Do you think we can remove it?

src/backend/tests/app/test_flask_app.py
```diff
@@ -0,0 +1,35 @@
...
+            
+            query = 'SELECT * FROM users;'
+            g.database.cursor.execute(query, ())
+            rows = g.database.cursor.fetchall()
+            assert rows or rows == []
+
+            expungeservice.request.teardown(None)
+            assert g.database.cursor.closed
+            
+    def test_hello_request(self):
```
NickSchimek: This is a good test to have in the endpoint directory. I don't think it belongs here. Plus it will be breaking when remove the hello endpoint in the future. Can we get it moved to the `tests/endpoints/hello`?

src/backend/expungeservice/request/request.py
NickSchimek: If an exception does occur while trying to close the connection perhaps we can log it? I'm not familiar with how to do that in a python/flask environment, are you?

src/backend/expungeservice/__init__.py
wittejm: Gotcha. I'll fix that and look into the automatic solution for the future, thanks!

src/backend/expungeservice/app.py
wittejm: yep.

src/backend/tests/app/test_flask_app.py
wittejm: Sure. I put it in with the assumption that it might get populated later, but that makes more sense to remove it until it's needed.

src/backend/tests/app/test_flask_app.py
wittejm: I did hesitate to put this in, but I wanted a test that verifies the connection succeeds during an actual endpoint call... um but this test doesn't explicitly do that (only verifies that an error isn't thrown during the attempt). Is it is indeed a useless test since it only tests a toy feature. The other tests on endpoint code definitely belong elsewhere. I will remove this.

src/backend/expungeservice/request/request.py
wittejm: I can redo this to just check there is an open connection to be closed, without a try-except block. The only failure conditions that occurred to me are that the connection didn't get opened for some reason which is easy to check for. My thinking here was that it if the attempt fails then the connection will eventually close from python garbage collection anyway (trying to close it explicitly just seems like a good practice). 

I don't know logging practices in flask, but that's definitely something worth figuring out.

src/backend/expungeservice/request/request.py
wittejm: Teardown functions in flask run regardless of whether a request succeeded, and may get passed an exception, whereas after_request functions only run if the request succeeded. So any necessary exception-handling belongs in teardown. 

I confess that I didn't look this through very carefully, but it looks like there is no need to re-raise errors that aren't handled in teardown. Unhandled errors are already raised to the server context. Some documentation: http://flask.pocoo.org/docs/0.12/reqcontext/#callbacks-and-errors

I'll remove the raise command here.

wittejm: In fact, the Database object @erikjasso wrote has a close_connection() function that it looks safe to just call without extra condition-checking.

src/backend/expungeservice/request/request.py
wittejm: In fact, the Database object @erikjasso wrote has a close_connection() function that it looks safe to just call without extra condition-checking.

wittejm:
@NickSchimek Thank you for all the feedback! I made changes based on your comments.

src/backend/expungeservice/request/request.py
NickSchimek: Thanks for looking into that

src/backend/expungeservice/request/request.py
NickSchimek: Thanks for explaining, and looking into it further

src/backend/tests/app/test_flask_app.py
NickSchimek: ok

src/backend/tests/app/test_flask_app.py
```diff
@@ -0,0 +1,27 @@
...
+        with self.app.app_context():
+
+            expungeservice.request.before()
+
+            assert g.database
+
+            query = 'SELECT * FROM users;'
+            g.database.cursor.execute(query, ())
+            rows = g.database.cursor.fetchall()
+            assert rows or rows == []
```
NickSchimek: I'd be tempted to change this to just `assert rows == []`. The reasoning behind this is that before every test case is ran the database should be clean. Currently `assert rows` is falsey, and it evaluates `rows == []`. In the future if the database does not get cleaned before the test run and there are users in the table then `assert rows` will evaluate to true, however if its changed to `assert rows == []` then this will be a good indicator to let us know that the database is not being cleaned before each test run.  What do you think?

wittejm: The reason I wanted to allow it to pass on a non-empty result is that a user might have manually put in some rows that don't need to be cleaned up by/for pytest. I've been doing this when poking at the database interactively. This test doesn't distinguish the result, only that a generic hit to the database worked.

NickSchimek: Understood. Ideally we would have a separate test database, development database, and production database. I've worked with apps that have tests that depend on data already being in the test database, and it's an OK approach to take, but it's been my experience that it causes unexpected issues. Just something to keep in mind.

PR #189, opened 2019-06-08
Set empty string as default value for section attr

### LINKED ISSUE #150: Return empty string for section method
## Description
This method should return an empty string instead of None when the section cannot be determined.
https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/charge.py#L95
Having None can throw an exception here when checking if the section contains a digit. 
https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/charge.py#L79

Also rename the method to something more appropriate such as `_get_section` or `_extract_section`
### This task is done when:
 - Add test that causes a charge to throw an exception at line 40
 - Change return type to an empty string
 - rename method

NickSchimek:
This is being done because isDigit() method cannot be called on
NoneType.

Fixes #150

### DIFF
--- a/src/backend/expungeservice/models/charge.py
+++ b/src/backend/expungeservice/models/charge.py
@@ -92,7 +92,7 @@ class Charge:
     def __set_section(statute):
         statute = Charge.__strip_non_alphanumeric_chars(statute)
         if len(statute) < 6:
-            return None
+            return ''
         elif statute[3].isalpha():
             return statute[0:7]
         return statute[0:6]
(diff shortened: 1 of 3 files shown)


NickSchimek:
Yes, I was a bit hesitant to add it, but it was necessary to expose the bug. When I do my next PR I will be adding test cases for parking tickets, and then at that time this test case can be removed since parking tickets have two digit statutes.

PR #190, opened 2019-06-10
Added install info for docker stack to readme

wittejm:
Just a minimal addition: install links for Docker and the single make command that builds our docker stack. I linked here the longer document that I had written for our Docker setup also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,13 @@ Follow installation instructions in:
 [Ubuntu Installation](https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-using-the-repository)
 Links to install instructions for other linux distros listed in the sidebar.
 
+9. Create a local .env file
+
+While in the directory of your local repo, run:
+```
+cp .env.example .env
+
+```
 ## Running Components
 
 ### Backend
--- a/README.md
+++ b/README.md
@@ -15,20 +15,20 @@ This README is primarily for the backend. For info on the frontend see [src/fron
 - [License](#license)
 
 ## Installation
- 
-1. **[Fork](https://help.github.com/articles/fork-a-repo/#fork-an-example-repository)**, 
+
+1. **[Fork](https://help.github.com/articles/fork-a-repo/#fork-an-example-repository)**,
 	and **[clone](https://help.github.com/articles/fork-a-repo/#step-2-create-a-local-clone-of-your-fork)** the repo.
 
 2. Install Python:
 
-    * **Mac** 
-    
+    * **Mac**
+
       To install the latest version of Python on Mac run:
-		
+
           $ brew install python3
 
       <!-- -->
-		note: This will pull the latest version of Python, so when Python 3.8 or 
+		note: This will pull the latest version of Python, so when Python 3.8 or
 		greater is released it will install that version.
 
     * **Linux**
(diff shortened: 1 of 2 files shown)


wittejm:
Requesting reviews from a variety of folks since I want to check these few steps are sufficient to get the complete dev setup working (pytest now fails without a working Docker stack).

NickSchimek:
Why does pytest fail without a working docker stack?

wittejm:
It's the database test cases. They fail if the docker database service isn't running. So just those, not the whole test run :)

NickSchimek:
You don't need to docker to have the database service running. I have postgres setup locally running the record_expunge database without any issues.

wittejm:
Woops, forgot the step to copy the .env file into place. I don't like this approach exactly because the .env file has some contents redundant with https://github.com/codeforpdx/recordexpungPDX/blob/master/config/postgres/postgres.env

But deciding the correct approach to this I think is a bigger task than this minor change to provide easy setup instructions, and is part of issue #118

wittejm:
That's fair, and we could have the pytest database tests succeed that way, but there are no install instructions to set up postgres to run locally -- that's also not the setup that is deployed in Travis or will deploy to prod. I never did that setup myself because I joined the project after the Docker database service was implemented. It makes sense to me to deploy the database only once. What do you think?

NickSchimek:
Sorry, it was decided early on that we would develop locally, and use docker for deployment. Yes, only one deployment... To me there are a few different paths that we can take. 
Have setup/install instructions for both local & docker environments: Giving the developer an option to choose.
or
Have setup/install instructions only for docker.

Thanks for creating this issue to start this discussion. 

I've used docker a few times (in total less than a few hours) in the past, but haven't done any serious developing on it. I know of the pro's and con's, and will need to try out docker a few times more to form more.

NickSchimek:
@codeforpdx/team-record-expungment Please provide your input if we should have install instructions for both local and docker, or have everyone who joins the project develop on docker? Docker would make the install progress easier, but I'm not sure if it would scare people away... What do you all think?

maxkwallace:
I feel strongly that we should have "one true way" to do things, and provide install instructions, documentation, and support for that way only ("convention over configuration"). If someone wants to do something different, maybe that's okay (?) but we shouldn't support that officially.

I haven't used Docker yet, so I don't really trust myself to form an opinion about whether or not we should use it in dev. I skimmed some articles online but they didn't help much. I think it ultimately comes down to whether or not using Docker in dev would reduce the amount of time folks spend debugging config and environment issues. I can install it and get it set up at the next meeting, and would be able to offer more of an opinion then.

wittejm:
Thanks for the thoughts @maxkwallace . I agree with what you're saying; my hope is that the documentation I'm adding here is enough to get the docker stack running easily (and that's the feedback I'm looking for specifically) and ideally it's something to run and mostly forget about during development. I may be biased and optimistic because I became comfortable messing around with a docker stack through weeks of building out parts of it. 

Unfortunately I won't make it to next week's meeting so I will miss whatever conversation happens there, but will be eager to hear what comes from it.

wittejm:
I added a note on using `docker swarm init` based on @gingerben93 's feedback. I think this is ok to merge to provide the install instructions, regardless of whether we expect each local developer to use the docker vs native database install. Thoughts, @NickSchimek ?

PR #201, opened 2019-06-13
Set birth year given multiple formats

### LINKED ISSUE #200: extract birth year when given a date in the format of mm/dd/yyyy
## Description
When the user performs a search with entering in the birth date then the response lists the birth year of the person on each case, however when the search is done with the birth date included then the response lists the birth date on each of the record. Currently the case is expecting a birth year only. Fix the case initialize the birth year when either format is encountered.

### This task is done when:
 - Add tests to expose bug
 - Birth year attribute can be initialized when given either format

NickSchimek:
Added test case for birth year 'mm/dd/yyyy' format
When the user searches for a record passing in the birth date format
then the response returned also includes that birth date instead of the
default value (birth year).

To set the birth year the argument is split on '/' and the year is assigned.

Perhaps we could just assign the string that is returned when parsed?

Fixes #200

### DIFF
--- a/src/backend/expungeservice/models/case.py
+++ b/src/backend/expungeservice/models/case.py
@@ -29,6 +29,6 @@ class Case:
     @staticmethod
     def _set_birth_year(info):
         if len(info) > 1:
-            return int(info[1])
+            return int(info[1].split('/')[-1])
         else:
             return ''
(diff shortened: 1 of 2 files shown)


NickSchimek:
@wittejm Merging this in. It's a hot fix for Michael. Let me know if you have any comments/questions.

wittejm:
@NickSchimek Gotcha. At a glance it looks good to me, I would give it a closer look but at the moment I'm having a bit of brain constraint for actually looking at the code :)

PR #212, opened 2019-06-20
Add database operations to users and auth endpoints

### LINKED ISSUE #172: Connect /users/ endpoint to the database
Where we are using dummy data or stubbed calls replace w/data from the database reads and writes.

In `endpoints/user.py` do:

    from flask import g
    from expungeservice.database import user

The Flask `g` object will have a handle to the database stored on it in `g.database`. Use `user.create_user()` to create a user and `user.get_user_by_email()` to read a user, passing in `g.database` as the `database` parameter.

More details TBD.

wittejm:
The users post method now creates a new user entry in the database. Also the auth_required decorator now checks the user's login credentials against the database before providing an auth token. 

In order to obtain the logged-in user's email within the users endpoint code, I attach the auth JWT payload to Flask's g object as the field `auth_hdr_payload` . I'm not sure this is the best solution but I liked it better than decoding the auth token twice, during both the auth validation step and the new-user creation step. 

I refactored the auth_test as a TestCase object to make use of setUp and TearDown for managing the database content. 

closes #172 .


a-ilango:
A little late to the party... but if we want to avoid using the global storage, we could define a new `auth_required_admin` decorator that can be used to decorate the POST method to /users/. This new decorator will have the additional step of checking if the user is an admin. Question: are we going to add admin users themselves manually to the DB?

wittejm:
@a-ilango cool, I like the idea of having a second decorator better than what I've done here.

I think we'd either need to add at least one admin user to the database manually, or we could do it in a deployment script for prod. I don't know the convention for this sort of thing, but a sql script that inserts the set of emails and hashed passwords would work. After that any admin user can create other admin users with this endpoint call.

a-ilango:
> I think we'd either need to add at least one admin user to the database manually, or we could do it in a deployment script for prod

That sounds reasonable. 👍

src/backend/expungeservice/endpoints/auth.py
```diff
@@ -31,7 +31,7 @@ def wrapper(*args, **kwargs):
             if auth_hdr == None:
                 return 'Missing Authorization header!', 401
 
-            match = re.match('Bearer\s+(.+)', auth_hdr)
+            match = re.match('Bearer +(.+)', auth_hdr)
```
maxkwallace: Why do we need to support a variable number of spaces after "Bearer" in the header? If this is something we need to do, it should be commented.

This is a simple regex so it's probably fine but just as a style issue I'd vote for replacing the regex with string methods instead.

src/backend/expungeservice/endpoints/auth.py
```diff
@@ -41,8 +41,10 @@ def wrapper(*args, **kwargs):
                 current_app.config.get('JWT_SECRET_KEY')
             )
 
-            # TODO temp hack - replace with table from DB
-            if not payload['sub'] in users:
+            g.auth_hdr_payload = payload
+
+            user_db_result = get_user_by_email(g.database, payload['sub'])
```
maxkwallace: Maybe this is out of scope for this PR, but it's best practice to use a immutable, numeric ID as the primary key for users at all times.

(The only exception to this is when someone logs in and you have to look them up with their email address.)

Given the nature of this project, maybe this is fine (maybe we don't allow users to change their emails) but at least this should be commented somewhere (if it's not already) as a form of technical debt.

src/backend/expungeservice/endpoints/users.py
```diff
@@ -2,24 +2,50 @@
...
         data = request.get_json()
 
         if data == None:
             abort(400)
 
-        # TODO temp hack - replace with table from DB
-        if data['email'] in users:
-            return 'User already exists! Please login', 202
+        """
+        - the @auth_required decorator identifies the logged in user. If no auth, return 401 UNAUTHORIZED.
```
maxkwallace: Should this be a comment for the auth_required method? Better to have it there, I think, someone who is reading this method and isn't sure what `auth_required` would ideally look at the method definition.

src/backend/expungeservice/endpoints/users.py
```diff
@@ -2,24 +2,50 @@
...
+        - Verify the logged in user is admin. if not, return 403 FORBIDDEN.
+        - Verify the password meets the 8-character length requirement. If not, return 422
+        - Attempt to insert the user into the database. If failure due to duplicate entry, return 422
+        - If success, return the new user's email, admin flag, and creation timestamp.
+        """
+
+        authorized_user_db_result = get_user_by_email(g.database, g.auth_hdr_payload['sub'])
+        if not authorized_user_db_result['admin']:
```
maxkwallace: This will result in a 500 error if the user can't be found and `get_user_by_email` returns `None`, right? Seems at least worthwhile to comment, even if this is an edge case.

src/backend/expungeservice/endpoints/users.py
```diff
@@ -2,24 +2,50 @@
...
+            return 'Logged in user not admin', 403
+
+        if len(data['password']) <8:
```
maxkwallace: This will throw a KeyError if there is no `'password'` entry in `data`, right? Just seems worthwhile to comment, since this would be a client error.

We'd ideally want a general method to use in Controllers that, given a dict and a list of required keys, checks the dict to ensure those keys are present.

src/backend/expungeservice/endpoints/auth.py
erikjasso: I'm with Max. I suggest splitting on whitespace or 'Bearer' and checking the second item in the list.

src/backend/expungeservice/endpoints/auth.py
erikjasso: @maxkwallace I think that's correct. The initial lookup upon login has to be by email or username (that's the only info we have about the user) but after the password check the user_id should go in the token and on subsequent requests authentication should be done by decoding the token and checking user_id.

erikjasso: Another point to make here is that `jwt.decode` can throw a number of exceptions and those exceptions should be checked when decoding the token. See this page for more info: https://pyjwt.readthedocs.io/en/latest/api.html

src/backend/expungeservice/endpoints/users.py
erikjasso: @maxkwallace makes a good point here. Any time a `POST` or `PUT` body comes in we should check it for the required fields.

config/postgres/initdb/scripts/create-tables-and-functions.sql
```diff
@@ -72,7 +72,7 @@ WITH USER_INSERT_RESULT AS
             INSERT INTO AUTH (auth_id, hashed_password, user_id)
             SELECT uuid_generate_v4(), $2, user_id FROM USER_INSERT_RESULT;
             
-SELECT USERS.USER_ID::varchar, EMAIL, ADMIN, AUTH_ID::varchar, HASHED_PASSWORD
+SELECT USERS.USER_ID::varchar, EMAIL, ADMIN, AUTH_ID::varchar, HASHED_PASSWORD, DATE_CREATED, DATE_MODIFIED
```
erikjasso: Does using `UUID` type work? As in`SELECT USERS.USER_ID::uuid, ...`? If not, let's open a bug for it.

In the meantime let's be consistent (you use `::text` below in the test).

src/backend/expungeservice/endpoints/auth.py
erikjasso: Ah, after looking at the full file I see you are checking for exceptions. Never mind!

config/postgres/initdb/scripts/create-tables-and-functions.sql
wittejm: I did start converting these to strings because converting the psycopg result to a dict gets uphappy about uuid type. Sounds good to look into this but for now I'll just make everything text type.

src/backend/expungeservice/endpoints/users.py
wittejm: The auth_required decorator checks that the named the user exists just before this database fetch is called. An error would throw if the user is deleted from the database between these two ops. I feel like this is safe enough to not worry about it.

src/backend/expungeservice/endpoints/users.py
wittejm: Sounds good, moving that over.

src/backend/expungeservice/endpoints/auth.py
wittejm: Yep I was sloppy with this, replacing it with split() and strip() string ops make sense.

src/backend/expungeservice/endpoints/users.py
maxkwallace: I agree we shouldn't do anything about it in the logic itself! But I'd say just take the reply you wrote ^ and paste it in to the code as a comment. Takes 10 seconds and you'll probably save people time in the future.

NickSchimek:
@wittejm I looked over your tests and they all look good to me! Thanks for doing this, and thanks @erikjasso and @maxkwallace for reviewing the code!

wittejm:
Thanks @a-ilango @maxkwallace @erikjasso @NickSchimek for the feedback!

Here's a new commit. I fixed minor issues mentioned (processing the auth_hdr string, a few comment lines), and did some bigger changes:
 - Defined separate endpoint decorators user_auth_required and admin_auth_required.
 - The auth_token object now stores the logged in user_id instead of their email.
 - to make that useful I added a database operation to lookup a user by user_id. There are now two functions to look up the user by either id or email, but if y'all think a single function with a flag argument is a better approach, my ears are open. 
 - With the new admin decorator there's currently no longer a need to identify the logged-in user in endpoint code, so I removed the payload data from Flask.g. 
 - wrote a new function to check that the required request fields are present, put it in request.py for lack of a better spot in our current code structure.
 - While adding a few test cases for the new auth decorate, I removed protected.py and its associated endpoint from main code and moved its content into the unit test file.

doc/design.md
```diff
@@ -152,22 +152,24 @@ Creates a user
...
+    * admin
 
 Returns: New user
 
 - format: `JSON`
 - fields:
     * email
-    * admin permissions
+    * admin
     * timestamp
```
maxkwallace: user_id I should be in the returned fields, I think. (Sorry, should have thought of this during the first review.)

src/backend/expungeservice/endpoints/users.py
wittejm: @maxkwallace I added a function  to request.py to use for checking required data field in the request. Let me know if you think there's a better place to put this.

src/backend/expungeservice/endpoints/users.py
wittejm: @maxkwallace I didn't put this in because my refactor makes it irrelevant: the check for admin status is made on the same db result -- in the admin_auth_required function.

src/backend/expungeservice/endpoints/users.py
maxkwallace: I don't know the conventions / inheritance structure well enough for this code to comment authoritatively on where to put it. I think having it in `request.py` is probably fine since it's easy to move later if we need to.

src/backend/expungeservice/endpoints/users.py
maxkwallace: Sounds good! Thanks for pointing that out!

src/backend/expungeservice/endpoints/auth.py
wittejm: I think my refactor addresses this: it no longer uses email as an identifier except in the auth_token GET function (I realized this function is missing the check for required data fields though. Will add that.)

doc/design.md
wittejm: Sure I can add this. I think earlier we (someone?) had decided the frontend doesn't need to see the user_id. I'm curious what any uses would be, but I can add this because doing so doesn't seem to be a problem.

maxkwallace: Oh yeah, if the frontend doesn't need to see it, then it's probably best to leave it out. But there should be a comment saying that this is an intentional decision due to the design of the app: no user needs to see the ID.

The main usecase I can think of is for admin users: if they see a list or table of other users, it would be correct to use the user_id as a primary key on the frontend as well. And if admin users can do update or delete operations on other users then they'll need the user_id for those.

src/backend/expungeservice/endpoints/users.py
```diff
@@ -2,24 +2,49 @@
...
+        - If required fields are missing in the request, return 400
+        - Password must be 8 or more characters long. Otherwise return 422
+        - Email must not already be in use by an existing user. Otherwise return 422
+        If success, return 201 with the new user's email, admin flag, and creation timestamp.
+        """
+
         data = request.get_json()
 
         if data == None:
             abort(400)
```
wittejm: @maxkwallace Aside question: I followed the pre-existing pattern that most endpoint errors are returned as a string and status code, but this one written before me is being raised instead. I also am using abort(400) in check_data_fields() so the error will propagate. Do you have thoughts on raising vs returning errors generally?

maxkwallace: Nope, I don't know enough about the conventions in Flask to comment offhand.

wittejm:
Pushed new changes to cover all the remaining comments.

Issue #139, opened 2019-05-16
Crawler should return a Record object


NickSchimek:
## Description
Currently the crawler returns a RecordParser object. This is not ideal and is confusing to say the least. There is a Record class in the models directory that can be returned instead of a record parser.

More detailed information will come later.

### This task is done when:

- [ ] Crawler returns a Record object

PR #214, opened 2019-06-22
Create a record object

gingerben93:
Added Record Object and Tests. To be used by Crawler.
Notes: Make test fails with database and endpoints test after sync to upstream. I assume those are problems not caused by changes made by me, but i always get nervous ;-). Could also just be problems on my end. I'm still trying to familiars with code base and first time set up stuff!


NickSchimek:
Thanks @gingerben93 

Regarding the failing tests on you local dev env... Those tests require the database to be setup. Unfortunately there were no install/setup instructions added to the README when those tests were added to master. Unless you are using the docker setup those tests will fail without a local database setup. It's pretty easy to get the database setup, and I can walk you through it at the next meeting.
We need to either add instructions to the README to set that up, or advise people to use docker <- which is what we need to decide at the next meeting. 

All tests are passing on the Ci so we're good there 👍

src/backend/tests/models/test_record.py
```diff
@@ -0,0 +1,34 @@
+import unittest
+
+from expungeservice.models.case import Case
+from tests.models.record_object import Record
+from tests.factories.case_factory import CaseFactory
```
NickSchimek: The CaseFactory is not be used here. This can be removed. Although we could use it below

src/backend/expungeservice/models/record_object.py
```diff
@@ -0,0 +1,21 @@
+from expungeservice.models.case import Case
```
NickSchimek: This can be removed

src/backend/expungeservice/models/record_object.py
```diff
@@ -0,0 +1,21 @@
...
+    def charges(self):
+        list_charges = []
+
+        for case in self.cases:
+            list_charges.extend(case.charges)
+
+        return list_charges
+
+    def total_balance_due(self):
+        total = 0.0
```
NickSchimek: total can be an integer `total=0`

NickSchimek:
@gingerben93 The Record class looks good! I have some suggestions on how to test it. Can we go over it next meeting or would you prefer I add comments?

wittejm:
> Notes: Make test fails with database and endpoints test after sync to upstream.

@gingerben93  I have an open PR #190 that has instructions to install the docker stack, which will get an instance of the database running on your system. That should make those tests pass. Also if you give those instructions a shot and let me know how it goes I'd really appreciate it! (hopefully you're not on windows because I didn't look into how to install Docker there :( )

gingerben93:
@NickSchimek We can go over ways to test next meeting if that's most convenient for you, but feel free to add comments if you want.

NickSchimek:
OK I will add a few comments, and we can also discuss some strategies at the next meeting too if needed.

src/backend/expungeservice/models/record_object.py
```diff
@@ -0,0 +1,21 @@
+from expungeservice.models.case import Case
+
+class Record:
```
NickSchimek: There is nothing wrong with the Record class. You did an excellent job! I can see it will work as intended. One thing that we can do, that I just learned about, is to use the `@property` decorator on the `charges` method. This will make the method act like an attribute so that we can ask `Record.charges` instead of `Record.charges()` -> Obviously not that big of an issue, but if you'd like to see how it can be used here is a good site: https://www.programiz.com/python-programming/property

gingerben93:
@wittejm I followed your instructions and now have all the test working; however, I've used Docker before and already had it installed. The only small hiccup I encountered what with the the first command "docker swarm init". I needed to add "docker swarm init --advertise-addr 10.0.0.37". I'm on a VM using Ubuntu and I have the connection bridged. Anyone with multiple MAC address available will have to specify the IP address to use.

src/backend/tests/models/test_record.py
```diff
@@ -0,0 +1,34 @@
...
+        self.case1.set_balance_due("123")
+        self.case2.set_balance_due("246")
+
+        self.list_cases.append(self.case1)
+        self.list_cases.append(self.case2)
+
+        self.record = Record(self.list_cases)
+
+    def test_print_balance_in_cents(self):
+        assert self.record.total_balance_due() == 369.0
```
NickSchimek: Ideally someone should be able to look at a test and know what it's testing. When I look at this test I have no idea where 369 comes from. Obviously if I look at the setup method then I can see it's origin, however the setUp method is about 12 lines long and has the potential to become longer as more tests are added following this pattern. Ideally the setUp method should only contain code that is common to all tests, hence it DRY's up the tests. 

Here is one example:
```
def test_total_balance(self):
    record = RecordFactory.create(CaseFactory.create(balance_due='123.00'),
                                                         CaseFactory.create(balance_due='246.00'))

    assert record.total_balance == 369.0
```

So looking at this test I can now easily parse that I have a record with two cases, and can see the balance due on each of them. And then I always like to test for emptiness. for example:

```
def test_total_balance(self):
    record = RecordFactory.create()

    assert record.total_balance == 0.0
```
Here the record has no cases (depending on the implementation of the create method) and so it should have a balance of 0

NickSchimek: I wonder if we could tell python that when returning the total_balance to do it with a precision of two so that we would get back `369.00`

gingerben93: Thanks for the suggestions Nick. I will go back and make a record factory class; I agree that removes all the clutter from the setUp method. I will get to that and other recommendations you had as soon as I have the time.

src/backend/expungeservice/models/record_object.py
NickSchimek: One thing I would like to do is change all the current code in master to use the `@property` tag. But it's not a priority at this time. However I would like to see new code use... That is why I bring it up.

wittejm:
Merging this PR to unblock issue #139 . Creating a new issue to complete the changes requested here by @NickSchimek .

Issue #174, opened 2019-06-06
Update build process


erikjasso:
We want to have different environments for `dev` and `prod` application containers w/different contents. For example, we want `dev` containers to have tests but not the `prod` containers. To achieve this we will change the application build process.

### This task is done when:
- there is a `build` target in `Makefile` that:
    - makes a directory called `build` at the same level as `src`
    - copies contents of `src/backend/expungeservice` into `build` (we may want to be explicit about what is copied, such as excluding the `README`)
- there are `build_prod` and `build_dev` targets that have `build` as a pre-requisite
- `build_prod` does nothing else
- `build_dev` also copies the `tests` dir
- `make dev` has `build_dev` as a pre-requisite
- `make dev` successfully brings up the service container and the container has the tests in it

wittejm:
@erikjasso Did you mention also that each make build target should wipe out the build directory after everything else is finished?

wittejm:
A change to this issue: with deploying our app to Heroku, there is no need to use Makefile targets for the prod deploy workflow. I think all of it can go into its own shell script. The remaining todo item in this issue is to have the Docker expungeservice_image include the test code (and run it at container launch time?).

PR #218, opened 2019-06-26
Nginx webserver Docker image

wittejm:
* Docker image that contains the static web content and the configured nginx server process, which handles uri redirection for api calls.
* New stanza in docker-compose.yml defining the webserver service.
* port 3000 of the webserver container is published to 3000 in docker-compose.dev.yml, matching the behavior of running the frontend with node. I assume in prod the server would instead publish to 80 or 443, which are already exposed by the nginx base image.
* New make target `webserver_image` that builds the docker image.
* In the make file, webserver_image is added to `make dev` which builds the docker stack.
* The Nginx base docker image already redirects the server access and error logs to std_out and std_err, which get captured by docker logging.
* The expungeservice Docker image now runs a wsgi server instead of the Flask server. The Nginx reverse-proxy is configured to forward api calls as wsgi instead of http.
* The nginx server will need to be configured for https when we have a certificate. For that we need a domain name. For now I think it's fine to merge this branch to Master without it since we're still a ways from deployment to prod.
* This configuration is specific to the dev environment. This will be refactored as part of #174 but that is out of scope for this PR.
* All the expungeservice endpoints have a leading "/api/" in their uri. I added this to the /hello/ endpoint also for consistency. I think keeping this endpoint for now makes sense because it's the only one that doesn't require any data in the request.
* The nginx Dockerfile is based on the tutorial here: https://medium.com/thepeaklab/how-to-deploy-a-react-application-to-production-with-docker-multi-stage-builds-4da347f2d681
* Building the new docker image in Travis slows down the build again :(((( . It might be possible to separate the build into two stages like we did to speed up the expungeservice build but that's another issue.

Addresses #43 and #45. With the new wsgi.py file I think this finishes off #44 also.


wittejm:
I'm debugging an issue related to environment variables which is showing itself in the expungeservice container. Errors will be easier to locate once we have test cases running in Travis with #174

wittejm:
Fixed the build. Navigating to localhost:/api/hello in-browser was getting a 502 error because the backend was failing due to a missing env variable, but this error wasn't initially detected in the Travis build. I can merge this branch soon but if anyone wants to review I request they pull and deploy the docker stack to verify it's not just #WorkingOnMyMachine .

wittejm:
I decided to take the webserver_image build out of the travis script to avoid the extra build time. I don't think that having it in there adds much value to the CI check, since frontend testing isn't exactly part of what Travis does.

PR #224, opened 2019-06-29
Refactor type analyzer

### LINKED ISSUE #207: Refactor type analyzer
## Description
The type analyzer currently goes through a series of if/elif statements to determine the type of charge that it has. ~~I suggest refactoring this to use the state design pattern so that the type analyzer asks the charge what type it is and calls the appropriate method/object to determine type eligibility.~~
It was not possible to refactor it in that way. Well... to clarify it is possible, but that was not the best approach.  A factory method and polymorphism should get the job done.

Doing this will also have the side effect of reducing the public interface on charge which is quite overwhelming right now.
### This task is done when:
 - Update the charge class to have an attribute and/or method that returns what type of charge it is. https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/charge.py#L11
 - Update the type analyzer to invoke the correct state based on the return value of the newly implemented charge method.
https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/charge.py#L11

NickSchimek:
Well the type analyzer was actually deleted. It is no longer needed.

Each charge is of a specific type, however the current implementation did not know which type the charge belonged to unless it was ran through the type analyzer, and there was no way to know the type of charge at instantiation. To remedy this a factory method was created to instantiate a charge.

Inheritance is setup for each type of charge to inherit from the BaseCharge. This has many advantages over the previous implementation, and should allow for greater flexibility as the expunger evolves.

FYI: Tests were not added as the behavior has not changed, only the design. Some tests were updated to reflect the new design pattern.

Fixes #207


NickSchimek:
Thanks @wittejm Yes I agree that renaming those two files would be good. I was giving it some thought at the time I did this refactor as well. Now that's it's complete I should give it some more attention. I do kind of like having the class containing the factory method named charge, because then I can say `charge.create`, but it's probably better to have it named something else.

As far as the initial project spec is concerned the logic is complete, however the initial project spec did not fully capture the complexities of the law, and the ambiguous and missing data from the OJCIN site. Additionally the law is a living document and changes to it are inevitable.  There are a couple other reasons for this refactor that I can go into detail about in a later post.

PR #238, opened 2019-07-06
Create doc for proposed json format

NickSchimek:
This is an updated version than the one that I posted on slack a while ago. Now that we have a record object it made sense to update. Also would like to get this into the docs so that it is easy to find and update.

### DIFF
--- a/doc/results_format.json
+++ b/doc/results_format.json
@@ -1,0 +1,100 @@
+{
+    "data":
+    {
+        "record":
+        {
+            "total_balance_due": 199.99,
+            "cases":
+            [
+                {
+                    "name": "Doe, John",
+                    "birth_year": 1970,
+                    "case_number": "HA05555555",
+                    "citation_number": "HA08643443",
+                    "location": "Multnomah",
+                    "date": "09/05/2008",
+                    "violation_type": "Municipal Parking",
+                    "current_status": "Closed",
+                    "charges":
+                    [
+                        {
+                            "name": "Loading Zone",
+                            "statute": "29",
+                            "level": "Violation Unclassified",
+                            "date": "09/04/2008",
+                            "disposition":
+                            {
+                                "date": "11/18/2008",
+                                "ruling": "Convicted"
+                            },
+                            "expungement_result":
+                            {
+                                "type_eligibility": false,
+                                "type_eligibility_reason": "Ineligible under 137.225(5)",
+                                "time_eligibility": null,
+                                "time_eligibility_reason": "",
+                                "date_of_eligibility": null
+                            },
+                        }
+                    ],
+                    "balance_due": 199.99,
+                    "case_detail_link": "https://publicaccess.courts.oregon.gov/PublicAccessLogin/CaseDetail.aspx?CaseID=11918348"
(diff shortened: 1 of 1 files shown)


doc/results_format.json
```diff
@@ -0,0 +1,100 @@
+{
+    "data":
+    {
+        "record":
+        {
+            "total_balance_due": 199.99,
+            "cases":
+            [
+                {
+                    "name": "Doe, John",
```
SteelFace87: Should we move the name and birth year to the record object since they repeat in each of the objects in the cases array?

NickSchimek: It's possible an alias is used, hence we can't assume they repeat for each case. I haven't taken a look at the front end requirements in a while so as your working on this let me know if you find anything that is missing.

NickSchimek:
@hmarcks Could you take a look at this when you get a chance, and let us know if anything is missing, or anything extra that you'd like to see added?

doc/results_format.json
hmarcks: Here's a mockup of what grouping by cases would look like:

group-by-cases

NickSchimek: hmmm. not sure. It would be best to get Michaels input... I know one thing for sure he wants is to have all eligible cases grouped together. Of course there could probably be a situation where a case has both eligible and ineligible charges. 

Perhaps what he means is to groups all cases that have eligible charges at the top so that the person can quickly navigate to cases that have eligible charges...

Even if we show Michael this he could change his mind after seeing it and using it...

hmarcks: Yeah, I'll ask Michael about not requiring the fields.

A somewhat related question - if a search/analysis is preformed that returns multiple people (say only first name and birth date is submitted), is it correct that the time eligibility would look at the records/cases as a whole?

If so, we wouldn't get an accurate time eligibility since it's considering more than one person.

NickSchimek: Yes, if the results returned contain multiple people it will treat them all as belonging to the same person. However, Michael said that they would only put in search parameter's that will return a single person. Also the system requires first and last name, but you can use `*` filter to get more results such as first-name: `nic*` and last-name: `sch*`... Although we don't have to worry about that, because they wont be using it like that.

hmarcks: Got it, thanks Nick!

erikjasso:
I think this looks good. I just have a few suggestions.

- I would remove the outer "data"; it seems unnecessary unless there will be other objects under it
- I would split "name" into "first_name" , "middle_name", and "last_name" to avoid having to parse it and to make sorting easier. Have you seen any additional parts like "jr", "sr", etc? If so I would add a fourth "name_suffix" or something similar
- Dates should be in ISO 8601 format such as `YYYY-MM-DD`. If you're using Python's `datetime` module it has a `toisoformat()` function to convert to ISO 8601. (I know our audience is not international but using this standard format is good practice)

doc/results_format.json
hmarcks: Michael said "Yes that would definitely be more helpful" to group by cases. I'll create some markup to support this and update the designs.

PR #242, opened 2019-07-09
Inject Record object into expunger

### LINKED ISSUE #241: Update expunger to use record object
Since we have a record object now that is returned from the crawler it should be injected into the expunger instead of a list of cases.

### This task is done when:
 -
 -

NickSchimek:
small refactoring. However this does change the api for the expunger. It now takes a Record object instead of a list of cases. @ash-tamraz you OK with this change?

I'm going to continue refactoring on this branch over the next couple days, but wanted to get this in since it affects @ash-tamraz work on the search endpoint. 

Fixes #241

### DIFF
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -13,15 +13,15 @@ class Expunger:
     query's if/when we start persisting the model objects to the db.
     """
 
-    def __init__(self, cases):
+    def __init__(self, record):
         '''
         Constructor
 
-        :param cases: A list of cases
+        :param record: A Record object
         '''
-        self.cases = cases
+        self.record = record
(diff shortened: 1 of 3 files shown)


ash-tamraz:
@NickSchimek This all looks good to me! I don't think it should present any problems for the search endpoint.

ash-tamraz:
The Travis CI build check passes. I made a mistake which restarted the check, and merged before the build finished, which is why GitHub is flagging a failed build.

NickSchimek:
Interesting... It looks like it couldn't pull the PR after being merged...
```
$ git fetch origin +refs/pull/242/merge:
fatal: couldn't find remote ref refs/pull/242/merge
```
well the build is passing on master! so we're all good!

Thanks @ash-tamraz

PR #255, opened 2019-07-11
Tag and skip charges without dispositions

### LINKED ISSUE #210: Tag and skip any closed charge that does not have a disposition
## Description
Cases that are closed may not have a disposition header assigned due to the way the judgement was entered by the administrator. That creates an issue that a charge does not have a disposition when the case is closed. 

When that happens there is an error due to the disposition does not exist when there should be one...
```
  File "cli.py", line 52, in <module>
    expunged = expunger.run()
  File "/Users/miafro/Documents/codeforpdx/recordexpungPDX/src/backend/expungeservice/expunger/expunger.py", line 49, in run
    self._categorize_charges()
  File "/Users/miafro/Documents/codeforpdx/recordexpungPDX/src/backend/expungeservice/expunger/expunger.py", line 81, in _categorize_charges
    if charge.acquitted():
  File "/Users/miafro/Documents/codeforpdx/recordexpungPDX/src/backend/expungeservice/models/charge.py", line 28, in acquitted
    return self.disposition.ruling[0:9] != 'Convicted'
TypeError: 'NoneType' object is not subscriptable
```

### This task is done when:
 -
 -

NickSchimek:
Sometimes a charge in a closed case will not have a disposition, tag and
skipping per clients request.

Fixes #210

### DIFF
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -22,6 +22,7 @@ class Expunger:
         self.record = record
         self._charges = record.charges
         self.errors = []
+        self._dispositionless_charges = []
         self._most_recent_dismissal = None
         self._most_recent_conviction = None
         self._second_most_recent_conviction = None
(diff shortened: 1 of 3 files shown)


NickSchimek:
@ash-tamraz This time I pushed up the test first that way you can see it's actually testing something and failing! Fortunately, on this repo, we don't have to wait 15 hours for it to run!

PR #274, opened 2019-07-18
Update Docker make targets and docs

wittejm:
- removed the obsolete `make run` because the dev backend should run in Docker, not  with the Flask server.
- added a new make target `dev_start` and  few recommendations for using Docker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,13 +3,10 @@
 install:
 	pipenv install
 
-run: install
-	pipenv run flask run
-
 clean:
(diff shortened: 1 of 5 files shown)


wittejm:
Folks just starting to use Docker @ForrestLonganecker @maxkwallace @NickSchimek @cjbreaux , let me know if you have related questions beyond these doc additions. Forrest said that `make dev_stop` was not actually stopping the docker stack which seems odd. If anyone else is having that problem let me know because I don't really know what to make of it.

Issue #87, opened 2019-03-28
Add SearchResults Component in React


maxkwallace:
See the lower card under "Results for Firstname Lastname" on http://dev.huntermarcks.net/search/

This task is to build a Component that contains the HTML you see in the box with "Results for {firstname lastname}" at the top. This box has a header showing the name of the person searched for, with a variable-length list of results. The results sections should be implemented using one or more separate Components.

These Components do not need to interface with the backend data source. You are welcome to hard-code data into your Components that matches the mock data visible on http://dev.huntermarcks.net/search/

The UI should match the mockups as well.

### This task is done when...
 - SearchResults and other Components (as needed) to create the search results view exist in React

![Record Expunge Search UI Data v2](https://user-images.githubusercontent.com/6887982/55674238-b1898900-5866-11e9-8f8c-aa2ae42312bf.png)

hmarcks:
Closing because it's a duplicate of https://github.com/codeforpdx/recordexpungPDX/issues/86

maxkwallace:
@hmarcks I didn't intend for this to be a duplicate. My bad if the issue was confusing.

My intention was for #86 to be only the top card containing the input fields for the search, and this issue to be only the lower card containing the search. These should be separate components in React. Let me know if that doesn't make sense or you think the issues should be structured differently.

hmarcks:
@maxkwallace gotcha, I was thinking #64 would be the integration work, perhaps these should be combined.

maxkwallace:
@hmarcks ah, I lost track of #64 , my bad. I was trying to break things down a bit more. For a project like this it's really crucial that each task is super small so no one get stuck waiting for a big chunk of work they have a dependency on. 

The way this works with Redux is that the search and results Components actually never communicate directly with each other, which is good, because then they can be treated independently.

The search component just puts the results into the redux store, at which point the results component displays them. That work can be started without the backend but we'd be guessing a bit at the format of the data. 

My personal opinion is that #64 be the integration task to finish the feature after the backend is returning results in response to HTTP requests from the frontend, but I'm open to other options as well.

hmarcks:
Sounds good to me, thanks @maxkwallace !

SteelFace87:
I am starting to work on this

maxkwallace:
@SteelFace87 awesome! Thanks for letting us know.

NickSchimek:
@SteelFace87 is this complete? Or is there still work to be done?

[...]

PR #281, opened 2019-07-23
Create api service, test, and a mock file

SteelFace87:
With this PR I am trying to accomplish several tasks

1)I want to have a universal API service using AXIOS that will take any url, method and payload, and is capable or returning errors and data.  This way we will all be creating API calls the same way.

2)Create a mock function that allows testing to pass while back end routes are in development, and to give JSON data so that I can start using Redux with the SearchResults Component (issue #87)


src/frontend/src/service/api-service.ts
```diff
@@ -0,0 +1,41 @@
...
+  | 'GET'
+  | 'get'
+  | 'DELETE'
+  | 'delete'
+  | 'head'
+  | 'HEAD'
+  | 'put'
+  | 'PUT'
+  | 'patch'
+  | 'PATCH';
```
maxkwallace: I think we should standardize on either capitalized or uncapitalized method names and not allow both. I personally prefer lowercase names but it's not a strong preference; I'm open to either.

src/frontend/src/service/api-service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+type Req = {
```
maxkwallace: I think we should rename `Req` to a full word, so it's clear whether this is "requirements" or "request". https://github.com/airbnb/javascript#naming--descriptive

src/frontend/src/service/api-service.ts
```diff
@@ -0,0 +1,41 @@
...
+  data?: object;
+  method: MethodOptions;
+};
+
+export default function apiService(req: Req) {
+  return axios
+    .request(req)
+    .then(res => [res.statusText, res.data, res.status])
+    .then(([statusText, data, status]) => {
+      if (statusText) {
```
maxkwallace: Given that this is an `if`, when will `statusText` not be truthy? That should be commented.

If we don't expect `statusText` to ever be falsy then IMO we should throw an exception and `alert` if it is.

src/frontend/src/service/api-service.ts
```diff
@@ -0,0 +1,41 @@
...
+          status,
+          statusText
+        };
+      }
+    })
+    .catch(e => {
+      return {
+        error: e.message
+      };
+    });
```
maxkwallace: Correct me if I'm wrong-- 

but as written, all calls to the `apiService` function will return a promise that resolves to an Object, regardless of whether they succeed or fail. This means that people using `apiService` will need to check whether the returned Object has an `'error'` key or not to determine whether or not the request succeeded. 

Why shouldn't we just return the axios promise to people using `apiService`? The `.then` and `.catch` Promise APIs were created for this use case, so I think we should use them in application code. It's similar to a `try` / `catch` block where you want to separate the business logic for success and failure so people don't have to use an if statement to figure out what to do. (The other issue here is that people may want to access properties of the error other than the message, e.g. the error code.)

I'm open to iterating on this API. It doesn't have to be 100% correct the first time. Just my $0.02.

src/frontend/src/service/__mocks__/axios.js
```diff
@@ -0,0 +1,66 @@
...
+      });
+    } else if (
+      request.url.includes(
+        'http://localhost:5000/api/' & (request.method === 'GET')
+      )
+    ) {
+      return Promise.reject(new Error('Request failed with status code 404'));
+    } else if (!request.url.includes('http://localhost:5000/api/')) {
+      return Promise.reject(new Error('bad url'));
+    }
```
maxkwallace: Should we have a final `else` that throws an exception to guard against typos in the URL when people are writing tests? This is a small addition that could save someone a lot of frustrating debugging.

src/frontend/src/service/api-service.ts
SteelFace87: I have not updated the MethodOptions to look as seen below.  However when I try to change my api-service.test.js tile to  a typescript file, the type check fails and it says "type string is not assignable to method options".  Would someone know how to fix this TypescriptError?
enum MethodOptions {
  post = "post",
  get = "get",
  delete = "delete",
  head = "head",
  put = "put",
  patch = "patch"
}

src/frontend/src/service/api-service.ts
SteelFace87: I have updated req to request.  Thank you for the article, I will keep this in mind.

src/frontend/src/service/api-service.ts
SteelFace87: I was under the assumption that if something is to go wrong with the fetch then it wouldn't return statusText, data, and or a status.  I am simply checking if we got a response by checking if statusText is available.  In other words statusText will be truthy as long as the fetch worked properly.  The if statement might as well have checked if data or status is available.  In either case if we don't get a response from the service then the error is caught and returned instead.    Let me know if this makes sense, and or how you would want to modify the code.

src/frontend/src/service/api-service.ts
SteelFace87: Are you suggesting that we cut out all of the logic and leave it up to the user on how to handle the response and or error.  The only purpose of the api-service.ts file then would be to make sure that the correct format is used when sending a request through Axios by using TypeScript.  In this case the code would look as follows.
export default function apiService(request: Request) {
  return axios
    .request(request)
    .then(res => res)
}

src/frontend/src/service/__mocks__/axios.js
SteelFace87: I updated the error message to provide the correct base url incase there is a typo.

src/frontend/src/service/api-service.ts
maxkwallace: Not necessarily. If you want to keep the `.then` logic that's fine with me, as long as we're open to possible changes if someone needs to access a property of the result object other than `statusText`, `data` or `status`.

My main point is that I think in application code, people should be able to write (A):

    apiService(...).then(result => {

    }).catch(result => {

    })

if they prefer, rather than requiring everyone to write (B):

    apiService(...).then(result => {
      if (result.hasOwnProperty('error')) {

      } else {
      
      }
    })

If this second way were better in all cases, then why would the creators of Promises ever have added the `catch` API? There are things you can do with A that you can't do with B, and sometimes it's good to be able to do those things in application code. If you prefer B that's fine. You can add a second wrapper for apiService that's optional to use. I'm happy to talk about this more at the meeting next week.

src/frontend/src/service/api-service.ts
SteelFace87: I have now fixed the type error.

src/frontend/src/service/api-service.ts
SteelFace87: With TypeScript I had to refactor the code to look like this, else type checks were failing.

 .then((response: Response) => {
      return {
        data: response.data,
        status: response.status,
        statusText: response.statusText
      };
    })

src/frontend/src/service/api-service.ts
SteelFace87: I removed the then logic to leave it up to the user on how they want to handle the response and or error.

src/frontend/src/service/__mocks__/axios.js
SteelFace87: I made a default error message incase none of the "else if's" wont get triggered.

src/frontend/src/service/api-service.ts
maxkwallace: @SteelFace87 sounds good, let me know when you're done with all new updates to the PR and I'll re-review the whole thing.

src/frontend/src/service/__mocks__/axios.js
```diff
@@ -0,0 +1,66 @@
...
+            case_detail_link:
+              'https://publicaccess.courts.oregon.gov/PublicAccessLogin/CaseDetail.aspx?CaseID=9036658'
+          }
+        ],
+        status: 200,
+        statusText: 'OK'
+      });
+    } else if (
+      request.url.includes(
+        'http://localhost:5000/api/' & (request.method === 'GET')
```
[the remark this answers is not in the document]
SteelFace87: You are right,  I just fixed this :)

src/frontend/src/service/api-service.ts
```diff
@@ -0,0 +1,16 @@
...
+  url: string;
+  data?: object;
+  method: Method;
+};
+
+export default function apiService(request: Request) {
+  return axios
+    .request(request)
+    .then(response => response)
+    .catch(error => error);
```
maxkwallace: I'm not sure about two things here:

a. Why do we need this line?

        .then(response => response)

I don't think it does anything. You could delete it and the code would still function the same.

b. Why do we want this line?

    .catch(error => error)

If we catch errors here, anyone using `apiService` is still in the situation I described in https://github.com/codeforpdx/recordexpungPDX/pull/281#discussion_r307092169 except worse, because there's no obvious way to differentiate an error from a response.

SteelFace87: Ah, I see what you mean now.  I have removed the lines described in a and b, and updated the testing to work with the new code.  

I used npm i axios,  I'm not sure why my code was working without having axios in package.json.

maxkwallace: @SteelFace87 if the package is installed into your local `node_modules/` any code using that package will still work even if the package isn't listed in the `package.json`.

Issue #82, opened 2019-03-27
OECI login component


hmarcks:
Create view for user to log in to OECI (Oregon eCourt Case Information) so that they can search records.

HTML template to transfer: https://github.com/hmarcks/expunge-pdx-frontend/blob/master/index.html
Live template: http://dev.huntermarcks.net
Design system: http://dev.huntermarcks.net/design-system

User flow:
![OECI login flow](https://user-images.githubusercontent.com/6887982/55119747-53afb100-50b0-11e9-865b-e4b947f250ee.png)

maxkwallace:
Addressed in https://github.com/codeforpdx/recordexpungPDX/pull/285 (thanks @cjbreaux !)

The loading state for the login button is listed as an outstanding item in https://github.com/codeforpdx/recordexpungPDX/issues/75 so I think it's safe to close this issue.

Issue #75, opened 2019-03-14
Main login integration


hmarcks:
Integrate the UI to log in to the app, reset password, and set password.

User flow: 
![login flow](https://user-images.githubusercontent.com/6887982/54334025-e137cf00-45e1-11e9-9121-5aaa8fabc2a2.png)

Template w/ error messages: http://dev.huntermarcks.net/log-in
Design system: http://dev.huntermarcks.net/design-system

htharker42:
Thanks Hunter!

maxkwallace:
This issue is mostly done as of https://github.com/codeforpdx/recordexpungPDX/pull/285
Thanks @cjbreaux !

Here's what I think is left:
- Loading state for the 'send password reset email' button
- Change routes to the login page after resetting password
- Loading state for the login buttons

There will probably be at least a few other small issues we find since the complete feature is far from done. I recommend keeping this issue open for now. The loading states and validation/error message won't be able to be fully tested to final requirements (full "integration testing" for the feature) until we can link this functionality up with the AJAX requests to the backend.

@cjbreaux going to unassign you for now, but definitely reassign yourself if you're still working on this!

hmarcks:
Closing because the account feature has been removed

PR #285, opened 2019-07-23
Main login integration (Issues #75 & #82)

cjbreaux:
To tackle the main login pages, I implemented 'React Router'. There are now views and routes for the OECI login, forgotten password, and reset password. Some basic validations were added but more will be required to check for valid credentials. 

The "/passwordreset" route can only be reached manually atm as I believe it will be linked from an external email.


src/frontend/src/components/ForgotPassword/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    });
+  };
+
+  handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    this.validateForm();
+  };
+
+  emailCheck = (email: string) => {
+    return /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(email);
```
maxkwallace: Using a regex to validate an email address is notoriously fraught with difficulty, see https://stackoverflow.com/q/201323/1067145

I am not going to test your regex here. Just wanted to call this out and make sure the regex is relatively permissive and errs on the side of only rejecting emails that are definitely invalid.

src/frontend/src/components/ForgotPassword/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  validateForm = () => {
+    this.setState({
+      emailHasInput: this.state.email.trim().length === 0
```
maxkwallace: Should this be `!==`?

src/frontend/src/components/ForgotPassword/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    });
+    this.setState({
+      missingInputs: this.state.email.trim().length === 0
+    });
+    this.setState({
+      emailIsValid: this.emailCheck(this.state.email)
+    });
```
maxkwallace: I think you can combine these setState calls into one.

src/frontend/src/components/LoggedInHeader/index.tsx
```diff
@@ -21,7 +22,10 @@ class LoggedInHeader extends React.Component {
               Account
             </button>
             <LogOut>
-              <button className="link hover-blue f6 f5-ns dib pa3">
+              <button
+                onClick={() => history.push('/')}
+                className="link hover-blue f6 f5-ns dib pa3"
+              >
```
maxkwallace: If we use React Router's Link Component, won't it update the history automatically? That seems like a better solution. Let me know if there's something I'm missing as to why you didn't go that route. I haven't used React Router before, I'm just going off the docs at https://reacttraining.com/react-router/web/guides/quick-start/example-basic-routing

src/frontend/src/components/OECI/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+    this.validateForm();
+  };
+
+  validateForm = () => {
+    this.setState({
+      userIdHasInput: this.state.userId.trim().length === 0
+    });
+    this.setState({
+      passwordHasInput: this.state.password.trim().length === 0
+    });
```
maxkwallace: Same comments as earlier:

- I think these should be `!==`?
- I think these setState calls can be combined into one.

src/frontend/src/components/OECI/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+        missingInputs:
+          this.state.userId.trim().length === 0 ||
+          this.state.password.trim().length === 0
+      },
+      () =>
+        this.state.missingInputs === false
+          ? history.push('/recordsearch')
+          : null // Using setState(updater, callback) because setState doesn't immediately update component
+    );
+    //need validation for userId & PW
```
maxkwallace: Should be

    // Need validation for userId & password.

src/frontend/src/components/OECI/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+      {
+        missingInputs:
+          this.state.userId.trim().length === 0 ||
+          this.state.password.trim().length === 0
+      },
+      () =>
+        this.state.missingInputs === false
+          ? history.push('/recordsearch')
+          : null // Using setState(updater, callback) because setState doesn't immediately update component
+    );
```
maxkwallace: I realize the backend logic isn't in place so it's not possible to totally anticipate how this code should look, but we should not mutate global state (the history) as a side effect of validation. The history should only be updated if the login attempt was successful and the user is actually changing pages in response to that. So the logic in this callback should be moved to `handleSubmit` or removed for now.

src/frontend/src/components/PasswordReset/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    this.setState(
+      {
+        passwordIsValid: this.state.password.trim().length >= 10 // Need to implement a more robust password check.
+      },
+      () =>
+        this.state.passwordIsValid === true &&
+        this.state.missingPasswordInput === false
+          ? history.push('/')
+          : null // Using setState(updater, callback) beccause setState doesn't immediately update component
+    );
```
maxkwallace: Same comment as above, this side effect (state change) should be in handleSubmit, not as a side effect of validation. Validation should validate only and do nothing else.

src/frontend/src/components/App/index.tsx
```diff
@@ -2,23 +2,33 @@ import React from 'react';
...
 import { get } from 'lodash';
 import { AppState } from '../../redux/store';
+import { Router, Route } from 'react-router-dom';
 
+import history from '../History';
 import Footer from '../Footer';
 import LogIn from '../LogIn';
 import LoggedIn from '../LoggedIn';
 import RecordSearch from '../RecordSearch';
+import OECIlogin from '../OECI';
```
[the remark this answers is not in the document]
cjbreaux: Thanks for pointing this out Max! I can see why following a standard convention would be more important than having the names look "nice". I'll change those over to PascalCase.

src/frontend/src/components/App/index.tsx
```diff
@@ -2,23 +2,33 @@ import React from 'react';
...
 class App extends React.Component {
   public render() {
     return get(this, 'props.system.loggedIn') ? (
-      <LoggedIn>
-        <RecordSearch />
-      </LoggedIn>
+      <Router history={history}>
+        <LoggedIn>
+          <Route path="/oeci" component={OECIlogin} />
+          <Route path="/recordsearch" component={RecordSearch} />
```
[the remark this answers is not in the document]
cjbreaux: That makes sense to me. I'll update the route paths.

src/frontend/src/components/ForgotPassword/index.tsx
cjbreaux: Hmm, this is only checking if the syntax is correct not if the address is actually "valid". It's fairly permissive but won't accept addresses with the format:
`_@_@._`

src/frontend/src/components/LoggedInHeader/index.tsx
cjbreaux: I had intended to wrap the button in the Link component but was cautioned against it here:

https://stackoverflow.com/questions/42463263/wrapping-a-react-router-link-in-an-html-button

as nesting it that way is not valid HTML.

src/frontend/src/components/OECI/index.tsx
cjbreaux: Right!

src/frontend/src/components/ForgotPassword/index.tsx
cjbreaux: Ah, this kept tripping me up. The logic is correct but I've renamed the property "ariaInvalid" as I'm using it to set that property on the input field.

src/frontend/src/components/ForgotPassword/index.tsx
cjbreaux: Indeed you can, thank you!

src/frontend/src/components/OECI/index.tsx
cjbreaux: I've moved the logic to `handleSubmit` and changed it utilize `<Redirect to='#'>`

src/frontend/src/components/PasswordReset/index.tsx
cjbreaux: Using `<Redirect>` as above. Moved out of validation.

maxkwallace:
@cjbreaux thanks for the updates! I'll take a look again tonight.

cjbreaux:
@maxkwallace Thank you, your feedback is greatly appreciated!

src/frontend/src/components/ForgotPassword/index.tsx
maxkwallace: Okay! Can you take the gist of what you wrote in your comment ^ and make that a comment in the code? It's hard for someone reading the code to make sense of a regex presented without any additional context.

src/frontend/src/components/LoggedInHeader/index.tsx
maxkwallace: Hmm, okay. In that case, could you leave a comment linking to that stackoverflow post and explaining your rationale as you just did?

maxkwallace:
Also-- stop me if you're already familiar with this issue and don't need this comment-- it looks like this branch will need to be updated with the recent changes in `master` before it can be merged.

My personal preferred approach to doing this is to squash all the commits into one commit in the branch, then rebase that commit (i.e. the branch) on master, and then merge it. The possible downside is that you lose your commit history (if that's something you care about). 

If you have any questions about the process or run into any issues, feel free to let me know here or on Slack.

PR #304, opened 2019-08-02
Create a record object

gingerben93:
Ok nick i made charges a property just for you!


src/backend/expungeservice/models/record_object.py
```diff
@@ -0,0 +1,26 @@
+from expungeservice.models.case import Case
+
+from decimal import Decimal
```
NickSchimek: Can you remove both of these imports? They're currently not being used.

src/backend/tests/factories/record_factory.py
```diff
@@ -0,0 +1,10 @@
+from expungeservice.models.record_object import Record
+from expungeservice.models.charge import Charge
+from tests.factories.case_factory import CaseFactory
```
NickSchimek: Can you remove the charge and casefactory import?

src/backend/tests/models/test_record.py
```diff
@@ -1,34 +1,51 @@
 import unittest
-
-from expungeservice.models.case import Case
-from expungeservice.models.record import Record
+from expungeservice.models.record_object import Record
```
NickSchimek: Can you remove this Record import? I don't see it being used.

src/backend/tests/models/test_record.py
```diff
@@ -1,34 +1,51 @@
...
+    def test_num_cases(self):
+        
+        assert len(self.record.charges) == 3
+        
+    def test_charges_index_0(self):
+
+        assert self.record.charges[0] == self.charge_zero
+
+    def test_charges_index_1(self):
+        record = RecordFactory.create([self.case_zero, self.case_one])
```
NickSchimek: This temp var is not being used. Can this line be removed also?

src/backend/tests/models/test_record.py
```diff
@@ -1,34 +1,51 @@
...
+
+        assert self.record.charges[1] == self.charge_one
+
+    def test_charges_index_2(self):
+        record = RecordFactory.create([self.case_zero, self.case_one])
```
NickSchimek: This var is also not being used. Can it be removed too?

NickSchimek:
Thanks @gingerben93 the tests look great!
This can be merged although I'd suggest removing any unused code.

gingerben93:
Ok should be ready to go!

PR #328, opened 2019-08-19
Heroku deployment

wittejm:
Changes / Features
 - Overall: added the config scripts and instructions needed for deployment to Heroku. It needed some additional file changes to enable the two deployments (to dev and prod) to work in parallel, described a little more below:
 - Added two new Dockerfile.web files for the Heroku deployment of the frontend and backend apps. 
 - Needed to rename some deployment scripts to include `.dev.`, where there is a divergence between dev and prod environments.
- the line `\\c record_expunge` in the sql scripts doesn’t work in the heroku-postgres environment. Removed it; which is nice because it also removes the redundant database name. Instead the database name gets specified as a command-line arg to the `psql` command.
 - the backend’s Database connection code is generalized to work in both prod and dev, depending on the env variables that have been defined to specify database connection info. 

Some remaining design issues that should be addressed in another PR:

 - inconsistent treatment of environment variables especially for the database access info. Heroku’s connection between the api app and database relies on env variables so we should still do likewise for dev. But the Makefile, .env file, and config/postgres/postgres.env all redundantly define some of these variables. Gross.
 - inconsistent filenames for dev and not-dev config files. E.g., all `Dockerfile` names should be renamed to `Dockerfile.dev`, which in turn needs an update to the make targets which must refer to those files.


wittejm:
This PR could def use some cleaning up, but it *should* work without breaking anyone's dev environment, despite several refactors. 

If reviewing, it'd be awesome if you pull this branch and see if the docker stack in your dev environment still builds. And if you want to then try pushing the app to heroku, that's worth testing too and I can send you the heroku login credentials I'm using.

PR #340, opened 2019-08-29
React router

ForrestLonganecker:
Added Links to the logged-in navbar: url now populates with the correct information.

added Routes for Stats/Admin/Account components


maxkwallace:
@ForrestLonganecker awesome! Will take a look at the code soon.

src/frontend/src/components/LogOut/index.tsx
```diff
@@ -20,6 +21,7 @@ class LogOut extends React.Component<Props, State> {
     this.props.logOut();
     event.preventDefault();
     event.stopPropagation();
+    window.location.pathname = '';
```
maxkwallace: Do you know if there's any way (or a "correct" way) to do this through an official React Router API? 

If not, I think we should have a comment here explaining why this update is necessary, since we generally want to avoid modifying global variables at any point in the code.

src/frontend/src/index.tsx
```diff
@@ -13,7 +14,9 @@ import store from './redux/store';
 const rootElement = document.getElementById('root');
 ReactDOM.render(
   <Provider store={store}>
-    <App />
+    <BrowserRouter>
+      <App />
+    </BrowserRouter>
```
maxkwallace: I don't know a lot about React Router yet.

https://medium.com/@ivanmontiel/changing-old-habits-in-react-router-4-8483ee5e087c says:

> With the new BrowserRouter and NativeRouter components, you don’t have to worry about setting up your history.

> If you’ve worked with React Router 3, you’ve probably written something like this: <Router history={browserHistory}> ...

We have this code in App/index.tsx:

>       <Router history={history}> 
>          ....

Should we get rid of the `history={history}` part there?

src/frontend/src/components/LogOut/index.tsx
ForrestLonganecker: Thanks for catching this! I thought that had merged out when I pulled in the other changes. The use of _History_ takes care of this I've dropped the line and can re-submit.

Tested to make sure the behavior of clearing to the base url when logging out.

src/frontend/src/index.tsx
ForrestLonganecker: After tinkering with it when I tried to remove all references to history the router started throwing errors about props.history being undefined. I believe that how it is currently set up works well enough for now. This is the first time I have tried working with react-router and I believe @cjbreaux set it up that way. Do you have any ideas on this subject Chris?

ForrestLonganecker: But it also works by simply not including the _BrowserRouter_ component so I will just drop that one as well!

maxkwallace: Hmm okay :/ 

Ideally we'd understand why or why not BrowserRouter (and the history reference) are or are not needed. 

But I understand that's not always 100% possible and I'm fine with merging if the observed behavior is an improvement on what we have now.

maxkwallace:
@ForrestLonganecker looks good, go ahead and merge!

cjbreaux:
@ForrestLonganecker I'll take a peek at it tonight!

src/frontend/src/index.tsx
ForrestLonganecker: I will definitely look into the difference more as well, I completely agree that it is better to understand why we would choose one approach over the other.

The approach that I was learning and applying used _BrowserRouter_ , Chris got it working without it so I feel that it's best to move forward with what he place. 

I'll go ahead and merge this in and we can change things to _BrowserRouter_ In the future if it appears to be optimal.

PR #341, opened 2019-08-29
Feature/search results redux

SteelFace87:
This PR is to indicate the current path and state of implementing Redux, and specifically implementing Redux for the searching and displaying of the search results.

I propose that all components that subscribe to the Redux store be placed under a container folder.  
If you go to the src/container/AllSearchComponents.tsx file I have made some comments on how I plan on implementing Redux for the search functionality.  

Currently when AllSearchComponents component mounts, it triggers a fetch that adds mock data to Redux, and map state to props will grab the records array into the props of the container component AllSearchComponents.   

I propose that the RecordSearch component gets rendered within the AllSearchComponent and gets the fetch passed down to it as a prop.  Then the fetch functionality to add records to Redux can be triggered by creating a new search. 

I am waiting on feedback before continuing down this road.


maxkwallace:
@SteelFace87 Thanks! Will take a look soon.

maxkwallace:
@SteelFace87 Thanks for pushing this up! Here's my overall thoughts:

I'll definitely have more feedback when the code becomes more polished and gets closer to merging. And I can't promise that some things won't need to be changed later. But the basic approach of having a container component for passing search results as a prop plus presentational components for displaying results and triggering the search action seems fine.

Because this is draft code and not final, it can be a bit hard to walk the line between being overly critical at an early stage versus erring on the side of giving lots of feedback to possibly prevent the need for more changes later. 

I'm a bit confused by what your goals are for how you're using Redux in the current code. Though I also wouldn't expect you to have full comments-- because, again, this is draft code-- I'm just concerned that the action structure here is overly complicated for what we need to do.

It looks like you're using Redux Thunk and mapDispatchToProps? How does that fit in to your goals for the feature? In this answer https://stackoverflow.com/a/35415559/1067145 Dan Abramov says:

> don’t use anything (including thunks) if you don’t have the genuine need for them.

and I'm not sure there's a genuine need here. We're just loading some stuff into the store. 

- - - 

#### container/

Without knowing exactly what you mean by "container/" vs. "components/" Components, I'm going to assume you're following the approach discussed at:

- https://redux.js.org/basics/usage-with-react#presentational-and-container-components
- https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

Is that right? I'm fine with having a `container/` folder for Components that follow that pattern. But this is a small project and we want to keep things simple so I don't see the need for making it forbidden for Presentational Components to connect to the Redux Store. I think we can make that change later it this is a big pain point or lots of people want it.

The `container/` folder should have the name structure as the current `components/` folder, so if I made a Foo Container Component, it'd be at `container/Foo/index.tsx`. 

#### Smaller stuff

- I'd prefer if AllSearchComponents.tsx didn't have "components" in the name since it's redundant.

- In recordexpungPDX/src/frontend/src/service/get-records.ts the request should be a GET (and if it's not, that needs to be fixed in the backend)

- Why did you get rid of the records/types.ts file for redux/types.ts but leave redux/system/types.ts? This is confusing. I think we should continue to have "feature-specific" types.ts files in the same folders as the actions and reducer files.

SteelFace87:
The pattern that I have always used with Redux to load data to the store is by using some type of middleware to handle async calls and mapDispatchToProps.  Since Redux Starter Kit comes out of the box with Thunk middleware installed that is the route that I went.   I will do some research and go through those articles to see if I can simplify the logic.  

As far as the Container, I will not use it if you feel that we shouldn't enforce it.   The container has been nice since if you want to see what is connected to the store you can just open it up.  Having some but not all things that connect to the Store in the Container folder may be more confusing.  

As far as the smaller stuff, I will make those changes.

I hope to have something back to you late this upcoming week, or early the following.

maxkwallace:
Regarding the container convention, I am open to being persuaded, but I am mostly concerned about increasing boilerplate and complexity / the learning curve for new people who may not be familiar with all of the React conventions. If a majority of people want to enforce the container convention then I'm fine with it. To your point about seeing what's connected to the store, would something like this work? I realize it's not perfect. Could also grep for the `connect(` call itself.

```
mkw:frontend/src/components$ grep connect $(find . -type f) | grep redux
./App/index.tsx:import { connect } from 'react-redux';
./LogOut/index.tsx:import { connect } from 'react-redux';
./LogIn/index.tsx:import { connect } from 'react-redux';
./RecordSearch/index.tsx:import { connect } from 'react-redux';
```

maxkwallace:
@SteelFace87 to your point about Redux and patterns, I think patterns are great! I just want the code to be relatively self-explanatory if possible, so I would need to be convinced that using the pattern is the best way to solve the problem and is something we can easily explain to people in the future. "this is what I've always used" is hard to explain to new people. Code comments are one way to explain things, and I think linking to high-quality articles talking about the pattern is also really good since you can just link those to new people in the future and have them read them.

I just took a quick look at the code again. Based on https://daveceddia.com/redux-mapdispatchtoprops-object-form/ would it be possible to change:

```
const mapDispatchToProps = (dispatch: Function) => ({
  fetch() {
    //change loadSearchRecordsMockAction to loadSearchRecordsAction when get back end running
    return dispatch(loadSearchRecordsMockAction());
  }
});

...

export default connect(
  mapStateToProps,
  mapDispatchToProps
)(AllSearchResults);
```

to just

```
export default connect(
  mapStateToProps,
  //change loadSearchRecordsMockAction to loadSearchRecordsAction when get back end running
  { fetch: loadSearchRecordsMockAction }
)(AllSearchResults);
```

? This confused me a bit when I first glanced at the PR.

SteelFace87:
Ok, I did some experimenting with not using Thunk, and I still feel Thunk would be a good solution for the following reasons, however, if you still feel it will be easier then I will remove the Thunk.


1)  If someone ever feeds the connect function a 2nd argument, this.props.dispatch won't be available and the code will break.  It took me a long time to troubleshoot this when I was removing Thunk from my code.  A good article can be found here.  [Missing dispatch](https://react-redux.js.org/using-react-redux/connect-mapdispatch#why-is-my-component-not-receiving-dispatch)
2)  I find it confusing that when you have synchronous code it is dispatched differently than asynchronous code if we chose to not use Thunk.  I could see this leading to errors if someone gets confused about whether or not a function is synchronous.  I grabbed these snippets from the article you provided, it also goes into detail about this issue.
Below is asyncronous with Thunk
`
this.props.dispatch(showNotificationWithTimeout('You just logged in.')) 
`
Vs. asyncronous with out Thunk
`
showNotificationWithTimeout(this.props.dispatch, 'You just logged in.')
`
3)  Redux Starter Kit comes standard with Thunk.  

4) If we chose to go the container route in the future, and not use Thunk, then we will need to pass down dispatch as a prop to all child components

As far as using grep for connect calls I'm fine with that.

Finally, if we chose to go with Thunk and mapDispatchToProps, then I will update the connect to
`
export default connect(
  mapStateToProps,
  //change loadSearchRecordsMockAction to loadSearchRecordsAction when get back end running
  { fetch: loadSearchRecordsMockAction }
)(AllSearchResults);`
to prevent confusion

maxkwallace:
@SteelFace87 just making sure-- was there new code you wanted me to review? I don't see any new commits since August 28th.

maxkwallace:
@SteelFace87 to answer your most recent comment about Thunk:

Sorry if my earlier comments weren't totally clear. What I meant to say was not "don't use Thunk" but "from reading the code in this PR, it's not clear to me why Thunk is necessary and leads to better code". My hope was that it would be possible to-- regardless of whether we use Thunk or not-- update the code so that it's easier to understand why it's written the way it is.

If you think it's better to use Thunk then I agree that that's the best approach to move forward with in this PR. I am pretty agnostic about whether or not we use Thunk, just that anytime there is a new pattern or paradigm introduced, the additional complexity needs to be justified or explained so that new folks can understand it easily.

Due to time constraints, my own lack of expertise, and because this was draft code I'm sorry I wasn't able to pinpoint the exact sources of confusion during my first review. I don't have time to reread the code this minute but from what I remember it seemed like there was some ambiguity re: separation of concerns and naming to distinguish Redux Actions from Redux Action Creators. I should be able to take another look at the meeting tonight.

maxkwallace:
Olli and I just discussed this in-person at the meetup and had a really productive discussion, leaving a quick comment to keep the PR up-to-date. Decided to move ahead with using Thunk and using the "Object shorthand form" of mapDispatchToProps (see https://react-redux.js.org/using-react-redux/connect-mapdispatch#two-forms-of-mapdispatchtoprops), also had a few quick discussions about stuff in the diff.

SteelFace87:
I did the following updates and the PR is now ready for review.
1)moved the types.ts file into the records folder
2)moved all of the logic into the records folder. Previously I had created another folder since I wasn't sure what the original logic in the records folder was for.
3)renamed AllSearchComponents container to AllRecords to make the name less confusing
4)changed mapDispatchToProps to object shorthand form.

**NOTE
now that the apiService from src/service/api-service is being updated with an additional parameter of dispatch, this will mean that the file from src/service/get-records.ts will need to be updated when that PR #383  goes through.

maxkwallace:
@SteelFace87 awesome! Just as an FYI, I likely won't be able to get to it until tomorrow.

maxkwallace:
screenshot

@SteelFace87 just checking, is this what the user is supposed to see?

Also, when I fill in the fields with what I believe are valid values and click the search button, nothing happens. Need some more info on the status of this PR before I can review it again.

maxkwallace:
Also, I'm seeing this when I run `npm start`:

./src/containers/AllRecords.tsx
  Line 5:  'loadSearchRecords' is defined but never used  @typescript-eslint/no-unused-vars

here's the top of the output from `git log` (to make sure we're looking at the same code);

```
commit 00b48d4a6119808d2c078f44f7481ecbda47ddbf (HEAD, olli/feature/search_results_redux)
Author: ollinieminen <otn987@gmail.com>
Date:   Fri Sep 20 08:42:24 2019 -0700

    moved types.ts folder into records folder from redux folder.
```

SteelFace87:
> 
@SteelFace87 just checking, is this what the user is supposed to see?
> 
Yes, this is not the end result that we are after, however, I am trying to get vertical and implement the next step in a following PR.   The AllRecords container that currently only renders "Console Log Current Records" is now connected to the store and maps the most recently fetched records to props, along with a function that will fetch a new set of records.

In the next PR I am looking to import the record search component into the Container and pass down the fetch functionality to the record search component as a prop.  Then we can display the records in the search results component (the original purpose of this PR) by importing it and passing down the most recently fetched records to it as a prop.    In this process, I plan on removing the "Console Log Current Records" since the search results component will be displaying the records.

> 
Also, I'm seeing this when I run npm start:

./src/containers/AllRecords.tsx
Line 5: 'loadSearchRecords' is defined but never used @typescript-eslint/no-unused-vars

here's the top of the output from git log (to make sure we're looking at the same code);
> 

I am expecting this to be the case, the loadSearchRecords is currently not being used because I am using a mock function to load mock record data to redux.  Once we get the back end to talk to the front end, I plan on removing the mock function and using the unused loadSearchRecords function to fetch real data from the database.

maxkwallace:
@SteelFace87 Alright, I think breaking this up into smaller PRs is great, almost always helpful to do that. That said, while I can't speak for anyone else, I'm not okay with merging this while:

- "Console Log Current Records" appears on the main page after logging in, like in the screenshot
- "Line 5: 'loadSearchRecords' is defined but never used @typescript-eslint/no-unused-vars" shows up when running `npm start`

These are obviously issues that should be addressed, they are not "correct" behavior, and they're not difficult to fix, so it's a bit inconsiderate to other devs to merge them in, since they will raise questions and/or cause confusion. I can take a look at the code again once you've cleaned them up.

The basic idea is that we don't want to "break the build" in the `master` branch for anyone else who's also working on this project. Having work-in-progress code in the `master` branch is fine, but it should clearly be commented as such and should not interfere with other folks' work.

maxkwallace:
@SteelFace87 to clarify, let me know when you've address the issue discussed in my previous comment, and I can take a look at the code again afterwards.

SteelFace87:
I have now implemented the following updates.
1) Import AllRecords Container to App and have AllRecords container now render RecordSearch component.
- Now that the container is connected to the store, we will be able to pass records down to the SeachResults component (this has not been created yet), without the need of having to connect the SearchResults component to the store also.
- We can now also pass down the fetch functionality to RecordSearch component without the need of connecting it to the store and have this triggered when a new search is created.  **Note this has not been implemented yet.  I am currently running a mock function to get fake data upon componentDidMount of the AllRecords Container itself.   
2) I have removed loadSearchRecords.  This was due to the following reasons
- Because we are currently using a mock function, loadSearchRecords was not being used and caused a @typescript-eslint/no-unused-vars" when calling npm run start.
- Due to all the changes to the apiService and reducers, there are now so many merge conflicts that it will be easier to just recreate the loadSearchRecords when we are ready to transition from the mock function.
3) I have removed the "console log current records" which was previously rendered onto the screen by AllRecords container

src/frontend/src/containers/AllRecords.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+class AllRecords extends Component<Props> {
+  componentDidMount() {
+    this.props.fetch();
+    //Next we need to pass down the fetch to RecordSearch component as a prop, and trigger it
+    //on a new search instead of componentDidMount.  When this happens we shoud replace the loadSearchRecordsMock function with a real api call
+    //and use the populated fields in the RecrodSearch component as paramaters for the API request.
+  }
+  render() {
```
maxkwallace: Needs a newline between these function definitions https://github.com/airbnb/javascript#whitespace

src/frontend/src/redux/records/types.ts
```diff
@@ -1,17 +1,3 @@
...
-//
-// Pasted from the UI:
-//
-// Time: Eligible now
-// Type: Eligible 137.225(5)(b)
-// Charge: 4759924B - Poss Controlled Sub 2
-// Disposition: Convicted
-// Convicted: 2/12/1987
-// Case: ZA0061902
-// Case Balance: None
```
maxkwallace: @SteelFace87 why'd you delete this comment?

src/frontend/src/service/api-service.test.ts
```diff
@@ -1,3 +1,4 @@
+//NOTE, JSON data has changed, this file need to be updated
```
maxkwallace: Needs a space before `NOTE` -- https://github.com/airbnb/javascript#comments--spaces

This comment itself is really unclear. By "JSON data" do you mean the format of the mock record objects?

src/frontend/src/redux/records/types.ts
```diff
@@ -22,18 +8,15 @@ export interface Record {
...
 }
 
-// These constants are used as the 'type' field in Redux actions.
-export const LOAD_RECORDS = 'LOAD_RECORDS';
-
-interface LoadRecordsAction {
-  type: typeof LOAD_RECORDS;
-  records: Record[];
+interface SearchRecordsAction {
+  type: string;
```
maxkwallace: Can you keep this as the typeof the constant used to signal this action? That way makes it much clearer for someone reading the code to see the relationship.

src/frontend/src/redux/records/types.ts
```diff
@@ -22,18 +8,15 @@ export interface Record {
...
 }
 
-export interface RecordsState {
+export const LOAD_SEARCH_RECORDS = 'LOAD_SEARCH_RECORDS';
+
+export interface SearchRecordState {
   records: Record[];
 }
 
-// These constants are used as the 'type' field in Redux actions.
```
maxkwallace: Why'd you get rid of this comment?

src/frontend/src/redux/records/types.ts
```diff
@@ -22,18 +8,15 @@ export interface Record {
...
-interface LoadRecordsAction {
-  type: typeof LOAD_RECORDS;
-  records: Record[];
+interface SearchRecordsAction {
+  type: string;
+  search_records: Record[];
 }
 
-// Add other Action types here like so:
-// export type RecordActionTypes = LoadRecordsAction | OtherRecordsAction;
```
maxkwallace: Why'd you get rid of this comment?

src/frontend/src/redux/records/reducer.ts
```diff
@@ -1,22 +1,20 @@
...
-): RecordsState {
+  state = initalState,
+  action: SearchRecordsActionType
+): SearchRecordState {
   switch (action.type) {
-    case LOAD_RECORDS:
-      // The new state is the records returned in the
-      // action. We ignore existing records and do not
-      // necessarily include them in the new state. This
-      // is a "destructive update".
```
maxkwallace: Why'd you get rid of this comment?

maxkwallace:
@SteelFace87 also, thanks for including a comment summarizing the changes you made to the PR! Very helpful.

src/frontend/src/containers/AllRecords.tsx
SteelFace87: Fixed.

src/frontend/src/service/api-service.test.ts
SteelFace87: Space has been added and comment is now more clear.  Yes, the response from the API no longer looks like what the test implies.

src/frontend/src/redux/records/types.ts
SteelFace87: Originally I made another types.ts file since I wasn't sure if the original file was intended for what I was doing, and I didn't want to alter it.  I then replaced the original folder with what we currently have, and overlooked the comments. Comments are back in.

src/frontend/src/redux/records/reducer.ts
SteelFace87: Originally I made another reducer.ts file since I wasn't sure if the original file was intended for what I was doing, and I didn't want to alter it.  I then replaced the original file with what we currently have, and overlooked the comments. Comments are back in.

src/frontend/src/redux/records/types.ts
SteelFace87: Originally I made another types.ts file since I wasn't sure if the original file was intended for what I was doing, and I didn't want to alter it.  I then replaced the original file with what we currently have, and overlooked the comments. Comments are back in.

src/frontend/src/redux/records/types.ts
SteelFace87: Done

SteelFace87:
** Update.  All changes have been made as per the notes above.  Finally, I also noticed that in api-service.test.ts file there was a missing dependency for configureMockStore which was added on PR #383.  I went ahead an installed the missing dependency that was causing an error.

Issue #91, opened 2019-04-11
Create "Users" Component for /admin


maxkwallace:
This task is to create the Component that's comprised of the second-from-the-top card on the page with the title "Users", including transferring all the HTML from http://dev.huntermarcks.net/admin.

Wireframes for more context around the whole users flow:
![user-account-wireframe](https://user-images.githubusercontent.com/6887982/69019529-a78c4480-0976-11ea-853d-4c69937e0deb.png)

Accessible validation technique: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19

We don't necessarily need to implement routes and/or URLs for the React app (it can be a "single page app", at least to start) but if there are routes, or you want to implement one, this Component should appear on the /admin route. I used "/admin" in the title of the issue only because it's a straightforward way to reference the mockup.

The actual functionality of interacting with the backend (responding to clicks on buttons) should be handled in a separate issue (create it if you want to work on it).

The Component should support displaying a variable-length list of Users, again matching the mockup, and you should use hardcoded data in Javascript to test this functionality if it's not yet possible to get live data from the backend. 

### This task is done when...
 - The Component described above exists and visually matches the mockup, and the user can see it when they navigate to the page.
 
If you create a PR to address this issue you're welcome to send it to me @maxkwallace for review.

maxkwallace:
@ForrestLonganecker okay! Do you have existing work in progress on this? If you're not currently working on it, would it be feasible to open this up for someone else to work on? Again, can discuss tonight, those are the main things I'd want to know re: making sure everyone has something they're excited to work on.

ForrestLonganecker:
I had started but didn't get far, of someone else wants to take over that's
okay with me. You can open it back up!

ForrestLonganecker:
@maxkwallace I had started but didn't get far, of someone else wants to take over that's
okay with me. You can open it back up!

ForrestLonganecker:
@maxkwallace to follow up I am now actively working on this ticket!

ForrestLonganecker:
Just to update the state of this issue, currently the `<UserList/>` component renders data that it retrieves from the store.

Some of what still needs to be done:
- create a get/users request that will return a list of users and their information if the current user is an admin and place that list in the store.
- render something/redirect users that do not have admin privileges.

ForrestLonganecker:
As of #583 the user component is now fulfilling these requirements.

PR #353, opened 2019-09-01
UserList component now connected to store

ForrestLonganecker:
`<UserList />` component passed through parent `<Admin />` component for the page routing

At this point the routes are not made to pull users from the database, for the time being I am using _**loadUsers**_ action to update the store from within the `<UserList />` component to mock out how the component should appear.

Style note:
Looking this component by itself the styling at the top of it's container overlaps the nav. I tested putting the plain html for the `<Groups />` component that will reside above the `<UserList />`:

--> When `<Groups />` is in place the margin looks correct. 

We may also want to have some kind of "Loading" animation take place while waiting for results or when there are no users yet. We can talk about implementing something for all components that need to wait on data to load at some point, but is not pertinent at this moment.

current behavior:
if users --> display users
else --> display no users


ForrestLonganecker:
This should be the initial implementation of #91

PR #356, opened 2019-09-04
Endpoint users/ GET method

wittejm:
This is a backend API call to get the list of users currently in the database. 
The new code includes the necessary db operations, the endpoint, test cases, and the design spec added to our design.md document. 

This endpoint is needed for completing the frontend Users component, issue #91

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -157,9 +157,58 @@ Returns: auth token
 - `400 BAD FORMAT`: missing email or password
 
 
+**`GET`** `/api/users/`
+
+Fetches the list of existing users
+
+Required headers:
+
+- `Authorization: <JWT string>`
+
+Returns: List of users:
+
+- format: `JSON`
+- fields:
+    * users :: list
+        * email
+        * admin
+        * timestamp
+
+Status codes:
+
+- `200 OK`
+- `401 UNAUTHORIZED`: authorization rejected; missing or invalid auth token
+- `403 FORBIDDEN`: authorized user is not admin
+
+
+**`GET`** `/api/users/EMAIL`
+
+Required headers:
+
+- `Authorization: <JWT string>`
+
+Returns: Requested user
+
+- format: `JSON`
+- fields:
+    * email
+    * admin
+    * timestamp
+
+Status codes:
+
+- `200 OK`
+- `401 UNAUTHORIZED`: authorization rejected; missing or invalid auth token
+- `403 FORBIDDEN`: authorized user is not admin
+
+
 **`POST`** `/api/users/`
 
 Creates a user
+
+Required headers:
+
+- `Authorization: <JWT string>`
 
 `POST` body:
 
@@ -189,24 +238,14 @@ Status codes:
 - `422 UPROCESSABLE ENTITY`: duplicate user or password too short
 
 
-**`GET`** `/api/users/EMAIL`
-
-Returns: Requested user
-
-- format: `JSON`
-- fields:
-    * email
-    * admin permissions
-    * timestamp
-
-Status codes:
-
-- `200 OK`
-
 
 **`POST`** `/api/search`
 
 Performs search of remote system
+
+Required headers:
+
+- `Authorization: <JWT string>`
 
 `POST` body:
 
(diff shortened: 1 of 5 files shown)


doc/design.md
```diff
@@ -157,10 +157,59 @@ Returns: auth token
...
+
+Status codes:
+
+- `200 OK`
+- `401 UNAUTHORIZED`: authorization rejected; missing or invalid auth token
+- `403 FORBIDDEN`: authorized user is not admin
+
+
+**`GET`** `/api/users/EMAIL`
+
```
erikjasso: Using the user's email address as the unique identifier for lookup in  a URL feels weird (as a login name it's fine). I'd rather see a user id or username here. Not adamantly opposed, just want to discuss it before it's implemented

wittejm: Yea, I agree it's weird and think it's worth revisiting. This particular endpoint isn't implemented in this PR, and I'm also unsure how we would even use this, i.e. what is the frontend component that would fetch only a single user's info? Apart from when that user is logging/ logged in, in which case the login endpoint provides this information.

NickSchimek: I agree with @erikjasso as I'm not opposed to it, but I believe its common practice to use a user id for this action.

src/backend/expungeservice/endpoints/users.py
```diff
@@ -51,5 +51,24 @@ def post(self):
 
         return jsonify(response_data), 201
 
+    @admin_auth_required
+    def get(self):
```
NickSchimek: Perhaps we can name this index as it is indexing all users? I think the name get is very generic as it could refer to `get_users`, `get_user`,  `get_user_new`, or `get_user_edit`?

doc/design.md
wittejm: In our spec, email is the only public-facing identifier for a user because we eliminated the "username" field, deciding it was functionally redundant with just using an email address, and the user_id field is a UUID generated by the database. 

I can pull this design question into a new issue since it's not in this PR

src/backend/expungeservice/endpoints/users.py
wittejm: This function is called get because it's the GET http request of the /api/users/ endpoint, and the MethodView class in Flask for defining endpoints needs to use the same name. 

I agree it's all a little too generic and needs redesign. I think the present users.GET to get all users seems fine but users.POST to create a new user is confusing. This deserves its own issue to hash out a naming scheme.

doc/design.md
ForrestLonganecker: I have seen this done as an `/api/users/search` as a **post** where you send the required information along in the request ie: Email. That way you don't have to expose the UUID to the front end.

I agree that it could be beneficial to spec out naming conventions and requisite data for the endpoints at our next meeting.

src/backend/tests/database/test_database.py
```diff
@@ -66,6 +66,37 @@ def test_get_user(self):
...
+            SELECT * FROM USERS;"""
+        self.database.cursor.execute(verify_query)
+
+        verify_rows = [r._asdict() for r in self.database.cursor.fetchall()]
+
+        assert len(verify_rows) == len(users_get_endpoint_result)
+
+        for (email, hashed_password, admin) in [
+        (r['email'], r['hashed_password'], r['admin']) for r in users_get_endpoint_result]:
+            self.verify_user_data(email, hashed_password, admin)
```
NickSchimek: What is this for loop for after the assert statement?

src/backend/expungeservice/endpoints/users.py
NickSchimek: I see. Should we write down all of our endpoints (URI's), and document how they're going to be routed so that we can use some naming conventions that imply intent?

src/backend/tests/database/test_database.py
wittejm: It manually checks that the entire returned set of entries from the endpoint match what's in the db.

src/backend/expungeservice/endpoints/users.py
wittejm: Yes, lets.

wittejm:
Merging this. this GET endpoint needs to be refactored a bit, but it's fairly minor and this endpoint may be useful for now for working on frontend integration.

PR #357, opened 2019-09-05
Do not analyze charges without a disposition

### LINKED ISSUE #291: Unable to sort convictions for cases with missing dispositions
## Description
Unable to sort conviction charges, because one of the dates is a string. Closed cases with partial dispositions are creating charges that do not have a disposition

### Current behavior
The current behavior is that charges for without a disposition are being marked as missing disposition and are being skipped in the expungement process, however this is only working when a case does not have any disposition transactions.

See PR #255 for when this behavior was added.

### Expected behavior
Any charge that does not have a disposition should be skipped regardless if the case has disposition transactions. This is needed for closed cases that have disposition transactions for some, but not all charges.

### Stack trace
```
Traceback (most recent call last):
  File "cli.py", line 117, in <module>
    expunged = expunger.run()
  File "/recordexpungPDX/src/backend/expungeservice/expunger/expunger.py", line 55, in run
    self._set_most_recent_convictions()
  File "/recordexpungPDX/src/backend/expungeservice/expunger/expunger.py", line 94, in _set_most_recent_convictions
    self.convictions.sort(key=lambda charge: charge.disposition.date)
TypeError: '<' not supported between instances of 'str' and 'datetime.date'
```

### Possible fix:
One issue is that there is a mutable default in the base_charge initializer.
https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/charge_types/base_charge.py#L12
This should be replaced with None and then add a guard for creating a disposition.

### This task is done when:
 - Use case fixture 'CASE_WITH_PARTIAL_DISPOS` to create a system test that exposes bug
 - Create unit test that also exposes bug
 - Any charges that do not have dispositions should be tagged as such and skipped by the time analysis.

NickSchimek:
## Description

Excluding charges without a disposition from the analyzer and marking them as needs further investigation. This is being done because the disposition date cannot be used to run a time analysis when there is no disposition.

Sometimes cases that are closed will not have charges with dispositions for various reasons.

- They can be improperly labeled
- They can be withheld
- Simply not added, because "everyone" knows that the charge is not eligible therefore disposition does not matter.... and possibly other f-reasons

Added additional test case to expose main issue, and removed default disposition class from initializer as it returns a singleton class in python which can cause unintended behaviors.

This could "possibly" be further improved by using the null object pattern or if you want to get fancy? NLP.

### Changes made

Adding one commit at a time so that you can see the failure/passes on CI

Added test case for when a record only contains charges without a disposition.

Removed mutable object from base_charge initializer.

Updated test to conform to the new base_charge initializer.

Check if there is a `disposition` instead of a `disposition.date` or `disposition.ruling` since disposition is now `None` as it should be when there is not a disposition.

This fixes #291

### DIFF
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -67,7 +67,7 @@ class Expunger:
 
     def _tag_skipped_charges(self):
         for charge in self.charges:
-            if charge.skip_analysis() or Expunger._dispositionless(charge):
+            if Expunger._dispositionless(charge) or charge.skip_analysis():
                 self._skipped_charges.append(charge)
 
     @staticmethod
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -67,7 +67,7 @@ class Expunger:
 
     def _tag_skipped_charges(self):
         for charge in self.charges:
-            if charge.skip_analysis():
+            if charge.skip_analysis() or charge.disposition is None:
                 self._skipped_charges.append(charge)
 
     def _remove_skipped_charges(self):
@@ -94,7 +94,7 @@ class Expunger:
             self.second_most_recent_conviction = self.convictions[-2]
 
     def _assign_most_recent_charge(self):
-        self.charges.sort(key=lambda charge: charge.disposition.date if charge.disposition.date else (date_class.today() + relativedelta(years=-1000)), reverse=True)
+        self.charges.sort(key=lambda charge: charge.disposition.date, reverse=True)
         if self.charges:
             self.most_recent_charge = self.charges[0]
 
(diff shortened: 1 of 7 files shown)


NickSchimek:
Thanks for the review @wittejm It probably makes issue 259 obsolete. I will verify when I look over this PR again in the next few days (might be our hack night when I finally get to it again).

Oregon expungement laws are quite complicated which leads to logic that needs to cover all of their asinine rules.  The flow chart is inaccurate, but still serves a purpose for now... I don't really like to use the word edge case in this scenario, because that implies to me that it is a rare occurring issue, however this is a common theme because a lot of Oregon records are incomplete, and the state just now realizes that their past record collection policies may prevent them from rolling out (or wreak havoc on) future tech projects, which is the main reason why an expunger, in Oregon, is a complicated project.

I'm hoping someone (maybe @aemrich ?) will want to learn, and pair with me, so that they too can learn about Oregon expungement rules and the current expunger design so that I'm not the only one that knows the ins and outs of this code. Also I plan to roll out a document that details the design decisions and implementation, although before I do that I would like the expunger to be a bit more stable.

Issue #360, opened 2019-09-07
Spec out the required backend endpoints


wittejm:
Only a few endpoints are listed in the design.md document and have been implemented, but the click-through prototype http://dev.huntermarcks.net/ shows we need several new endpoints, and may need to refactor the data tables.

### This task is done when:
 - required endpoints identified, including function signatures and usage in the frontend, with a clear naming scheme

Issue #119, opened 2019-05-09
Design the data collection for search results and stats reporting


wittejm:
The app should save some of the user activity and search results so that we can track the app’s overall usage and impact over time. 

We'll need to sufficiently anonymize the data which may mean not tracking an individual record as a unit in the database. It is easy to de-anonymize a record based purely on the collection of cases and charges grouped together in that record.

Includes the design for #49 , but also includes the data persistence ops performed during search and the associated database tables. 

Some of this information is included in [this document ](https://docs.google.com/document/d/1_X6RLU9K9_QmPGaheVXu0ZEXPTVCCbSLU-_WEhF7AZo/edit#heading=h.j5wg95sanplr) linked from the project wiki

### This task is done when:
 - Database tables and functions have been designed for the persisted data. 
 - Information added to design doc. The existing Analyses table should be removed from the database and design doc.
 - After the PR is merged, one or more issues have been written for implementing the designed features.

wittejm:
I wrote some more initial ideas for this feature here:
https://github.com/codeforpdx/recordexpungPDX/blob/master/doc/design.md#database-functions

Unassigning myself from this task for now.

wittejm:
Closed with #576

PR #361, opened 2019-09-08
New Endpoint Design Spec

wittejm:
Major expansion to the Design spec for the backend. I tried to identify every feature in the Frontend Prototype http://dev.huntermarcks.net/ that needs an API call. 

I included large Question tags because I see several non-trivial issues that need to be worked out. 

Adds 5 new endpoints that support the following app features:
- construct OECI login workflow and the change-password workflow
- activate a new account via registration email and choosing a first password
- change a forgotten password

Also refactors some existing endpoints.

Addresses issue #360


wittejm:
It may be helpful for readability to display the "rich diff" of the document. 

Asking for a lot of reviewers because it involves a number of high-level app design decisions including for the  frontend and Crawler.

I would love to go over this at our next meeting (Sept 11th)

doc/design.md
```diff
@@ -273,6 +293,121 @@ Returns:
...
+Is it a webpage redirect, like the frontend prototype?
+
+Or is it an authentication serverside, like the current Crawler implementation?
+These behaviors need to be reconciled.
+
+I see three design versions:
+
+- Rely on a frontend page redirect, which obtains login access from the OECI directly to allow subsequent requests by Crawler. This requires a Crawler redesign, which currently passes creds to the remote site. It also requires caching a login token provided by OECI. Does the OECI login allow that, e.g. use a cookie?
+
+- Roll the login  into the `/search/` endpoint.
```
NickSchimek: I believe we had settled upon this version, and we could set a cookie on the client side that contains encrypted credentials that will get sent with each request. The cookie should expire after a very short time period.

doc/design.md
```diff
@@ -273,6 +293,121 @@ Returns:
...
+If so, we need to add oeci credentials to the **`POST`** `/search/ ` body. And also cache the login credentials clientside so we can re-send them on every search (bad).
+
+- Send oeci login credentials to the backend once, and the /search/ endpoint uses them for every /search/ call. That means caching the OECI login creds in the backend.
```
NickSchimek: No, caching credentials on the backend sounds just as bad as caching them in the front end. Plus that means the backend needs to hold a state for each user, which does not sound very restful.

doc/design.md
```diff
@@ -273,6 +293,121 @@ Returns:
...
+What's the workflow for accessing OECI?
+
+Is it a webpage redirect, like the frontend prototype?
+
+Or is it an authentication serverside, like the current Crawler implementation?
+These behaviors need to be reconciled.
+
+I see three design versions:
+
+- Rely on a frontend page redirect, which obtains login access from the OECI directly to allow subsequent requests by Crawler. This requires a Crawler redesign, which currently passes creds to the remote site. It also requires caching a login token provided by OECI. Does the OECI login allow that, e.g. use a cookie?
```
NickSchimek: We will not be holding onto the cookie (at least in the backend) sent back from OECI for multiple requests. We can create our own cookie to set on the frontend that can be used for each request. I'm not sure but maybe we could send the OECI cookie to the client to send back on further requests, but I prefer using our own cookie so that we have more control and less dependencies. FYI: the OECI cookie expires after ~5mins of inactivity.

doc/design.md
```diff
@@ -317,17 +452,22 @@ Database Schema
...
-    clients (uuid, first_name, last_name, dob, date_created, date_modified), uuid primary key
-
     result_codes (uuid, code) uuid primary key
 
     rules (uuid, text)
 
-    analyses (client_id, case_id, result_code, statute, date_eligible, rules[], date_created, date_modified, expunged, date_expunged)
+## Question:
+
+Do we store client information? I think we have at some point decided not to.
```
NickSchimek: This is not really our decision to make. Of course it would make some things easier... But ultimately the client needs to sign off

doc/design.md
```diff
@@ -157,7 +157,7 @@ Returns: auth token
 - `400 BAD FORMAT`: missing email or password
 
 
-**`GET`** `/api/users/`
+### **`GET`** `/api/users/`
```
ForrestLonganecker: need to add in the response:
- name
- group

wittejm: Fixed! thanks.

doc/design.md
```diff
@@ -273,6 +295,121 @@ Returns:
     * tbd
 
 
+### **`POST`** `/oeci_login` ?
+
+## Question
+
+What's the workflow for accessing OECI?
```
maxkwallace: Based on what Nick already said, I'm assuming OECI auth looks like this:

- send username/password credentials in a request
- In the response, OECI server sets a cookie which authenticates future requests
- The cookie expires after 5 minutes of inactivity

If this is not how OECI works then my comments below should possibly be disregarded. Additional info on OECI auth if applicable would be helpful.

After reading through the below, I'm not sure I totally understand the proposed designs, but here are my thoughts:

If OECI auth uses a cookie, redirecting the user to the OECI login form/page won't work because the Javascript running in our app (on a different URL/domain) will never be able to read the cookie, and thus unable to make its own requests to OECI or pass that cookie to the backend to do so on its behalf.

AFAIK the only feasible approach is for the user to input their credentials into our React app, have that app send the credentials to the backend, and then have the backend send the credentials to OECI, then get the cookie back if they're correct. 

The backend can then send the cookie back to the frontend to send with future requests (preferred) or store it on the backend (seems cumbersome).

doc/design.md
```diff
@@ -273,6 +295,121 @@ Returns:
...
+
+Returns: auth token usable as a regular POST Authorization header.
+
+- format: `JSON`
+- fields:
+    * auth_token
+    
+### **`POST`** `/api/change_user_settings`
+
+Alter user account settings, like password, email address, and admin access, or account information like Name or Group
```
maxkwallace: Following REST, this is typically done with a PUT to /api/user

doc/design.md
```diff
@@ -273,6 +295,121 @@ Returns:
...
+- `5xx` of some kind if the email-send failed
+
+
+### **`POST`** `/api/update_password`
+
+Follows a password reset link (from a sent email), that provides login authorization
+
+## Question: How does this work?
+
+A password reset link provided in an email needs to be a GET request, so the endpoint requires a different method for yielding auth access. If this method is a POST and auth-protected, we need to first obtain an auth_token via the **`GET`** `/api/auth_token` method below.
```
maxkwallace: Not sure I totally understand the approach suggested here. This would be good to discuss at the next meeting.

Because we are using a JWT for authentication, and the JWT needs to be stored in our React app in a certain way for the frontend to think it's authenticated, Javascript needs to be involved before our app can become authenticated in the usual way.

doc/design.md
```diff
@@ -123,7 +123,7 @@ Admin page for creating users
 Backend Endpoints
 -----------------
 
-These endpoints comprise our API. All requests of these endpoints go through the web server.
+These endpoints comprise our API. All requests of these endpoints go through the frontend web server and get routed to the backend server with the nginx reverse-proxy.
```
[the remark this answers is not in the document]
wittejm: thanks, this is clearer. I'll put that in.

doc/design.md
```diff
@@ -194,40 +205,52 @@ Returns: Requested user
...
-**`POST`** `/api/users/`
+## Question:
 
-Creates a user
+Is 404 still appropriate if it's due to a query param and not a path?
+Or is this a  `422 UPROCESSABLE ENTITY` as though it was a field in POST body?
 
-Required headers:
 
+### **`POST`** `/api/new_user/`
```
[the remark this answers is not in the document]
wittejm: sounds good. This change was following Nick's comment that the various user function names are too similar, but yay for following convention.

doc/design.md
wittejm: Ok that's what I had believed. Wanted to bring this up again before removing these tables from the design doc, but I think we can go ahead.

doc/design.md
```diff
@@ -238,14 +261,13 @@ Status codes:
 - `422 UPROCESSABLE ENTITY`: duplicate user or password too short
 
 
+### **`POST`** `/api/search`
 
-**`POST`** `/api/search`
+Performs search of remote OECI system.
```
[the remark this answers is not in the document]
wittejm: This endpoint will include recording some info about the search being made (e.g. possibly the user_id), when we implement our "stats" feature. I don't think that needs to be complicated. Issue for determining the exact behavior is  #119

doc/design.md
wittejm: Cool. I'm well past due to go and learn RESTful design.

doc/design.md
wittejm: Ah, yep I see what I was leaving out here. I agree, let's go over this in a meeting.

doc/design.md
wittejm: From what I can tell on the Crawler side, it looks like the flow is to instantiate the Crawler, call crawler.login(oeci_username, oeci_password), and then can make subsequent search calls. when the crawler gets reinitialized (i.e. with every /search/ from the frontend), it needs to run crawler.login() again. 

That means we never directly / explicitly do anything with an OECI cookie.

@NickSchimek does this sound right?

So the details to be worked out on exactly how the oeci (encrypted) credentials get passed back and forth with the oeci_login response and with the  /search/ request/response, but it does look like the login-then-search crawler operations need to be run on every /search/

Pinging @ash-tamraz because he's working on /search/, I'm wondering what you think?

wittejm: @ForrestLonganecker and I are looking into best practices with this workflow , and low-level implementation of cookie usage both on the frontend and backend.

doc/design.md
wittejm: Cool. I just posted a new response on this issue to Nick's comment above, here: https://github.com/codeforpdx/recordexpungPDX/pull/361#discussion_r324431898 , to zero in on what works with both the Crawler and our desired frontend workflow.

(This is my attempt to "merge" the comment "branches" on this PR :) )

wittejm:
@NickSchimek @maxkwallace @ForrestLonganecker I'm going to reject this PR soon (and leave it up for now so folks have a chance to see it), make a smaller merge to reflect only the completely resolved design changes . I can create new issues for the unresolved workflows because it sounds like we have narrowed down some of those discussions pretty well.

doc/design.md
maxkwallace: IMO there actually isn't a ton to know about REST. Main principle that applies here is to use PUT for idempotent state changes and POST for non-idempotent state changes, both excepting cases where DELETE makes more sense. 

I am following the Rails convention which was designed based on REST. In Rails, you can just say `resource :user` in the `routes.rb` file and it will automatically generate RESTful routes for you for the "user" resource.

doc/design.md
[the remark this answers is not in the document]
maxkwallace: Yeah it should probably still be a GET in that case. There's always side effects (logging, etc.) but people still use GETs.

doc/design.md
wittejm: Ok, to correct myself from earlier:  looking more closely at the crawler, I see that the oeci cookie is set to the crawler's Session object by a successful oeci login request. It's then sent automatically with subsequent requests. That's all just handled by python's Requests package, #TIL.

PR #363, opened 2019-09-09
Backend Users object: add Name and Group_name fields 

wittejm:
Just adds the two columns "name" and "group_name" to the Users table and the endpoints that read/insert Users.


aemrich:
Seems good to me.

PR #367, opened 2019-09-11
fix the default credentials insert into DB

wittejm:
Fixed to match recent DB change.

Now the dev environment has an initial admin and regular user for testing:
email: `user` , password: `user`
email: `admin` , password: `admin`


ForrestLonganecker:
Looks good to me!

Issue #379, opened 2019-09-18
Add Redux Middleware to add authorization header for applicable requests


maxkwallace:
See my comment at https://github.com/codeforpdx/recordexpungPDX/pull/371#discussion_r325462866

Note: https://auth0.com/blog/secure-your-react-and-redux-app-with-jwt-authentication/

Waiting on https://github.com/codeforpdx/recordexpungPDX/pull/371 to be merged and then I'll start this.

ForrestLonganecker:
Just letting you know that #371 is now merged

maxkwallace:
@ForrestLonganecker yup, I'll work on this tonight.

PR #383, opened 2019-09-19
Add Request/Authentication Middleware

maxkwallace:
Addresses https://github.com/codeforpdx/recordexpungPDX/issues/379


maxkwallace:
Other reviewers/questions are welcome as well, I only requested Forrest for now since this PR is sort of a follow-up to https://github.com/codeforpdx/recordexpungPDX/pull/371 and I don't want to request too many people since obv there's stuff going on elsewhere in the app as well. Lots of work to do!

@SteelFace87 FYI this PR changes the API for `apiService` so it now accepts `dispatch` as the first argument in addition to the request data. Open to hearing any questions/ideas you have, this was the best way I could figure out to handle authentication. Other than that, should not affect your work in https://github.com/codeforpdx/recordexpungPDX/pull/341

cc @wittejm 

Other possible (rejected by me for this PR) approaches to authentication:
- The approach used in https://auth0.com/blog/secure-your-react-and-redux-app-with-jwt-authentication/ -- seems like too many breaking changes from what we have now, and not sure if it's actually a good idea to isolate Axios request/response objects from application code
- Importing the store in api-service -- https://daveceddia.com/access-redux-store-outside-react/ -- since it basically turns the store into a global variable

The main change this code introduces is making `apiService` accept `dispatch` as the first argument. This encourages putting calls to `apiService` in Action Creators rather than Components. It doesn't have to be necessary to pass in `dispatch` for API calls that don't require authentication but it seems better to have a consistent API; I don't think the lack of flexibility with `apiService` here is much of a disadvantage.

src/frontend/src/redux/system/actions.ts
```diff
@@ -1,10 +1,21 @@
-import { LOG_IN, LOG_OUT, LogInData } from './types';
+import apiService from '../../service/api-service';
+import history from '../../service/history';
+import { LOG_IN, LOG_OUT } from './types';
```
ForrestLonganecker: These changes look great!

Just so I understand, for a request that requires authentication we would include:
`authenticated: true` 
In the request, would we be pulling that value from the component and passing it in similar to:
`logIn(email, password, authenticated)`

src/frontend/src/redux/middleware/request.ts
```diff
@@ -0,0 +1,112 @@
...
+// This Action and Middleware is a sort of implementation detail of the API Service needed
+// to allow API Requests to interact with the application state via the Redux store.
+export interface RequestAction {
+  type: typeof REQUEST;
+  request: Request;
+  resolve: Function;
+  reject: Function;
+}
+
+// This is a custom Middleware that does three things:
```
ForrestLonganecker: This all looks well implemented, thanks for taking the time to do this! I feel like I would like to walk through it in person at some point and once it's merged I can see if I can figure anything out about the `any` types.

src/frontend/src/redux/system/actions.ts
maxkwallace: Whether a given request needs to be authenticated or not typically wouldn't change so it'd be declared statically in the action creator like e.g. this:

```
export function getRecords() {
  return (dispatch: Function) => {
    return apiService(dispatch, {
      url: '/api/records',
      method: 'get',
      authenticated: true
    }).then((response: any) => {
```

maxkwallace:
@ForrestLonganecker sounds good! I think the simplest way to do authorization is just e.g. check whether or not the logged-in User is an admin and show different stuff on the page based on that. The backend will have to actually authorize the requests by checking the User ID in the JWT and seeing if that person's an admin or not

maxkwallace:
Gonna leave this up for a day or two before merging in case anyone else has any comments.

src/frontend/src/redux/system/actions.ts
SteelFace87: @ForrestLonganecker @maxkwallace, would it make more sense to have the dispatch as the 2nd argument?  I would imagine that there are use cases where dispatch isn't required?  This would eliminate us needed to call apiService(null, request)

maxkwallace: @SteelFace87 if dispatch weren't required for every call to `apiService` it'd definitely be better for it to be an optional second argument. However that change can't be made without introducing other tradeoffs-- I hope my other comments above and my code comments in the PR make that clear. If you have a proposal for how to do this plus an argument for why it'd be better then I'm happy to hear it. 

Just to clarify, with the current implementation, one shouldn't do `apiService(null, request)` -- it won't work, since `dispatch` is used for every request regardless of whether it's authenticated or not.

maxkwallace:
Not comments since the 19th so going to go ahead and merge this.

PR #392, opened 2019-09-22
Endpoint oeci login

wittejm:
Rejected my previous PR #378 and incorporated feedback.

This adds the new endpoint POST api/oeci_login which takes oeci credentials, checks their validity by attempting to log in to the remote site, and if successful it returns the encrypted credentials in a cookie. It uses a python library for symmetric encryption, and uses the same JWT secret key that is already stored on the flask app.

The unit tests for this endpoint avoid doing an actual login by stubbing the crawler's login method.

The PR includes some refactoring to move api/user endpoints to their correct uri, and refactors endpoint test cases to remove a lot of redundant code .

Issue #37, opened 2018-12-13
Build a project plan with milestones


quarterback:
Partner requests a project roadmap with timelines for deliverables. 

Describe the task here, including any supplemental information. 
Tag yourself, or whoever is responsible for it. 

### This task is done when...
 - [ ] a project plan is developed and team syncs on deliverables
 - [ ] plan is shared with partner agency

maxkwallace:
Closing this issue since it hasn't been updated since Dec. To anyone: feel free to reopen it if you want to work on it!

PR #393, opened 2019-09-23
Endpoint users cleanup

wittejm:
This PR has some refactoring / code cleanup without adding any new features.

- POST /api/users is renamed to /api/user, to match our design update.
- The test cases for user, users, and auth endpoints are all cleaned up by removing redundant code.

(is a subset of my closed PR #392 )


NickSchimek:
As a convention I prefer to have all endpoints that involves the same resource in one file. Otherwise if the pattern in the PR is propagated then we may end up with 5 user files each pertaining to a different HTTP request involving the same `user` resource. Could we perhaps simply rename the original `users.py` file to `user.py` ❓

wittejm:
@NickSchimek Ok, I merged the two endpoints into a single source file.py

wittejm:
I went looking into a good pattern for GET all users vs GET one user and in this Flask documentation 

https://flask.palletsprojects.com/en/1.1.x/views/

they use the same MethodView class with the URI /users/. So... I think I'll refactor this again, put it all back into a shared /users/ endpoint rather than separated the way they are now. Fortunately I think the frontend is only connected to GET /users/ so far, so no added work there. @ForrestLonganecker , is that right?

Also applies to issues #37

ForrestLonganecker:
That should be correct, and the get uses endpoint is not yet integrated
into the frontend so it won't cause issues there either!

Issue #401, opened 2019-09-24
Endpoint: /api/oeci_token


wittejm:
This endpoint allows a user to "log in" to the remote OECI site by submitting their username and password for the OECI system, and the endpoint checks that the credentials are valid and, if so, returns them to the client encrypted in a cookie. No "logged in" state is maintained on the backed. Subsequent OECI record search relies on the encrypted cookie to log in again to the OECI site before performing the search.  

### This task is done when:

**POST** /api/oeci_login 
- Is user_authorization protected
- Request JSON body contains oeci_username and oeci_password.
- Check credentials are valid with the crawler.login() function
- login() == False result returns 401: unauthorized.
- Successful login returns 201 and returns the encrypted credentials in a cookie. 
- Uses a python library for symmetric encryption, and uses the same secret key that is already stored on the flask app for JWT.

Prior design discussion is on the closed PR #361

PR #405, opened 2019-09-25
oeci_login endpoint

wittejm:
API endpoint that accepts and validate's a user's oeci login credentials. #401 describes the feature implemented here. 

Other implementation notes:
 - added the cryptography package as a new dependency. the JWT package we're using for auth tokens doesn't support encryption features.
 - increased the size the app config's JWT secret key so that it's valid to be used for the Fernet symmetric encryption package. 
 - the test code stubs the crawler.login() call so that it doesn't attempt an actual log in to the remote site. 

I started the cookies portion of this code in PR #378  and incorporated @maxkwallace 's  feedback.

### DIFF
--- a/src/backend/expungeservice/endpoints/oeci_login.py
+++ b/src/backend/expungeservice/endpoints/oeci_login.py
@@ -1,0 +1,62 @@
+from flask.views import MethodView
+from flask import request, make_response, current_app
+import time
+from cryptography.fernet import Fernet
+import json
+import base64
+
+from expungeservice.crawler.crawler import Crawler
+from expungeservice.endpoints.auth import user_auth_required
+from expungeservice.request import check_data_fields
+from expungeservice.request.error import error
+
+
+class OeciLogin(MethodView):
+
+    @user_auth_required
+    def post(self):
+        """
+        Attempts to log in to the OECI web site using the provided username
+        and password if successful, encrypt those credentials and return them
+        in a cookie. If the credentials
+        """
+
+        data = request.get_json()
+
+        if data is None:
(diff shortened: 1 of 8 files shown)


src/backend/tests/endpoints/test_oeci_login.py
```diff
@@ -0,0 +1,62 @@
...
+
+        credentials_cookie_string = self.client.cookie_jar._cookies[
+            "localhost.local"]["/"]["oeci_token"].value
+
+        jwt_key = base64.encodebytes(self.app.config.get("JWT_SECRET_KEY"))
+        cipher = cryptography.fernet.Fernet(key=jwt_key)
+        creds = json.loads(cipher.decrypt(bytes(
+            credentials_cookie_string, "utf-8")))
+
+        print(creds)
```
NickSchimek: Let's remove this if no longer needed

src/backend/expungeservice/endpoints/oeci_login.py
```diff
@@ -0,0 +1,62 @@
...
+        encrypted = cipher.encrypt(bytes(credentials))
+
+        response = make_response()
+
+        response.set_cookie(
+            "oeci_token",
+            secure=True,
+            httponly=True,
+            samesite='strict',
+            expires=time.time() + 15 * 60,
```
NickSchimek: Can this calculation be put into a separate method so that future maintainers can tell more easily what the expiration time is set for?
for example a method named `fifteen_minutes` would be more clear. `expires=fifteen_minutes()`

PR #408, opened 2019-09-26
Dev docker suite

Rosanio:
This PR repurposes the `make dev` command to spin up a development environment using docker-compose. The database, backend and frontend are each run in their own container. Additionally, there's a separate NGINX container which runs on and exposes port 80, forwarding requests to the appropriate service.

To spin up the dev environment, you will need to have [docker](https://docs.docker.com/install/) and [docker-compose](https://docs.docker.com/compose/install/) installed locally. Once the docker daemon is running, navigate to the root directory of the project and run `make dev`. Once the containers are spun up, you can navigate to [http://localhost](http://localhost) to validate that the frontend service is reachable, and [http://localhost/api/hello](http://localhost/api/hello) to see the backend service. You can also modify the source code of the frontend or backend to validate that changes are picked up in real time.

**TODO**
- Travis seems to be failing, so that should be fixed
- README's need to be updated with instructions on how to spin up the dev environment
- The `network` blocks in the `docker-compose.dev.yml` file may not be necessary, they should be removed to see if everything still works without them


wittejm:
@Rosanio Thanks for the PR! I pulled your branch and got this setup working, I just needed one fix: I had to specify the dockerfile name in the build context for `expungeservice` and `db`. E.g. the db section was changed to:
`build:
            context: ./config/postgres
            network: host
            dockerfile: Dockerfile.dev
        `

I recently put a new suffix on those two dockerfile names because I started setting up Heroku, which was scooping up all the files named "Dockerfile" in its deploy step. 

It looks like the Travis errors are from the build script using `docker stack` on your new compose file. We definitely only want to use just one of `stack` vs `docker-compose`. I started looking into the difference here : https://vsupalov.com/difference-docker-compose-and-docker-stack/ ,

Since we're only using container orchestration for the dev environment, docker-compose seems like the way to go.

wittejm:
Also I needed to install docker-compose, for which I followed the official doc: https://docs.docker.com/compose/install/

Rosanio:
Thanks Jordan! I pushed a commit to fix the incorrect Dockerfile mapping. Looks like you're working on the Travis issues, so I'll leave that to you.

wittejm:
I found that the Pipfile conflict was resolved with a simple `git pull -r upstream/master` , and showed up because of my oeci_login branch merge from yesterday.

PR #417, opened 2019-09-30
 crypto module refactor

### LINKED ISSUE #401: Endpoint: /api/oeci_token

wittejm:
Pulls the encryption/decryption code used for holding oeci credentials into a new module. Simplifies the oeci endpoint code and will also simplify the search endpoint. 

This new DataCipher class can be used to symmetrically encrypt/decrypt anything that is json-serializable. Anyone need anything encrypted?

Closes #401


NickSchimek:
Thanks for doing this! I will review it tomorrow

PR #418, opened 2019-09-30
Some docs on backend testing with the dev docker environment 

wittejm:
This documentation clarifies how to do python testing and dev using the improved docker setup.

Also, I wanted to avoid breaking too much expected behavior from our prior setup, so this also ensures that the prior setup with pipenv still works. 

I'd like to see us migrate more fully to the docker setup because it should make all of our lives easier, but I'm not willing to tear out all of our old pipes just yet, so for now I'm trying to stand with my feet in two puddles, so to speak. 

Requested some specific reviewers (generally those already doing some backend) but if you work on this project please try the new setup and ping me on slack if you're having trouble.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,7 +142,3 @@ config/postgres/data
 
 #VS Code Files
 .vscode
-
-#Second Pipfile copy
-/Pipfile
-/Pipfile.lock
--- a/.gitignore
+++ b/.gitignore
@@ -142,3 +142,7 @@ config/postgres/data
 
 #VS Code Files
 .vscode
+
+#Second Pipfile copy
+/Pipfile
+/Pipfile.lock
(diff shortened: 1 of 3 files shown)


NickSchimek:
I don't feel comfortable with the code changes in this PR. Specifically copying the pipfile to a different location and gitignoring is not an ideal approach for an installation that is not going to be supported. Preferably if someone is going to run a local setup they should be working off the same pipfile so that it stays updated with any changes made by the dev. Here the changes are only copied in one direction. Can we remove the changes in `.gitignore` & `Makefile` and just submit the doc changes for now? I may have a different solution I can address in a separate PR. Although if everyone is having success with the new docker setup I'd say to remove the `make install` and `make test` commands completely.

wittejm:
@NickSchimek I agree with the changes you're suggesting. I wanted to prop up the old setup in some way but it does make sense that enforcing "the one true way"  now will make our lives easier rather than harder.

wittejm:
Reverted the code changes!

PR #419, opened 2019-09-30
Updated docker doc

wittejm:
Started updating this doc a bit. It needs a bit of proofreading and is missing some sections still, but if anyone wants to offer feedback on what's here, I would love to hear about any points that are confusing or new questions you have. 

This is basically a follow-up on the new and already merged features on #408, #416. Testing the new setup would be a great way to check the strength/weakenesses of this doc. Thanks already to @SteelFace87 for this.  

Related to but a slightly separate concern from #418 .

There is some complementarity between this doc and the section in the main Readme, and I'm not sure of the best organization or redundancy to have here. Documentation is hard, y'all.

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -1,111 +1,163 @@
 Record Expunge Docker Setup
 ===========================
 
-This document is deprecated.
+Our project uses Docker to containerize a local dev environment and for deploying the app to production. This document describes the general setup for each of our three environments: dev (local), test (travis-ci), and prod (heroku).
 
-Documentation
--------------
+General Documentation
+---------------------
 
 [Docker Tutorial](https://docs.docker.com/get-started/)
 
 
-Containers
-----------
 
-We will have three containers: a container for the web server and webapp; a container for the database; and a container for our service.
+Dev Environment Containers
+--------------------------
 
-Each container will have a Dockerfile that describes the container's contents. For conveninence any complex/multi-part command will have a shorter Make target.
+There are four containers which are configured in /docker-compose.dev.yml , and managed using docker-compose. The project's main README.md includes the commands to manage the containers locally.
 
-We will have two configurations:
+ - **webserver** npm dev server
+ - **expungeservice** wsgi server (hosting a Flask app)
+ - **db** postgres database
+ - **nginx** proxies /api/ paths to the backend and everything else to the frontend.
 
-- `dev` for development
-- `prod` for production
+Docker volumes sync the project's source code directories with the frontend and backend dev servers, so that the containers don't need to be rebuilt to reflect code changes, only new package dependencies.
 
-`dev` and `prod` containers will run concurrently, with each `webserver` and `database` containers running on distinct ports. This will allow us to deploy both to the production server and run tests w/o impacting the production database.
+Test Environment Containers
+---------------------------
+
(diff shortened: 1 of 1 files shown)


doc/docker.md
```diff
@@ -1,111 +1,163 @@
...
-5. To stop a single *service* in the running stack, run:
+4. To stop a single *service* in the running stack, run:
 
         docker service rm recordexpungpdx_*
 
-replacing * with the service name e.g.  webserver or expungeservice. This stops the corresponding docker container and also prevents the docker orchestrator from spinning up a replacement container. This is useful if you want to, for example, run the node server for frontend development, since the node daemon automatically and rapidly propagates source code changes.
+replacing * with the service name e.g.  webserver or expungeservice. This stops the corresponding docker container and also prevents the docker orchestrator from spinning up a replacement container.
 
-6. To explore the database:
+5. To explore the database:
```
NickSchimek: I'd suggest changing this to `To start a PostgreSQL interactive terminal`

And you may want to mention that `\q` exits

doc/docker.md
```diff
@@ -1,111 +1,163 @@
...
+make dev_up
+```
+
+Thist first builds the dev docker images if they don't already exist. It runs the docker stack in detached mode, so you can follow the container of logs with:
+
+```
+make dev_logs
+```
+
+While the dev stack is running, each service is accessible in the dev environment at `localhost:<port>`, at the exposed ports defined in the docker-compose.dev.yml file.
```
Rosanio: I'm actually thinking of removing the port mappings; I remember Olli had an issue while trying to access the frontend via port 3000. Maybe just mention that the frontend is accessible at `http://localhost` and you can hit the APIs via the `/api` path.

doc/docker.md
```diff
@@ -1,111 +1,163 @@
...
 
-We will have three containers: a container for the web server and webapp; a container for the database; and a container for our service.
+Dev Environment Containers
+--------------------------
 
-Each container will have a Dockerfile that describes the container's contents. For conveninence any complex/multi-part command will have a shorter Make target.
+There are four containers which are configured in /docker-compose.dev.yml , and managed using docker-compose. The project's main README.md includes the commands to manage the containers locally.
 
-We will have two configurations:
+ - **webserver** npm dev server
```
Rosanio: I'm being nit-picky, but it's technically the webpack dev server.

doc/docker.md
```diff
@@ -1,111 +1,163 @@
...
+To see which containers are running (and with the optional flag to see stopped containers also):
+
+
+        docker ps [-a]
+
+
+Running the Test containers
+---------------------------
+
+Because of the differences between the `dev` and `test` stacks, a different set of Dockerfiles and commands are used to build and run the test containers. You can do this for local testing as well as with the automated scripts on Travis. The deployment uses `docker stack` commands rather than `docker-compose`. This might be worth changing?
```
Rosanio: We could consider using `docker-compose` when running this with Travis, for the sake of simplifying things.

doc/docker.md
wittejm: I already found it useful to at least expose the database for making a local connection, and if we expose each it can allow interacting with each without relying on nginx. I should def clarify /expand the wording here.

doc/docker.md
wittejm: Cool, let's talk about it

Issue #229, opened 2019-07-02
Export Record data as json for /search/ endpoint.


wittejm:
The Record object and its members, including its Charge, Case, ExpungementResult, and Disposition objects, need to be converted to json data to return in a user search result. 

I see two general ways to do this: one is a programmatic approach that scoops up all of the public variables and properties on each of these objects. This could be done with a subclass of [JSONEncoder](https://docs.python.org/3/library/json.html ) recursively doing something like this:  

    import inspect
    result_data = {key: value for (key,value) in inspect.getmembers(record_obj) if key[0] != '_']

If the data objects are refactored so that everything that isn't a public field or a property (including getter/setter functions, etc.) has a leading '_' then this can return the exact data object we need. With @NickSchimek 's new PR #224 then the new Charge factory pattern could produce objects that look like this.

The other approach is manually exporting the list of relevant fields on each object, to construct a json object with only the explicitly named set of fields, evaluated (in the case of dynamically computed fields) at export time. I think the former programmatic approach would be cleaner and more maintainable. 

Is part of issue #48 

Includes @NickSchimek 's suggestion of refactoring data objects to use the property pattern.

### This task is done when:
 - Record, Charge, Case, ExpungeResult, and Disposition object data can be exported to JSON with exactly the fields required by frontend. 
 - design.md updated to reflect the format of the returned json data.

wittejm:
Related question: the Charge object includes secondary dynamically computed properties like acquitted(), recent_conviction() and others. Should any/all of those be included in the returned json? In this case the members of the returned data should all be listed in the design.md Data Model section. 

Pinging @NickSchimek and frontend folks @maxkwallace @hmarcks @cjbreaux  but all opinions welcome.

maxkwallace:
@wittejm I have a few opinions here. If I understood correctly (correct me if wrong), we're asking the general question: how do we serialize server-side objects into JSON for the client? (Sidenote: almost every webapp has to solve this problem, so there are many existing solutions.) The two general proposed solutions are:

A: By default, serialize all public fields and properties on objects of the relevant types. 

B: For each property/field on an object of a relevant type, opt-in to having it serialized.

[...]

PR #434, opened 2019-10-03
Search Endpoint 

### LINKED ISSUE #48: Endpoint: /api/search
Endpoint for searching the remote site. The web app will `POST` to this endpoint with search params in the body, and the endpoint will make a request of the remote site.

Data format TBD.

wittejm:
Implements the search endpoint to fetch record data from the OECI system. Some design thoughts:

- The publicly returned contents of the Record object are enumerated manually for conversion to JSON. any changes to the Record object need to be reflected in the ExpungeModelEncoder before those changes will be exposed in the API. 
- The /api/search method is **POST** because we'll be processing and saving some non-trivial stats, e.g. the number of eligible/ineligible cases. That will be implemented with #48 . 
- The "errors" field in the Record object refers strictly to expunger data errors, not HTTP errors or code bugs.  
- Even though the crawler.search() has `middle_name` and `birth_date` as optional parameters, I made them are required fields in the search POST body. It's a hassle to check if the optional JSON fields are there and conditionally pass them into the search function as keywork arguments. This means that if a user of the frontend doesn't provide middle name or birth date, then the api request will need to pass an empty string for each. Alternative suggestions welcome.
- The unit test for the record object serializer doesn't do much: it just makes sure that it turns a Record into json. The encoder object just replicates the spec from /doc/data_results.json, so I'm not sure that we can add much in the way of interesting/useful tests. 
- The unit test for the search endpoint mocks the crawler's `login` and `search` methods so that we don't hit the oeci website. 
- I could add some negative test cases for the search endpoint, for the various error codes that can get returned. Those tests all seem fairly trivial, though, given the simplicity of the endpoint code itself.

I think that as we go forward with the project we'll want to make sure the whole app continues to work with manually testing that it works with the oeci page. E.g. we could write an interactive script that hits the endpoints with manually entered credentials and search parameters. Like Nick's CLI, but it also will be a test for the endpoints. 

Closes #48 and #229 . Provides the backend requirement for #87 .


src/backend/expungeservice/endpoints/search.py
```diff
@@ -0,0 +1,58 @@
...
+
+        decrypted_credentials = cipher.decrypt(request.cookies["oeci_token"])
+
+        crawler = Crawler()
+
+        login_result = crawler.login(
+            decrypted_credentials["oeci_username"],
+            decrypted_credentials["oeci_password"],
+            close_session=False)
+
```
NickSchimek: We should check if the login was successful before attempting to do search. I would advise against assuming the login was successful as there are various reasons it could fail even with the correct username and password. For example: The e-courts site could be taken down for maintenance, or it could be offline for other reasons.

src/backend/expungeservice/endpoints/search.py
```diff
@@ -0,0 +1,58 @@
...
+        record = crawler.search(
+            request_data["first_name"],
+            request_data["last_name"],
+            request_data["middle_name"],
+            request_data["birth_date"])
+
+        if record.cases:
```
NickSchimek: There is no need to check if the cases array is empty. The expunger doesn't break if it is passed a record object with an empty array of cases.  I'd recommend to remove this line.

src/backend/expungeservice/serializer/serializer.py
```diff
@@ -0,0 +1,60 @@
...
+            "name": case.name,
+            "birth_year": case.birth_year,
+            "case_number": case.case_number,
+            "citation_number": case.citation_number,
+            "location":  case.location,
+            "date": case.date,
+            "violation_type": case.violation_type,
+            "current_status": case.current_status,
+            "charges": [self.charge_to_json(charge) for charge in case.charges],
+            "balance_due_in_cents": case.balance_due_in_cents,
```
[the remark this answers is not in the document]
wittejm: Woops. I did notice a mismatch here and I should have brought attention to it first: are we sure we don't want to include balance due per case? This seems like separate design question so I can pull it out of the endpoint for now.

src/backend/tests/endpoints/test_search.py
```diff
@@ -0,0 +1,72 @@
...
+        self.crawler = CrawlerFactory.setup()
+        self.crawler.logged_in = True
+        self.mock_record = {
+            "john_doe": CrawlerFactory.create(self.crawler)
+        }
+
+
+    def tearDown(self):
+        pass
+        search.Crawler.login = self.login
```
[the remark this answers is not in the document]
wittejm: Ah, it was working correctly because `pass` is just a no-op and the following lines still run. Obvs I should take it out though.

src/backend/tests/endpoints/test_search.py
```diff
@@ -0,0 +1,72 @@
...
+        assert(response.status_code == 200)
+        data = response.get_json()
+
+        """
+        Check that the resulting "record" field in the response matches what we gave to the
+        mock search function.
+        (use this json encode-decode approach because it turns a Record into a dict.)
+        """
+        assert data["record"] == json.loads(json.dumps(
+            self.mock_record["john_doe"], cls = ExpungeModelEncoder))
```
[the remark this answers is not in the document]
wittejm: Cool. I wasn't too happy about my testing approach so I'm glad to go over it again.

src/backend/expungeservice/serializer/serializer.py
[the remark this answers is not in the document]
NickSchimek: We want to include balance due per case

src/backend/tests/endpoints/test_search.py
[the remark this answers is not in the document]
NickSchimek: ah thanks.

src/backend/expungeservice/serializer/serializer.py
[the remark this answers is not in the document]
wittejm: Oh, woops, gotcha.

wittejm:
Fixed those things. If this looks good apart from improving the test cases, alright if we go ahead and merge? that seems ok to take care of later and meanwhile this endpoint can become accessible for the frontend.

src/backend/tests/endpoints/test_search.py
```diff
@@ -0,0 +1,71 @@
+import json
+
+from expungeservice.endpoints import search
+from tests.endpoints.endpoint_util import EndpointShared
+from expungeservice.crawler.crawler import Crawler
```
NickSchimek: Let's remove this unused import. It's not needed to do the stub.

src/backend/expungeservice/endpoints/search.py
```diff
@@ -0,0 +1,60 @@
...
+            request_data["last_name"],
+            request_data["middle_name"],
+            request_data["birth_date"])
+
+        expunger = Expunger(record)
+        expunger.run()
+
+        response_data = {
+            "record": record
+        }
```
NickSchimek: This should be wrapped in a data object. It doesn't look like it's being done here.
```
response_data = {
     "data": { "record": record }
     }
```

NickSchimek: https://jsonapi.org/format/#document-top-level

wittejm: Sure. None of our other endpoints follow this pattern though, so I'd be in favor of updating all of them at the same time to match. Maybe at the next meeting we could talk about specifications we want our app to be following? This is the first I've seen/heard of this spec.

NickSchimek: OK. The frontend folks are expecting the results returned by this endpoint to be wrapped in a data object as it's shown in the doc. https://github.com/wittejm/recordexpungPDX/blob/1ea990b729b90804c4cfa95a5f98697e8549bca2/doc/results_format.json#L2

Issue #441, opened 2019-10-04
Open a new cursor for a database operation


wittejm:
Psycopg closes a database cursor by default when it commits a transaction.And our database wrapper code only opens a new cursor when the database connection is created, at the beginning of a request.

This hasn't been a problem because our app doesn't make two sequential database updates in a single endpoint. But if it did, it would throw an error because the second op would be trying to use a closed cursor. 

This is currently slowing down the endpoint unit tests because every one creates a few temporary user accounts during setup. If we could reopen a cursor on sequential db update operations, we wouldn't need to fetch a new database connection each time.

Our options to fix this are to either create and then close a cursor on each operation, or keep the cursor open after commiting a transaction. This could use more looking into. I know there are DB standard practices that would apply here and that I'm not aware of. 

### This task is done when:
 - The app can handle multiple transactions in a single endpoint.

PR #442, opened 2019-10-04
Endpoint users get authorization fix

wittejm:
This makes a small change to the /api/users GET endpoint: a user is now authorized to get a user's data only if the requesting and requested user ids match, or if the requesting user is an admin. This matches the design doc. 

Fixed the design doc to meet our recent design change, so that there is no "user" endpoint, only a "users" endpoint which covers all the relevant operations. 

Also tried to improve organization of the endpoint test code a bit. One problem though is that the unit tests are now even slower. I made a new issue for this: #441

### DIFF
--- a/src/backend/expungeservice/endpoints/auth.py
+++ b/src/backend/expungeservice/endpoints/auth.py
@@ -71,9 +71,8 @@ def authorized(f, admin_required, *args, **kwargs):
         if admin_required and not user_data['admin']:
             error(403, 'Logged in user not admin')
 
-        # no endpoint code uses the logged in user_id so this is commented for
-        # now. This may change.
-        # g.logged_in_user_id = user_data['user_id']
+        g.logged_in_user_id = user_data['user_id']
+        g.logged_in_user_is_admin = user_data['admin']
 
         return f(*args, **kwargs)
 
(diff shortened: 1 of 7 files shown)


src/backend/expungeservice/endpoints/auth.py
```diff
@@ -71,9 +71,8 @@ def authorized(f, admin_required, *args, **kwargs):
         if admin_required and not user_data['admin']:
             error(403, 'Logged in user not admin')
 
-        # no endpoint code uses the logged in user_id so this is commented for
-        # now. This may change.
-        # g.logged_in_user_id = user_data['user_id']
+        g.logged_in_user_id = user_data['user_id']
+        g.logged_in_user_is_admin = user_data['admin']
```
NickSchimek: hmmm lets talk Wednesday about the potential security issue here. It's possibly a low security issue, however there are "best" practices to use to check if current user is an admin.

NickSchimek: I did some more reading up on JWT's and this is actually the common way to use them although it is a bit strange to me. Anyway's not super important, but we can talk about it if there is time.

PR #446, opened 2019-10-06
Add/search results

SteelFace87:
The goal of this PR is to get vertical with a SearchResults component and then add the missing functionality to the SearchResults component in another PR.  A future PR will also replace the mock fetch function with an API call that will fetch records from the back-end.  This is the first step for creating a fully functional SearchResults component that fulfills the needs described in issue #87.  This PR includes the following additions/modifications.

1) fetch function has been passed down to RecordSearch component and is now triggered when a new search is submitted.
- The fetch function is still running a mock search so no matter what a user places into the field, the redux store will be updated with the API response in the mock function.  In the future, the mock function will be replaced with a function that does a fetch from the back-end.

2) The main tag was lifted from RecordSearch component and is now in AllRecords container.  This is because both RecordSearch and SearchResults components should both be wrapped within the same main tag, this removes redundant code.

3) Cases component was added and handles mapping through each of the case instances and creates a complete list of cases. This cases list is what is rendered by the SearchResults component.

4) Case component was created in order to assist Cases component to create the list of cases.

5) Record type was modified to comply with the new format of the API response.

6) Mock actions were updated to reflect the new API response format

7) Reducer was updated so that the initial state format would match what it gets replaced with when the fetch function is invoked by RecordSearch component


src/frontend/src/components/SearchResults/Case.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
maxkwallace: Why are this Component and Cases in the SearchResults directory instead of their own directory? I don't think we do this anywhere else in the app right now. Is there a good reason to deviate from convention here?

src/frontend/src/components/SearchResults/Case.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface Props {
+  case: {
+    name: string;
+    case_number: string;
+    birth_year: number;
+    balance_due: number;
+  };
+}
```
maxkwallace: This is essentially a type definition for a Case, and so it should probably be in recordexpungPDX/src/frontend/src/redux/records/types.ts

src/frontend/src/containers/AllRecords.tsx
```diff
@@ -4,27 +4,25 @@ import { AppState } from '../redux/store';
...
-    //on a new search instead of componentDidMount.  When this happens we shoud replace the loadSearchRecordsMock function with a real api call
-    //and use the populated fields in the RecrodSearch component as paramaters for the API request.
-  }
-
   render() {
-    // We still need to create and import SeachResults component and pass it records down as a prop so that it can render new records
-    // This component should be rendered below the <RecordSearch /> coponent below.
+    //Fetch has now been passed down to RecordSearch componant and is triggered when submit button is pressed.
+    //Currentlly no matter what the user puts in the search fields, it still triggers a mock fetch with pre determined API response
+    //Once fetch adds records to Redux, a SearchResults container is rendered with updated records
```
maxkwallace: Need spaces after the `//`s when beginning the comments.

SteelFace87: Ugh I missed these! I just fixed these

src/frontend/src/components/SearchResults/Case.tsx
SteelFace87: Done

src/frontend/src/containers/AllRecords.tsx
SteelFace87: Ugh I missed these.  Thanks!

src/frontend/src/components/SearchResults/Case.tsx
SteelFace87: fetch has not been named fetchRecords for clarification.  Also, Cases and Case components have been moved to their own directory.

PR #449, opened 2019-10-08
Feature/status

SteelFace87:
The purpose of this PR is another vertical slice to add functionality to SearchResults component as mentioned on issue #87 

This PR adds the functionality of displaying the status of the current search by implementing the following additions/modifications.

1) Create a NoSearchResults component that is rendered when there aren't currently any search results to display.

2) Create a Spinner component that renders while an API call is being made to fetch records.

3) Create an AllStatus container that connects to the Redux store and uses mapStateToProps to determine whether the spinner or the NoSearchRersults component should be rendered.

4) Add a time delay to the mock action to mimic the time it takes for an API call to fetch records.  This way it is possible to test that the spinner functionality works.

5) Update the records reducer so that an action creator can be dispatched when an API call is triggered.  This will update the loading status of the API call which is used to determine whether a spinner or a NoSearchResults component is to be displayed.  

6) Update the initial state of the records reducer to have a loading status.


src/frontend/src/containers/AllStatus.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type Props = {
+  loading: boolean;
+};
+
+class AllStatus extends React.Component<Props> {
+  render() {
+    return (
+      <section>
+        {this.props.loading ? <Spinner /> : <NoSearchResults />}
```
ForrestLonganecker: This all looks great! One small thing is that the app will render "No search results found." before a search has been executed. I don't think that is super important at this point in time though!

SteelFace87: I was thinking about the same thing.  If we wanted tp workaround that we would have to add a bit more logic which is not a big deal.  "No Search Results Found" however is not a false statement so I left it as is unless we feel that this is a change worth doing.

PR #452, opened 2019-10-08
Db commit after request

### LINKED ISSUE #441: Open a new cursor for a database operation

wittejm:
Previously, the database users.create() operation ran a commit right away. I realized this isn't good and instead all the database ops in an endpoint should be treated as a single transaction. 

In this refactor, the database commit() runs only at the end of an endpoint request, and only if that request is returning without any errors.  

This also let me simplify the endpoint test code further, so that's included. Unit tests are a little faster now. 

Closes #441. This doesn't use the solution I proposed there, because that the wrong idea. If a request encounters any errors, it shouldn't persist any changes to the database. 

This runs a commit() even if no transactions are pending, in which case it has no effect.

### DIFF
--- a/src/backend/expungeservice/app.py
+++ b/src/backend/expungeservice/app.py
@@ -4,7 +4,7 @@ from .config import app_config
 
 # Add new endpoint imports here:
 from .endpoints import hello, auth, users, oeci_login, search
-from .request import before, teardown
+from .request import before, after, teardown
 import logging
 
 from .loggers import attach_logger
@@ -30,6 +30,7 @@ def create_app(env_name):
     search.register(app)
 
     app.before_request(before)
+    app.after_request(after)
     app.teardown_request(teardown)
 
     app.logger.debug("Flask app created!")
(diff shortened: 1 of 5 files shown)


erikjasso:
Looks good. I like that you've made creating a user atomic. Postgres can be configured to commit automatically but I like that it's explicit

PR #459, opened 2019-10-09
Endpoint /api/users PUT (update user)

### LINKED ISSUE #375: Endpoint: /api/users/<user_id> PUT
endpoint to update a single user's profile or account settings. 

### This task is done when:
 - fields in request JSON: all optional, but requires at least 1 of: email, name, group, admin, password
 - returns a User data object containing user_id, email, name, group, admin, date_created
 - user auth_token required. additional check for either admin authorization or that auth_token user_id matches user_id in request.
 - endpoint registered with Flask app
 - unit tests
 - update date_modified in database

wittejm:
Implements the PUT  endpoint for Users data, and required database ops. 

Highlights:
- Relies on a new SQL function, USERS_UPDATE, which requires all of these arguments: user_id, name, email, group_name, admin, hashed_password. If the caller doesn’t want to update all of those fields, the entry’s preexisting values need to be fed back into the function. So the database operation does the following: 
  -  reads the user data given a user_id
  - verifies the user exists
  - updates the user data in-memory with new values
  - calls the SQL update function with the set of user data
- the SQL function, database update function, and endpoint all return the complete set of user data, including both the updated and unaffected fields.

Closes #375


src/backend/expungeservice/endpoints/users.py
```diff
@@ -110,6 +110,62 @@ def post(self):
...
+
+        response_data = {
+            "user_id": update_user_result["user_id"],
+            "email": update_user_result["email"],
+            "admin": update_user_result["admin"],
+            "name": update_user_result["name"],
+            "group_name": update_user_result["group_name"],
+            "timestamp": update_user_result["date_modified"]
+        }
+        return jsonify(response_data), 201
```
erikjasso: If put() modifies an existing user and does not create a new one, you should return 200. If put() creates a new user then you should return 201.

doc/design.md
```diff
@@ -238,7 +238,46 @@ Returns: New user
...
+Returns: Updated user
+
+- format: `JSON`
+- fields:
+    * user_id
+    * email
+    * name
+    * group_name
     * admin
     * timestamp
```
erikjasso: What status codes does PUT return?

wittejm: Ah, yes I've updated the status codes for PUT.

src/backend/expungeservice/endpoints/users.py
wittejm: updated!

src/backend/expungeservice/endpoints/users.py
```diff
@@ -110,6 +110,62 @@ def post(self):
...
+        data = request.get_json()
+
+        if data is None:
+            error(400, "No json data in request body")
+
+        if not any([key in data.keys() for key in ["email", "name", "group_name",
+                                 "password", "admin"]]):
+
+            error(400, "Json data must define one or more of: \
+email, name, group_name, password, admin")
```
NickSchimek: I don't see a reason to check for this. 1. if there is no data being updated it wont change anything anyway. 2. we can ensure that there is always data by passing the user data to the edit form on the frontend. That way they can see what is currently in the user profile, and make changes as necessary. Of course there is no harm in checking for this, however I don't think it needs to be here.

src/backend/expungeservice/endpoints/users.py
```diff
@@ -110,6 +110,62 @@ def post(self):
...
+            error(403, "Logged in user not admin and doesn't match requested user id.")
+
+
+        data = request.get_json()
+
+        if data is None:
+            error(400, "No json data in request body")
+
+        if not any([key in data.keys() for key in ["email", "name", "group_name",
+                                 "password", "admin"]]):
```
NickSchimek: Correct me if I'm wrong, but user_auth_required checks if there is a currently logged in user. This would allow for the user to upgrade themselves to an admin?

NickSchimek:
Thanks for putting this together @wittejm  - It looks OK to me. I think eventually we should remove the admin column and hashed_password out of the user table for security concerns. I can elaborate more on this in person, however, for this app I think it's fine for now, especially for an MVP.

src/backend/expungeservice/endpoints/users.py
wittejm: Wow, yep that is an excellent point and needs to be changed.

PR #461, opened 2019-10-10
Fix Heroku deployment issues

Rosanio:
This PR fixes a couple of issues on the recordexpungepdx heroku deployment:

- The `\api\` portion of the URL path was being dropped when requests were made to the backend. While the exact reason for this is still a mystery, the NGINX config file has been updated with a fix.
- The production Dockerfile for the backend was missing a needed directory, and using an outdated command to install dependencies.

The deployment Makefile has also been improved to speed up deployment times and handle errors better.

### DIFF
--- a/src/backend/expungeservice/Dockerfile.web
+++ b/src/backend/expungeservice/Dockerfile.web
@@ -1,10 +1,10 @@
 FROM recordexpungpdx/recordexpungpdx:expungeservicebase
 
 RUN mkdir -p /var/www/expungeservice
+RUN mkdir /var/www/logs/
 COPY . /var/www/expungeservice
 WORKDIR /var/www/expungeservice
-RUN python3.7 -m pip install -r requirements.txt && rm requirements.txt &&\
-    python3.7 -m setup install
+RUN pip3 install pipenv && pipenv install --system --deploy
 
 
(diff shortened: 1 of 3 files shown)


src/backend/expungeservice/Dockerfile.web
```diff
@@ -1,10 +1,10 @@
 FROM recordexpungpdx/recordexpungpdx:expungeservicebase
 
 RUN mkdir -p /var/www/expungeservice
+RUN mkdir /var/www/logs/
```
NickSchimek: Regarding the log files that are being created:

@wittejm It seems we’re capturing some logs that are being stored in a log file. If so let’s look into two things for the purpose of ensuring our log file stays a reasonable size (for cost and speed reasons). Perhaps you’ve already thought of this and/or have implemented an approach.

1. we create logs files for each day. And every month or two months we delete the oldest files and keep the most recent. 
Or,
2. Perhaps Heroku has a logging file that we can send the log data to that is automatically kept a certain size. Therefore eliminating the need to create our own logging directory.

wittejm: These are good ideas. I'll make an issue for improving our logging approach

PR #484, opened 2019-10-24
Oeci login integration

ForrestLonganecker:
This introduces two mock files that will allow us to attain the `OECI_token` and make requests against the `/api/search` endpoint.

Before pushing to production we need to pass the cookies with both:

```
httponly=true
secure=true
```

`Secure` cookies requires HTTPS, in development we are running things as HTTP so the cookie is lost in transmission.

`HttpOnly` cookies can not be changed by the frontend app, we will need an endpoint that will remove the `oeci_token` on logout to clear it from the browser.

I have added the `developerUtils.md` to explain how we can use the two mock files.

Mock `search.py` endpoint returns static json in the format the crawler returns in that we can start building frontend components with.


src/frontend/src/components/OeciLogin/index.tsx
```diff
@@ -1,22 +1,21 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import { AppState } from '../../redux/store';
-import { logIn } from '../../redux/system/actions';
+import { OECILogIn } from '../../redux/system/actions';
```
maxkwallace: If you grep (case-sensitive) for Oeci and OECI in our code:

```
mkw@M4600:~/Documents/repos/re-pdx/recordexpungPDX/src/frontend$ gpnifr Oeci
./src/components/OeciLogin/index.tsx:class OeciLogin extends React.Component<Props, State> {
./src/components/OeciLogin/index.tsx:            id="OeciLoginForm"
./src/components/OeciLogin/index.tsx:)(OeciLogin);
./src/components/AppRouter/index.tsx:import OeciLogin from '../OeciLogin';
./src/components/AppRouter/index.tsx:          <AuthenticatedRoute path="/oeci" component={OeciLogin} />

mkw@M4600:~/Documents/repos/re-pdx/recordexpungPDX/src/frontend$ gpnifr OECI
./src/components/PublicRoute/index.tsx:    // if authenticated it redirects to the OECI login
./src/components/OeciLogin/index.tsx:            aria-label="OECI Login Form"
./src/components/OeciLogin/index.tsx:                Log in to OECI to search and analyse criminal records for
```

Note that "OECI" is used only in comments and user-facing copy. 

We should follow the existing capitalization convention and use Oeci instead of OECI, acronyms aren't exceptions to the rule.

https://github.com/airbnb/javascript#naming-conventions

src/frontend/src/service/api-service.ts
```diff
@@ -11,6 +11,7 @@ export type Request = {
   method: Method;
   headers?: object;
   authenticated?: boolean;
+  withCredentials?: boolean;
```
maxkwallace: What does `withCredentials` do and why do we need it?

src/frontend/src/redux/system/actions.ts
```diff
@@ -30,10 +31,24 @@ export function logOut() {
...
+export function OECILogIn(username: string, password: string): any {
+  return (dispatch: Dispatch) => {
+    return apiService(dispatch, {
+      url: '/api/oeci_login',
+      data: { oeci_username: username, oeci_password: password },
+      method: 'post',
+      authenticated: true,
+      withCredentials: true
+    }).then((response: any) => {
+      history.push('/record-search');
```
maxkwallace: This can be a todo for later, but instead of manually changing the history, it would be better to have routing logic that reads from the Redux state (are we authed with OECI?) and determines what page to show that way.

wittejm:
@maxkwallace The "copy/replace" approach was something that @ForrestLonganecker and I worked out because I was really committed to not adding testing logic to the main backend code (i.e., checking to see if we're in dev or test and serve a spoofed response in dev), logic that we would want to take out for deployment to prod. We agreed that this was a shortcut of a solution and could be cleaner.

An idea we considered, and that I actually just got around to checking if it would work (and it does!), is to map the spoofing versions of the two endpoint files via a docker volume command in the docker-compose file. This way, we can optionally run a setup that contains the spoofed endpoint files instead of the regular versions, and the source code is unaffected. I can plan to get this working and PR'd on Saturday, probably.

maxkwallace:
@wittejm @ForrestLonganecker :+1: if you two already discussed and both prefer the `cp` solution, let's keep that.

src/frontend/src/components/OeciLogin/index.tsx
ForrestLonganecker: I will rename this!

I think I was just confusing myself thinking that since it was an acronym the `ECI` would need to be capitalized but `oECILogIn` looked pretty strange.

I think the proper naming convention for a function would just be standard camelCase `oeciLogIn`

src/frontend/src/service/api-service.ts
ForrestLonganecker: `withCredentials` is the way you are able to make a request that passes cookies in axios

https://stackoverflow.com/a/43178070

[Axios github docs:](https://github.com/axios/axios)
```
  // `withCredentials` indicates whether or not cross-site Access-Control requests
  // should be made using credentials
  withCredentials: false, // default
```

src/frontend/src/service/cookie-service.ts
```diff
@@ -21,13 +21,18 @@ export function decodeCookie() {
...
+export function setLogInCookie(inputToken: string, inputId: string) {
   // 'max-age' is set to 1 hour
   document.cookie = `authToken=${inputToken}; max-age=36000;`;
   document.cookie = `userId=${inputId}; max-age=36000;`;
 }
 
+export function setOECICookie(inputToken: string) {
+  document.cookie = `oeci_token=${inputToken}; max-age=900`;
+}
+
```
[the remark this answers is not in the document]
ForrestLonganecker: This was something that I was using while debugging cookie behavior, I need to remove it.

Thanks for catching that!

src/frontend/src/redux/system/actions.ts
ForrestLonganecker: I can throw a conditional statement that will only redirect if the client has the `oeci_token` in browser.

Previous to this implementation we were doing something similar, just using `react-router` `<Redirect to='/record-search' />` instead. 

https://github.com/codeforpdx/recordexpungPDX/blob/58947092cbe202835374c28529cc43a154f8edd6/src/frontend/src/components/OeciLogin/index.tsx#L65

src/frontend/developerUtils/developerUtils.md
```diff
@@ -0,0 +1,21 @@
...
+
+To use them replace the corresponding files in the `/src/backend/expungeservice/endpoints/...` folder.  
+
+from the root directory of the app you can run:  
+
+`cp src/frontend/developerUtils/search.py src/backend/expungeservice/endpoints/search.py`  
+
+`cp src/frontend/developerUtils/oeci_login.py src/backend/expungeservice/endpoints/oeci_login.py`  
+
+## Before committing make sure to revert these two files to their original forms.
```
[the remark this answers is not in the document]
ForrestLonganecker: There was talk about trying to keep mock code out of the committed backend so this was an effort to work around that and provide the front-end devs a way to more similarly mock out the interaction between client and backend.

I agree that it is not the best implementation.

ForrestLonganecker:
I started running into an error where our custom-request-middleware was logging out when you input improper oeci log-in information.

Because the error came back as a 401 it would automatically log out completely.

I added a check to see if the `error.response.config.url !== '/api/oeci_login'` to account for this edge case.

I also created two new `make` commands:
`dev_utils_up`
and
`dev_utils_down`

you can now run these to add the developerUtils files or remove them in a single command.

Oeci login button was for some reason requiring two button presses in order to register. I refactored the component logic in a way that we now only need to press it once to log in.

I believe this had to do with the `validateForm()` function being called separate from the actual call to logIn, the first click would set the error messages and while they are being set it would check their values. Because they had not finished setting the `if` statement would return false and it would not make the request. 

Second time you click the button it would have the correct information from those error variables and send the form.

I have now tapped into the Synchronous callback from the `setState` call so that it will wait until the variables are set before attempting logIn.

I also added a check that will only redirect from `/oeci` to '/record-search' if there is an `oeci_token` present in the browser. Although I think that this may be a little redundant as the only non-error response returned from the request to `/api/oeci_login` is that you would have the cookie in the browser. This will however keep them at the oeci login page unless they truly have one set in browser.

src/frontend/src/service/cookie-service.ts
```diff
@@ -30,4 +30,9 @@ export function setCookie(inputToken: string, inputId: string) {
 export function removeCookie() {
   document.cookie = 'authToken=; max-age=0;';
   document.cookie = 'userId=; max-age=0;';
+  document.cookie = 'oeci_token=; max-age=0;';
+}
+
+export function hasOeciToken() {
+  return decodeCookie().oeci_token;
```
maxkwallace: `hasOeciToken` sounds like it returns a boolean. So IMO it should return a boolean or be renamed to `oeciToken`

ForrestLonganecker: While I was refactoring I realized that in this form it would return either a oeci_token `string` or `undefined` which satisfies the `if` statement. 

I will switch it back to returning `true/false`

Issue #72, opened 2019-02-22
Update case parser to check for probation information


NickSchimek:
The case parser also needs to collect information regarding probation violations. Update case parser to check if probation was revoked.
Use the test fixture in: https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/tests/fixtures/case_details.py#L639

### This task is done when...
 - [ ] Tests exists for new functionality
 - [ ] Case has a method Revoked that it responds to
 - [ ] CaseParser collects probation information

PR #490, opened 2019-10-26
Update case parser to check if probation revoked

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/72

This PR is probably easier to review if you go commit by commit instead of looking at all changes at once. The check for "Probation Revoked" is done as a simple whole page scan using a fuzzy search. A number of relevant refactorings were done along the way. 

Thanks to @NickSchimek for answering a number of questions related to this task.


KentShikama:
If anyone is wondering about the performance difference as a result of the FuzzySearch, below are the results of the pytest runs before and after this PR. Note that given the addition of 20 tests, the 1 second increase in time seems reasonable. Further evaluation is needed to test the effect on long (think 100+ case) records.
```

Before

======================= 292 passed, 2 warnings in 14.44s =======================

After

======================= 312 passed, 2 warnings in 15.88s =======================
```

Issue #385, opened 2019-09-21
Only set eligibility date if type eligibility is true


NickSchimek:
## Description

Currently eligibility date is set regardless when a charge is currently not time eligible, however it came to my attention that it could cause some confusion for users of the app. This could lead them to think that the charge will be eligible at that given date even though it says the charge is not type eligible. 

Currently the `TimeAnalyzer` is using the method `set_time_ineligible` in `BaseCharge.py`. The method can definitely be refactored/improved so that the change can be easily implemented.

Please add appropriate test cases for expected behavior.

Please reach out to `@poppa_nick` on slack if any questions.

### This task is done when:
 - Eligibility date is only persisted when a charge is type eligible. We should check on `type_eligibility != False` instead of `type_eligibility == True`, because type_eligibility can also be None when it "needs further analysis".

wittejm:
There are still some record results that have this bug, listed on the expungement the bug reports page

wittejm:
Completed by a recent logic change in the backend (always compute time eligibility) and a change to the frontend (hide time eligibility for type-ineligible changes)

PR #503, opened 2019-10-31
Refactor charge.py to set classification at only one location

KentShikama:
I made these refactorings while working on #385. Review should be easier if followed commit by commit.

The below is self-quoted from commit messages.

> The existing design relies upon unnecessarily many mutable methods to set the classification of a Charge. The structure of the method `_set_classification` implies "non-linear" logic though the logic itself is "linear": it attempts match a classification type to the current charge and then moves onto the next attempt. Using yields ensures that once the classification type is found, the loop short circuits to return what was found as a string representing the corresponding class name. Since the classification will always be "UnclassifiedCharge" if nothing else matches, the `cls.classification = None` can be removed.
> 
> Extract charge classifier into its own class

### DIFF
--- a/src/backend/expungeservice/models/charge.py
+++ b/src/backend/expungeservice/models/charge.py
@@ -14,6 +14,7 @@ from expungeservice.models.charge_types.parking_ticket import ParkingTicket
 from expungeservice.models.charge_types.person_crime import PersonCrime
 from expungeservice.models.charge_types.schedule_1_p_c_s import Schedule1PCS
 from expungeservice.models.charge_types.unclassified_charge import UnclassifiedCharge
+from expungeservice.models.charge_classifier import ChargeClassifier
 
 
 class Charge:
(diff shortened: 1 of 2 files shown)


src/backend/expungeservice/models/charge_classifier.py
```diff
@@ -0,0 +1,118 @@
...
+    @staticmethod
+    def _set_classification_by_level(level):
+        yield ChargeClassifier._non_traffic_violation(level)
+        yield ChargeClassifier._misdemeanor(level)
+        yield ChargeClassifier._felony_class_c(level)
+        yield ChargeClassifier._felony_class_b(level)
+        yield ChargeClassifier._felony_class_a(level)
+
+    @classmethod
+    def _marijuana_ineligible(cls, statute, section):
```
NickSchimek: Perhaps we can change all these methods to staticmethods since the class var is no longer needed?

src/backend/expungeservice/models/charge.py
```diff
@@ -14,127 +14,23 @@
...
     @classmethod
     def create(cls, **kwargs):
-        cls.classification = None
         case = kwargs['case']
         statute = Charge.__strip_non_alphanumeric_chars(kwargs['statute'])
         level = kwargs['level']
         chapter = Charge._set_chapter(kwargs['statute'])
         section = Charge.__set_section(statute)
-        Charge._set_classification(case, statute, level, chapter, section)
+        cls.classification = ChargeClassifier(case, statute, level, chapter, section).classify()
```
NickSchimek: It seems this can also be changed to a staticmethod, and replace the class var with a local var?

KentShikama:
@NickSchimek PTAL (Please take another look)

PR #506, opened 2019-10-31
Automatically register all endpoints

### LINKED ISSUE #88: Automate registering of endpoint routes in expungeservice/app.py
Currently, when adding endpoints to the backend, routes for the endpoints (found in `src/backend/expungeservice/endpoints/`) must be manually added to the `create_app` method of `src/backend/expungeservice/app.py`, e.g.:

```
    # Register endpoint routes here:
    hello.register(app)
    auth.register(app)
```

The goal would be to alter `src/backend/expungeservice/app.py` to search the endpoint directory and register the endpoint routes itself, so that regardless of removals or additions to the endpoints, the backend will find them and register them.

### This task is done when...
 - Routes for new endpoints for the backend don't need to be manually added to `src/backend/expungeservice/app.py`

KentShikama:
Closes #88

All endpoints have the method signature register(app) and thus we can use that assumption to reflectively call those methods.

### DIFF
--- a/src/backend/expungeservice/app.py
+++ b/src/backend/expungeservice/app.py
@@ -1,9 +1,11 @@
+import pkgutil
+from importlib import import_module
+
 from flask import Flask
 
 from .config import app_config
 
-# Add new endpoint imports here:
-from .endpoints import hello, auth, users, oeci_login, search
+from . import endpoints
 from .request import before, after, teardown
 import logging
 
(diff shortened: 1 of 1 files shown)


NickSchimek:
This looks great! @erikjasso @wittejm @ash-tamraz do you want to take a look?

erikjasso:
Looks good to me.

erikjasso:
@KentShikama Could you add a test for this feature? Something simple to verify it -- there must be a way to check if a specific module has been imported?

KentShikama:
@erikjasso If you comment out (line 42) register_fn(app), the following tests will fail as the endpoints do not get registered.

```
tests/endpoints/test_auth.py F.FFFFFF                        
tests/endpoints/test_oeci_login.py FF                        
tests/endpoints/test_search.py F                             
tests/endpoints/test_users.py FF.FFFFF.FFFFFFFF.F
```

In essence, I would say it is rather well tested already. It looks like there is no test for the `hello` endpoint but would like to see one for that?

PR #511, opened 2019-11-04
Remove unnecessary Promise wrapping from apiService

KentShikama:
We currently are doing an unnecessary double Promise wrapping with apiService. In the existing middleware, we return the result of the dispatch functions (at line 49 and 59 of request.ts) without the result of the axios request. By returning the result of the axios request directly, we no longer need to pass in the resolve/reject capabilities of the wrapping Promise.

Front-end unit test results

```
 PASS  src/service/api-service.test.ts
  API SERVICE TEST
    ✓ returns data with get (5ms)
    ✓ returns with error on bad base url (1ms)
    ✓ returns with 404 error on bad route
    ✓ returns search in JSON (1ms)

Test Suites: 1 passed, 1 total
Tests:       4 passed, 4 total
Snapshots:   0 total
Time:        2.01s, estimated 3s
Ran all test suites.
```

### DIFF
--- a/src/frontend/src/redux/middleware/request.ts
+++ b/src/frontend/src/redux/middleware/request.ts
@@ -45,7 +45,7 @@ const RequestMiddleware = (store: MiddlewareAPI) => (next: Dispatch) => (
   ) {
     let promise = makeUnauthenticatedRequest(action);
     let result = next(action);
-    return Object.assign({ promise: promise }, result);
+    return Object.assign({ promise }, result);
   }
 
   // This error shouldn't occur because app code has access to the store and thus knows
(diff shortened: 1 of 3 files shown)


src/frontend/src/redux/middleware/request.ts
```diff
@@ -45,8 +43,9 @@ const RequestMiddleware: any = (store: Store) => (next: Function) => (
   if (
     !(action.request.authenticated && action.request.authenticated === true)
   ) {
-    makeUnauthenticatedRequest(action);
-    return next(withoutUnserializables(action));
+    let promise = makeUnauthenticatedRequest(action);
+    let result = next(action);
+    return Object.assign({ promise: promise }, result);
```
maxkwallace: Can be:

return Object.assign({ promise }, result);

src/frontend/src/redux/middleware/request.ts
```diff
@@ -55,23 +54,22 @@ const RequestMiddleware: any = (store: Store) => (next: Function) => (
   // a failure with the request.
   throwUnlessCurrentlyAuthenticated(store);
 
-  makeAuthenticatedRequest(store, action);
-  return next(withoutUnserializables(action));
+  let promise = makeAuthenticatedRequest(store, action);
+  let result = next(action);
+  return Object.assign({ promise: promise }, result);
```
maxkwallace: Can be:

    return Object.assign({ promise }, result);

PR #521, opened 2019-11-07
Replace existing session management with flask-login

KentShikama:
Session management is now handled with HTTP cookies. By using the default login_user mechanism with remember=True, flask-login stores two HTTP cookies: "session" and "remember_token". As the name suggest, the "session" cookie lasts the session and has otherwise no expiration date. The "session" cookie is http only, meaning that no client side scripts can read it. When the user ends a session, the "session" cookie is restored from the "remember_token" assuming they did not hit the backend logout endpoint (not implemented yet) or the cookie isn't expired. When a "session" cookie is restored from the "remember_token", the session will be marked as "not fresh". In a future commit, some endpoints (especially related to account information changes) should be changed to only accept fresh sessions. On the front-end we check the existence of the "remember_token" to see whether the user is logged in, and thus skip the login front page. Note that even if an attacker fakes a "remember_token" on the frontend, they will still be blocked off from authentication required endpoints on the backend. Upon clicking the logout button on the frontend, the "remember_token" and "oeci_token" are removed. In the future, it should also hit the backend logout endpoint to clear the "session" cookie and invalidate the associated "remember_token".

Besides the replacement of the auth_token with the two HTTP cookies and any implicit features that flask-login brings, this commit should have no functional changes. All backend tests have been preserved in an "one-to-one" fashion. An exception is that I have added one test that checks the user_protected endpoint is not accessible without login.

Edit: As motivated by https://github.com/codeforpdx/recordexpungPDX/pull/521#discussion_r343949671, I've added a new endpoint `/api/user` to handle requests for the current user without explicitly passing in the `user_id`.


src/backend/expungeservice/config.py
```diff
@@ -18,6 +19,7 @@ class Production(object):
     DEBUG = False
     TESTING = False
     SQLALCHEMY_DATABASE_URI = os.getenv('DATABASE_URL')
+    SECRET_KEY = os.getenv('SECRET_KEY')
```
wittejm: At this point the JWT_SECRET_KEY is only used for symmetric encryption of oeci credentials. Could we use SECRET_KEY for that instead, and only have one key in the app config again?  Or would overloading the purpose of this secret key be considered bad practice?

src/backend/expungeservice/endpoints/auth.py
```diff
@@ -94,23 +19,34 @@ def post(self):
...
 
-        response_data = {
-            'auth_token': get_auth_token(current_app,
-                                         user_db_result['user_id']),
-            'user_id': user_db_result['user_id']
-        }
-        return jsonify(response_data)
+        user = from_dict(data_class=User, data=user_db_result)
+        User.login_user(user)
+        return jsonify({})
```
wittejm: The frontend still needs to obtain and store the logged in user’s id to read or update the user’s own profile info. Shouldn’t that still be here?

maxkwallace:
@KentShikama Thanks for the PR! In general, I'm very in favor of adopting existing libraries and conventions (like Flask-Login for authentication) for this project, and that's a big area where the existing code can be improved.

I haven't had a chance to read all the code in the PR yet, but I figure it'd still be better to comment now rather than waiting. I should also say that I've read https://flask-login.readthedocs.io/en/latest/ to the extent possible given time constraints. There are some things in the docs and your comment above that are unclear to me.

- Where does the session cookie come from and how it is managed? (When is it created and deleted, if ever?)

- Regarding:

>  When the user ends a session, the "session" cookie is restored from the "remember_token" assuming they did not hit the backend logout endpoint (not implemented yet) or the cookie isn't expired.

What does "ends a session" mean here? I would expect the session cookie to be restored from the remember token if the user reopens the page after previously being logged in. When you say "cookie isn't expired" is that referring to the "remember_token"?

- How does this PR handle the case where the user becomes logged out because the session expired? In this case, how would the React app know that the user became logged out? 

- Does Flask-Login require us to store state in the database indicating whether the user is logged in or not, or can that info be stored only in one of the cookies? (This would be "stateless" auth where the state info would have to be encrypted with an HMAC.)

src/backend/Pipfile
```diff
@@ -11,13 +11,15 @@ flask-bcrypt = "*"
...
-expungeservice = {editable = true, path = "."}
+expungeservice = {editable = true,path = "."}
 pytest = "*"
 requests-mock = "*"
 requests = "*"
 python-dateutil = "*"
 uwsgi = "*"
 cryptography = "*"
+flask-login = "==0.4.1"
+dacite = "==1.0.2"
```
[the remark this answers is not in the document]
KentShikama: Locking package versions ensures we won't be affected by anything sudden breaking changes.

src/backend/expungeservice/config.py
KentShikama: I just didn't bother thinking through the security implications of reusing the same secret key so I was being conservative. My guess is that it is fine.

src/backend/expungeservice/endpoints/auth.py
KentShikama: Ah yes I have yet to remove the user ids from the `/api/users` backend endpoint. I'll follow up with some commits on this PR soon.

KentShikama:
@maxkwallace 

> Where does the session cookie come from and how it is managed? (When is it created and deleted, if ever?)

The flask-login library creates it and sends attaches it to the response upon login. Almost all of its details are managed by the flask-login library. The session cookie is deleted when the browser session ends (the user quits their browser).

> What does "ends a session" mean here? I would expect the session cookie to be restored from the remember token if the user reopens the page after previously being logged in. When you say "cookie isn't expired" is that referring to the "remember_token"?

I've mentioned this above but specifically the user quitting their browser. Indeed the session cookie is restored in a non-fresh state when the user reopens the page after previously being logged in. And it is indeed referring to the "remember_token".

> How does this PR handle the case where the user becomes logged out because the session expired? In this case, how would the React app know that the user became logged out?

Since the flag remember=True is set to on by default, the user will never be logged out unless they either delete their "remember_token", their "remember_token" expires, or the user explicitly clicks logout (and thus the webapp hits the logout backend endpoint). The React app assumes that if the browser holds a "remember_token", the user is logged in. All three cases listed before are covered as follows. When the "remember_token" is manually deleted, this trivially holds. All modern browser implementations delete cookies once they are expired. The React app explicitly removes all cookies (see [remove cookie](https://github.com/codeforpdx/recordexpungPDX/pull/521/files#diff-498303f9496c1d6ce33ab4ff30b470a3R24)) when the user hits the logout button.

> Does Flask-Login require us to store state in the database indicating whether the user is logged in or not, or can that info be stored only in one of the cookies? (This would be "stateless" auth where the state info would have to be encrypted with an HMAC.)

It doesn't require us to store state in the database but it does in memory, and flask-login handles this for us. If memory is cleared (e.g., the app backend is restarted), then all users are logged out.

src/backend/expungeservice/endpoints/auth.py
KentShikama: I've added a new endpoint `/api/user` (`GET`, `POST`) to handle the respective requests for the current user. Please see the new commit and commit message for more information.

src/backend/Pipfile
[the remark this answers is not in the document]
wittejm: So far we've been managing that risk with reviewing dependabot's version bump PRs as they come in. In theory. An obvious drawback is the backlog of PRs that have built up, but it makes sense to me to have a consistent pattern for using the locked vs. the latest versions for all of our dependencies. 

I'm not offering a suggestion on which approach would be preferred here though :/

KentShikama:
@maxkwallace @wittejm Just note that I've edited my comment above. The major change is that I realized that the default flask session is completely stateless; for whatever reason I had the impression that it wasn't.

src/frontend/src/service/api-service.ts
```diff
@@ -15,14 +15,17 @@ export type Request = {
...
+
 export default function apiService(
   dispatch: Function,
   request: Request
 ): AxiosPromise {
-  // The Action dispatched here is handled in the Request Middleware, so you can call
-  // apiService(...).then(...).catch(...) like a normal Axios request.
-  let result = dispatch({ type: REQUEST, request } as RequestAction);
-  return result.promise;
+  dispatch({ type: REQUEST, request });
```
maxkwallace: I think we should get rid of this dispatch (and the REQUEST constant above) because AFAIK as of this PR this action does nothing,

src/frontend/src/components/Authenticator/index.tsx
```diff
@@ -23,8 +23,8 @@ class Authenticator extends React.Component<Props> {
     // Upon page refresh or navigation via manual url input this will
     // repopulate store with information stored in the browser cookie
```
maxkwallace: This comment should be removed or updated.

KentShikama:
@maxkwallace The `user_id` information is held within the `session` cookie. With the new endpoints provided in commit https://github.com/codeforpdx/recordexpungPDX/pull/521/commits/ead3b80aebfdeb78d720e1a6ff8aa2d8e3b84369, the `user_id` is assumed to be the currently logged in user. If the user really wants to know their own user_id, they can query the `/api/user` endpoint while logged in.

PR #526, opened 2019-11-08
Simplified mocked oeci endpoints using a Docker volume mapping 

wittejm:
This updates the oeci integration testing utility introduced in #484.

The mocked endpoint setup uses the same mocker code files but avoids manually replacing source files by using a docker volume mapping, so that the mock files override the regular files in only the docker container.

These mocked files will need to updated independently if applicable changes are made to the regular endpoint files e.g. in the open PR #521


KentShikama:
@wittejm A bit late but just verified that this works.

PR #529, opened 2019-11-09
Add logout endpoint and sync it with logout button on the frontend

### LINKED ISSUE #504: Create logout endpoint
## Description

The login endpoint creates a cookie, hence a logout endpoint is needed to destroy/invalidate the cookie.

KentShikama:
Closes #504 

~~Relative to #521~~

Quoted from individual commits:

> Add backend endpoint for logout
> 
> Sync backend logout endpoint with frontend logout button
> Note that when the frontend makes a request to a protected endpoint and fails because the backend says they aren't logged in, the frontend logs out the user without hitting the backend logout endpoint again.
>
>  Ensure sensitive endpoints require fresh login
> Sensitive operations like changing passwords or looking up your own user information should require a fresh login.

### DIFF
--- a/src/frontend/src/redux/system/actions.ts
+++ b/src/frontend/src/redux/system/actions.ts
@@ -19,10 +19,15 @@ export function logIn(email: string, password: string): any {
   };
 }
 
-export function logOut() {
-  removeCookie();
-  return {
-    type: LOG_OUT
+export function logOut(): any {
+  return (dispatch: Dispatch) => {
+    return apiService(dispatch, {
+      url: '/api/logout',
+      method: 'post'
+    }).then((response: any) => {
+      removeCookie();
+      dispatch({ type: LOG_OUT });
+    });
   };
 }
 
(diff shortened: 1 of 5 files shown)


src/frontend/src/redux/system/actions.ts
```diff
@@ -19,10 +19,15 @@ export function logIn(email: string, password: string): any {
...
-    type: LOG_OUT
+export function logOut(): any {
+  return (dispatch: Dispatch) => {
+    return apiService(dispatch, {
+      url: '/api/logout',
+      method: 'post'
+    }).then((response: any) => {
+      removeCookie();
+      dispatch({ type: LOG_OUT });
+    });
```
maxkwallace: Need to add a catch block, even if it just throws an alert to let the user know the request failed.

KentShikama: Thanks for the catch; I've added an issue to improve from the default `alert`: https://github.com/codeforpdx/recordexpungPDX/issues/538.

PR #532, opened 2019-11-11
Addresses issue #387 - eligible mrc and all other charges are arrests

### LINKED ISSUE #387: Apply new time eligibility ru
## Description
If the most recent conviction is eligible and if there are only other "arrests" on that case then the whole case is time eligible.

The term 'arrests' means the charge was acquitted.

Should wait until #382 is done before working on this.

Please add appropriate test cases for expected behavior.

Please reach out to `@poppa_nick` on slack if any questions.

### This task is done when:
 - arrests in the case belonging to the MRC are marked as time eligible.

aemrich:
Fixes #387. 

Created test case (test_mrc_with_arrests) and added _set_arrests_as_eligible method.

_set_arrests_as_eligible checks if most recent conviction is time eligible and if all other charges are arrests. If both cases are true, then marks all arrests as time eligible.


NickSchimek:
Thanks @aemrich ! I will look over this in the next couple days

KentShikama:
@NickSchimek What should be the outcome if there are charges that aren't just arrests that are from a different case than the one the most recent conviction is from?

NickSchimek:
> @NickSchimek What should be the outcome if there are charges that aren't just arrests that are from a different case than the one the most recent conviction is from?

@KentShikama Any charge that is not in the same case file as the MRC should be left unchanged, that includes arrests and convictions.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -71,6 +72,22 @@ def _mark_as_time_ineligible(charges, reason, eligibility_date):
         for charge in charges:
             charge.set_time_ineligible(reason, eligibility_date)
 
+    @staticmethod
+    def _set_arrests_as_eligible(expunger):
+        # If most recent conviction is time eligible
+        # AND all other charges are arrests
+        # Then make all other charges time eligible
+        if expunger.most_recent_conviction.expungement_result.time_eligibility.status:
+            arrest_dispositions = {'Dismissed', 'Acquitted', 'Dismissal', 'No Complaint'}
```
NickSchimek: A charge already knows if it is an arrest or a conviction via its public interface. By hard-coding these values here, we are duplicating code, and it has the potential to create hard to find bugs. For example: If the criteria for arrest changes in the future, we will have to change it in the charge class and remember to change it here.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -71,6 +72,22 @@ def _mark_as_time_ineligible(charges, reason, eligibility_date):
...
+            mrc = expunger.most_recent_conviction
+            non_mrc_charges = [x for x in expunger.charges if x != mrc]
```
NickSchimek: We're only concerned with charges in the same case as the mrc; all other charges are unaffected.

NickSchimek:
@aemrich Thanks for the PR! Please let me know if there are any clarifications that I need to make in the issue. I'd like to see some more test cases added: Specifically what happens to the charges when the MRC is not time-eligible, and what happens to charges that are not in the same case as the MRC. What happens when a case contains two charges and a dismissal?

Can you think of anymore test cases?

NickSchimek:
@aemrich 
I was reading through the [Oregon statute for setting aside convictions](https://www.oregonlaws.org/ors/137.225) and they state that a single violation (non-motor vehicle) does not count as a conviction 
>A single violation, other than a motor vehicle violation, within the last 10 years is not a conviction under this subsection.

And I confirmed with Michael that if the MRC case also has a single violation, then they are both time-eligible. Hence, we not only need to make arrests time eligible but also single non-motor vehicle violations if they're in same case file as the MRC.

Unfortunately, the answer to this question spawned more questions. I am going to put a list together of possible scenarios and send them to Michael for verification. Once I've got this done, I will update the issue.

Perhaps at the next meet up we can discuss this PR in more detail and explore possible implementations.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
aemrich: @NickSchimek ah ok, I'll make a change here.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
aemrich: @NickSchimek Cool, I wasn't sure if the expunger processed more than one case at a time - wanted to follow up on that.

aemrich:
@NickSchimek Ok, I'll think of more test cases. It'd be great to see some more of those scenarios before the meetup, if you get them back from Michael. Either way, let's talk more at the meetup.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
NickSchimek: ask questions anytime. A charges time-eligibility is determined by all the charges in the record. Hence, the list of charges is every charge in the record.

aemrich: Ah, makes perfect sense, thank you

NickSchimek:
@aemrich I documented the slack discussion we had with Michael in the issue for reference.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: @aemrich I'm going to start working on issue #601 , which will subsume this check for different ruling strings, and express convicted/dismissed as a boolean or more standardized string.

aemrich: @wittejm ok, sounds good. I'll keep that in mind (I'm going to use convicted() method from the Charge class for now, but will change as you work on #601 )

wittejm:
Superseded by #716

PR #533, opened 2019-11-11
Integrate search page with real search backend endpoint

KentShikama:
~~Relative to #529~~

This has been tested against the real OECI database on a local instance.

### DIFF
--- a/src/frontend/src/redux/records/actions.ts
+++ b/src/frontend/src/redux/records/actions.ts
@@ -1,6 +1,6 @@
 import { Dispatch } from 'redux';
 import apiService from '../../service/api-service';
-import { AxiosResponse } from 'axios';
+import { AxiosError, AxiosResponse } from 'axios';
 import { LOAD_SEARCH_RECORDS, LOAD_SEARCH_RECORDS_LOADING } from './types';
 
 export function loadSearchRecords(
@@ -22,11 +22,15 @@ export function loadSearchRecords(
       },
       method: 'post',
       withCredentials: true
-    }).then((response: AxiosResponse) => {
-      dispatch({
-        type: LOAD_SEARCH_RECORDS,
-        search_records: response.data.data.record
+    })
+      .then((response: AxiosResponse) => {
+        dispatch({
+          type: LOAD_SEARCH_RECORDS,
+          search_records: response.data.data.record
+        });
+      })
+      .catch((error: AxiosError) => {
+        alert(error.message);
       });
-    });
   };
 }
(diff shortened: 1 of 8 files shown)


src/frontend/src/redux/records/actions.ts
```diff
@@ -0,0 +1,33 @@
...
+    return apiService(dispatch, {
+      url: '/api/search',
+      data: {
+        first_name: first_name,
+        middle_name: '',
+        last_name: last_name,
+        birth_date: birthday
+      },
+      method: 'post',
+      authenticated: true,
```
maxkwallace: As of the other PR that removes the Request Middleware, `authenticated: true` no longer does anything so we should remove it.

I didn't think of this when reviewing the other PR, but yeah, we should remove 

```
  authenticated?: boolean;
```

from the Request type def.

src/frontend/src/redux/records/actions.ts
```diff
@@ -0,0 +1,33 @@
...
+      withCredentials: true
+    }).then((response: AxiosResponse) => {
+      dispatch({
```
maxkwallace: Should probably have some kind of validation for the response format here,

src/frontend/src/redux/records/actions.ts
```diff
@@ -0,0 +1,33 @@
...
+        type: LOAD_SEARCH_RECORDS,
+        search_records: response.data.data.record
+      });
```
maxkwallace: Should have a catch block, even if it just throws an alert onto the page to let the user know that the search failed.

src/frontend/src/redux/records/actions.ts
```diff
@@ -0,0 +1,33 @@
+import { Dispatch } from 'redux';
+import apiService from '../../service/api-service';
+import { AxiosResponse } from 'axios';
+import { LOAD_SEARCH_RECORDS, LOAD_SEARCH_RECORDS_LOADING } from './types';
+
+export function loadSearchRecords(
+  first_name: string,
+  last_name: string,
```
maxkwallace: These names should be camelCased

src/frontend/src/redux/records/actions.ts
KentShikama: Were you thinking of checking the response against a JSON schema?

KentShikama:
@maxkwallace PTAL (Please take another look)

src/frontend/src/redux/records/actions.ts
maxkwallace: @KentShikama nah, it can be something simple, just that from my experience it will save someone time in the future if we have some basic checks to make sure there won't be a property access error on `response.data.data.record` in case the backend doesn't send the correct format, or someone changes something in the future. If something goes wrong then the app ideally shouldn't silently fail or fail in an unexpected way. Just thinking 80/20 rule, doesn't have to be checked against a schema (though that's fine too).

KentShikama: Ok I've just checked that `data` and `record` properties exist under `response.data` for now.

PR #543, opened 2019-11-13
Clean up following search integration 

KentShikama:
Quoted from commits

>  Update api-service tests to use AxiosRequestConfig
> This should have been a part of efe9e80

>  Remove unused records/mock-actions.ts
> Note if the search frontend page needs to be tested without connecting to the backend, the `import axios from './__mocks__/axios';` can be used in service/api-service.ts instead of the real axios request.

> Refactor `__mocks__/axios.ts` request
> 
> - Use a normal function instead of jest.fn, given how we don't use any functionality of jest.fn.
> - Extract out fakeRecord into a const
> - Allow requestMock to take a generic type parameter so it typechecks

```
 PASS  src/service/api-service.test.ts
  API SERVICE TEST
    ✓ returns data with get (7ms)
    ✓ returns with error on bad base url (1ms)
    ✓ returns with 404 error on bad route
    ✓ returns search in JSON (2ms)
```


KentShikama:
Superseded by #572

PR #544, opened 2019-11-13
Various changes needed for Heroku production

KentShikama:
Review will probably be easier if done commit by commit.

Notable commits (quoted from commit messages)

> Fix 404 responses when navigating directly to non-root URLs

> Ensure Fernet receives 32 bytes even if SECRET_KEY is longer

Removes a number of unused files and various minor refactorings.


deployment/README_deploy.md
```diff
@@ -17,6 +17,7 @@ In the terminal, use `heroku` commands to build docker images locally and then p
 First, use the project login credentials to access heroku admin:
```
wittejm: Could you update this line to:

"First, use the project login credentials (username: recordexpungpdx) to access the heroku and docker admin accounts, respectively:"

PR #550, opened 2019-11-13
Remove obsolete targets from Makefile

wittejm:
The following make targets are no longer needed:

- `install`, `$requirements.txt` : we no longer use pipenv natively
- `dev_start`, `dev_stop`, and `dev_deploy`: we no longer use `dev stack` commands for running docker
- `test`: we no longer use pytest natively, and `dev_test` is the correct usage. 

Also added `&` to the `make clean` because failure on line 2 (if no files match) causes line 3 to not run, potentially leaving behind some artifacts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,8 @@
-.PHONY: install run clean
-
-install:
-	pipenv install
+.PHONY: clean
 
 clean:
-	rm -rf src/backend/*.egg-info
-	find . -type f -name \*~ | xargs rm
+	rm -rf src/backend/*.egg-info &
+	find . -type f -name \*~ | xargs rm &
 	find . -type f -name \*pyc | xargs rm
 
 IMAGES := database_image expungeservice_image webserver_image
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,9 +57,6 @@ applogs:
 weblogs:
 	docker logs --details -ft $$(docker ps -qf name=$(FRONTEND_CONTAINER_NAME))
```
erikjasso: I notice we're using two different methods for getting container logs (docker-compose logs and docker logs). Assuming they produce the same output, can we pick one?

Makefile
```diff
@@ -1,11 +1,8 @@
...
+.PHONY: clean
 
 clean:
-	rm -rf src/backend/*.egg-info
-	find . -type f -name \*~ | xargs rm
+	rm -rf src/backend/*.egg-info &
+	find . -type f -name \*~ | xargs rm &
 	find . -type f -name \*pyc | xargs rm
 
 IMAGES := database_image expungeservice_image webserver_image
```
erikjasso: I don't see that this variable is referenced anywhere. Is it still needed?

erikjasso:
We're not taking advantage of prerequisites (for example,  the up target could invoke the build target). Is there a reason for not doing that? Does docker-compose take care of it? Ah...nevermind...I see the docs address this question. Hooray for comprehensive documentation!

erikjasso:
@wittejm I'm done w/my review.

Makefile
wittejm: You're right, I can remove it.

PR #552, opened 2019-11-14
Refactor charge classes

NickSchimek:
It's best to review the commit message for the reasoning behind each refactoring. 

- Removed init method in charge subclasses
- Made default_type_eligibility method protected
- Replace default value with NotImplementedError
- Fixed spacing to comply with PEP8

### DIFF
--- a/src/backend/expungeservice/models/charge_types/charge.py
+++ b/src/backend/expungeservice/models/charge_types/charge.py
@@ -3,8 +3,7 @@ import weakref
 from datetime import datetime
 from datetime import date as date_class
 from dateutil.relativedelta import relativedelta
-from expungeservice.models.expungement_result import ExpungementResult, \
-    TimeEligibility, EligibilityStatus, TypeEligibility
+from expungeservice.models.expungement_result import ExpungementResult, TimeEligibility, EligibilityStatus
 
 
 class Charge:
@@ -22,7 +21,7 @@ class Charge:
         self._case = weakref.ref(case)
 
     def _default_type_eligibility(self):
-        return TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Examine')
+        raise NotImplementedError
 
     def case(self):
         return self._case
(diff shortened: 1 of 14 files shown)


src/backend/expungeservice/models/charge_types/charge.py
```diff
@@ -15,14 +14,14 @@ def __init__(self, case, name, statute, level, date, chapter, section, dispositi
...
-        type_eligibility = self.default_type_eligibility()
+        type_eligibility = self._default_type_eligibility()
         self.expungement_result = ExpungementResult(type_eligibility=type_eligibility, time_eligibility=None)
         self._chapter = chapter
         self._section = section
         self._case = weakref.ref(case)
 
-    def default_type_eligibility(self):
-        return TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Examine')
+    def _default_type_eligibility(self):
```
wittejm: Since this no longer has a default return value, could it be decorated with `@abstractmethod` for a tad of added clarity? Up to you though, since that'd definitely subjective.

NickSchimek: Thanks for the suggestion. I looked into adding the decorator and I don't feel its necessary here as the code is more clear now than what it was before. I trust developers will act accordingly when they see methods like this. Plus any attempt to initialize this class or a subclass without overriding the method will fail immediately. <- if that wasn't the case then I'd probably take you up on your suggestion.

I looked at it from this perspective. When someone see's this latch on the door then it conveys that it's meant to be pulled and that the door will swing open toward the user.
Screenshot 2019-11-16 00 37 42

However, does the door really need a sign on it when the latch strongly suggests to pull at it?
Screenshot 2019-11-16 00 33 35

wittejm: In rebuttal I shall cite the Zen of Python, line 2 :). (from `import this` , but everyone already knows that right?)

IMO `NotImplementedError` reads more like an implicit abstract method but could also be misinterpreted as a `TODO` line.

PR #556, opened 2019-11-15
Integrate mypy (an optional static type checker for Python) into backend

KentShikama:
This PR should barely affect anyone who isn't using the [new types feature in Python](https://docs.python.org/3/library/typing.html). Most commonly, it may occasionally require you to specify the type of the variable when using an empty list such as `list_charges: List[Charge] = []` but I would argue that this is a helpful use of types for a reader. Worst case just annotate your failing line with `# type: ignore` and `mypy` will move on. So far in this project's backend, I've been the only one to use types, albeit sparingly for now. As the project becomes larger, we should start adding types to our codebase.

The type checker doesn't apply to the tests/ or expungeservice/serializer/ folders for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,7 @@ weblogs:
 	docker logs --details -ft $$(docker ps -qf name=$(FRONTEND_CONTAINER_NAME))
 
 dev_test:
+	docker exec -t $$(docker ps -qf name=$(BACKEND_CONTAINER_NAME)) mypy
 	docker exec -t $$(docker ps -qf name=$(BACKEND_CONTAINER_NAME)) pytest
 
 dev_drop_database:
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The app is deployed on the webapp hosting service, [Heroku](https://www.heroku.c
 
 Our latest dev version (this repo's master branch) is publicly viewable! Here: https://recordexpungpdx.herokuapp.com/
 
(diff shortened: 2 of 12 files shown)


wittejm:
To offer a contrasting opinion, I'm a fan of static typing because I think it makes code easier to read, maintain / extend, and understand. I have no objection to adding it here. 

I don't have a strong opinion on the Parser code since I've personally never touched it. I’m interested in talking more about this at the next meeting!

KentShikama:
Dynamic typing can be clearer when the data structure is "close" to the variable such as `foo = "Hello"`, which clearer than `foo: String = "Hello"`. However for things that have "farther" assignments such as some of our model classes, it helps to have types on the models. For example, for the `date` field in the `Case` class it isn't immediately obvious without types whether a date object will be passed in or a str.

```
@dataclass
class Case:
    name: str
    birth_year: Optional[int]
    case_number: str
    citation_number: str
    location: str
    date: date
    violation_type: str
    current_status: str
    ...
```

You have to do a reference jump followed by a scroll up in the IDE to realize that we convert the parsed date in string format to a date before creating the Case object. Now imagine if that was three or four jumps. Typing can be optionally applied for these kinds of cases.

Types eliminate certain classes of errors and I agree they are no substitute for good tests. 

The way I have integrated it into travis is by including it with `make dev_test`: see 
https://github.com/codeforpdx/recordexpungPDX/pull/556/files#diff-b67911656ef5d18c4ae36cb6741b7965R82-R84. In essence, points 4 and 5 have already been taken care of. I will address the rest soon.

KentShikama:
@NickSchimek PTAL

PR #557, opened 2019-11-15
Update JSON response structure of search endpoint

KentShikama:
~~Relative to #543~~

The expungement result object will now more closely reflect the backend structure. The `type_eligibility` and `time_eligibility` are now their own separate objects. The status fields are no longer just booleans or null instead they are "Eligible", "Ineligible", or "Needs more analysis".

Examples

```
{
  "expungement_result": {
    "type_eligibility": {
      "status": "Eligible",
      "reason": "Eligible under 137.225(1)(b)"
    },
    "time_eligibility": {
      "status": "Ineligible",
      "reason": "Recommend sequential expungement",
      "date_will_be_eligible": "Sun, 15 Nov 2020 00:00:00 GMT"
    }
  }
}
```

```
{
  "expungement_result": {
    "type_eligibility": {
      "status": "Ineligible",
      "reason": "Ineligible under 137.225(5)"
    },
    "time_eligibility": null
  }
}
```

### DIFF
--- a/src/backend/expungeservice/models/charge.py
+++ b/src/backend/expungeservice/models/charge.py
@@ -57,9 +57,9 @@ class Charge:
             date_will_be_eligible = date_of_eligibility
         else:
             date_will_be_eligible = None
-        time_eligibility = TimeEligibility(status = EligibilityStatus.INELIGIBLE, reason = reason, date_will_be_eligible = date_will_be_eligible)
+        time_eligibility = TimeEligibility(status=EligibilityStatus.INELIGIBLE, reason=reason, date_will_be_eligible=date_will_be_eligible)
         self.expungement_result.time_eligibility = time_eligibility
 
     def set_time_eligible(self, reason=''):
-        time_eligibility = TimeEligibility(status = EligibilityStatus.ELIGIBLE, reason = reason, date_will_be_eligible = None)
+        time_eligibility = TimeEligibility(status=EligibilityStatus.ELIGIBLE, reason=reason, date_will_be_eligible=None)
         self.expungement_result.time_eligibility = time_eligibility
--- a/src/backend/expungeservice/models/charge.py
+++ b/src/backend/expungeservice/models/charge.py
@@ -57,9 +57,9 @@ class Charge:
             date_will_be_eligible = date_of_eligibility
         else:
             date_will_be_eligible = None
-        time_eligibility = TimeEligibility(status=False, reason=reason, date_will_be_eligible=date_will_be_eligible)
+        time_eligibility = TimeEligibility(status = EligibilityStatus.INELIGIBLE, reason = reason, date_will_be_eligible = date_will_be_eligible)
         self.expungement_result.time_eligibility = time_eligibility
 
     def set_time_eligible(self, reason=''):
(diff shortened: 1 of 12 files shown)


SteelFace87:
I currently have a PR Charges #566 which relies on the current JSON structure of the search endpoint.   This update would break the code if we chose to merge it.  I do not believe that it would be a lot of work to update the search results to take the new JSON format, however, my time as of late has been very limited.  If this is something @KentShikama wants to take on I am fine with it :)

KentShikama:
I would suggest to go ahead and merge #566 once @maxkwallace gives a thumbs up. I'll fix up the merge conflicts that follow.

NickSchimek:
> I would suggest to go ahead and merge #566 once @maxkwallace gives an approval. I'll fix up the merge conflicts that follow.

@KentShikama Now that the PR has been merged shall we update this PR to include the necessary changes to the front end?

KentShikama:
@NickSchimek @maxkwallace PTAL

src/frontend/src/components/Eligibility/index.tsx
```diff
@@ -38,16 +37,28 @@ export default class Eligibility extends React.Component<Props> {
...
-      } else if (type_eligibility === false) {
-        return ineligible;
-      } else {
-        return 'Unknown type or time eligibility';
+      switch (type_eligibility.status) {
+        case 'Eligible':
+          if (time_eligibility && time_eligibility.status === 'Eligible') {
+            return eligibleNow;
+          } else if (
+            time_eligibility &&
```
NickSchimek: Is this correct?

NickSchimek: oh nvm. I see what's happening.

NickSchimek:
I'm OK with the backend changes, however I'm not that familiar with react. Perhaps @maxkwallace  would like to take a look at the new changes?

The logic in the frontend does seem a bit hard to read. It is understandable, but seems a tad bit hard to read. Which brings me to this question. Do you think it would be good to move some of this logic to the backend? <- This would mean updating the JSON response. Or perhaps I can suggest ways to make it easier to parse/read?.

KentShikama:
@NickSchimek PTAL. I've done a bit of refactoring on the front-end. We can consider moving some of the logic to the backend but I think this PR is already large enough of a change.

src/backend/expungeservice/models/charge.py
```diff
@@ -57,9 +57,9 @@ def set_time_ineligible(self, reason, date_of_eligibility):
             date_will_be_eligible = date_of_eligibility
         else:
             date_will_be_eligible = None
-        time_eligibility = TimeEligibility(status=False, reason=reason, date_will_be_eligible=date_will_be_eligible)
+        time_eligibility = TimeEligibility(status = EligibilityStatus.INELIGIBLE, reason = reason, date_will_be_eligible = date_will_be_eligible)
```
NickSchimek: There should not be spaces around the equal sign per the pep 8 style guide.

> Don't use spaces around the = sign when used to indicate a keyword argument,

KentShikama: We may want to consider adding a post-commit hook for [`black`](https://black.readthedocs.io/en/stable/).

PR #566, opened 2019-11-17
Charges

SteelFace87:
This PR is to get closer to fulfill the requirements of issue #87, the only missing piece will be to not use a mock function, but to actually get the search functionality to talk to the back end.


NickSchimek:
Thanks for finishing this up @SteelFace87 ! As far as the visual inspection goes the results look good to me. I created issue #567 to get this connected to the backend. @KentShikama @maxkwallace can you review the code?

Screenshot 2019-11-16 18 16 42

KentShikama:
@SteelFace87 Do you need help resolving merge conflicts? Since you've allowed edits to your branch here, I could go ahead and fix them if you're busy.

src/frontend/src/components/Type/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    const eligible = (
+      <div className="relative mb3">
+        <i
+          aria-hidden="true"
+          className="absolute fas fa-check-circle green"
+        ></i>
+        <div className="ml3 pl1">
+          <span className="fw7">Type:</span> Eligible{' '}
+          <span className="nowrap">{type_eligibility_reason}</span>
```
KentShikama:
```suggestion
          <span className="nowrap">({type_eligibility_reason})</span>
```

Otherwise we will get messages like "Eligible Eligible under 137.225(1)(b)"

src/frontend/src/components/Type/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        </div>
+      </div>
+    );
+
+    const ineligible = (
+      <div className="relative mb3">
+        <i aria-hidden="true" className="absolute fas fa-times-circle red"></i>
+        <div className="ml3 pl1">
+          <span className="fw7">Type:</span> Ineligible{' '}
+          <span className="nowrap">{type_eligibility_reason}</span>
```
KentShikama:
```suggestion
          <span className="nowrap">({type_eligibility_reason})</span>
```

src/frontend/src/components/Type/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        </div>
+      </div>
+    );
+
+    const type = () => {
+      if (type_eligibility === true) {
+        return eligible;
+      } else if (type_eligibility === 'None') {
+        return review;
+      } else if (time_eligibility === false) {
```
KentShikama:
```suggestion
      } else if (type_eligibility === false) {
```

Looks like a typo

src/frontend/src/components/Type/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        return ineligible;
+      }
+      return 'error';
```
KentShikama: Nitpick: I would prefer an else clause

src/frontend/src/components/Type/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    return type();
```
KentShikama: Nitpick: My preference would be just to return as a function

```typescript
    return function type() {
      if (type_eligibility) {
        return eligible;
      } else if (type_eligibility === 'None') {
        return review;
      } else if (type_eligibility === false) {
        return ineligible;
      }
      return 'error';
    };
```

src/frontend/src/components/Type/index.tsx
SteelFace87: I made the change, this does seem more in line with the if statements.   I wonder if we want to return something more than just 'error'

src/frontend/src/components/Type/index.tsx
SteelFace87: Do you mean like this? 
`
} else {
        return;
      }
`

src/frontend/src/components/Type/index.tsx
SteelFace87: See the comment below :)

src/frontend/src/components/SearchResults/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  statute: string;
+  expungement_result: any;
+  name: string;
+  disposition: {
+    ruling: string;
+    date: string;
+  };
+}
+
+export interface CaseProp {
```
maxkwallace: What does Prop mean here? I'm not sure if this is the React "prop" or not. We should keep React-specific terms out of the type structure, since we don't want React implementation details mixed into our types and domain modeling, which should be entirely about the thing itself and not the implementation.

src/frontend/src/components/Cases/index.tsx
```diff
@@ -1,12 +1,14 @@
...
+import { CaseProp } from '../SearchResults/types';
 
 interface Props {
-  cases: any[];
+  cases: CaseProp[];
 }
 
 export default class Cases extends React.Component<Props> {
   render() {
+    console.log('cases', this.props.cases);
```
maxkwallace: Should this still be in here?

src/frontend/src/components/Charges/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+import Charge from '../Charge';
+import { ChargeType } from '../SearchResults/types';
+
+interface Props {
+  charges: ChargeType[];
+}
+
+export default class Charges extends React.Component<Props> {
+  render() {
+    console.log('mapping', this.props.charges);
```
maxkwallace: See above.

src/frontend/src/components/Type/index.tsx
KentShikama: Yeah; we probably want to log some kind error message too.

maxkwallace: Why can't the code just be this?

    if (type_eligibility === true) {
      return eligible;
    } else if (type_eligibility === 'None') {
      return review;
    } else if (type_eligibility === false) {
      return ineligible;
    } else {
      return;
    }

src/frontend/src/components/Case/index.tsx
```diff
@@ -1,30 +1,40 @@
 import React from 'react';
 import { CaseProps } from '../../redux/records/types';
+import Charges from '../Charges';
 
 export default class Cases extends React.Component<CaseProps> {
   render() {
-    const { name, case_number, birth_year, balance_due } = this.props.case;
+    const {
+      name,
+      case_number,
```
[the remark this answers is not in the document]
KentShikama: Field names are set by the JSON response. Maybe ticket this for now and we can discuss during the meetup if we want to take care of it using on the frontend with something like `https://lodash.com/docs/4.17.15#camelCase` or on the backend?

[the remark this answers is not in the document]
KentShikama: This also affects https://github.com/codeforpdx/recordexpungPDX/pull/557

src/frontend/src/components/Charges/index.tsx
SteelFace87: fixed

src/frontend/src/components/Cases/index.tsx
SteelFace87: fixed

src/frontend/src/components/SearchResults/types.ts
SteelFace87: I have renamed it to CaseType

src/frontend/src/components/Type/index.tsx
SteelFace87: I have updated it to be this way :)

src/frontend/src/components/Type/index.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+
+interface Props {
+  expungement_result: any;
+}
+
+export default class Type extends React.Component<Props> {
```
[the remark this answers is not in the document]
SteelFace87: I have renamed Type to RecordType

src/frontend/src/components/Time/index.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+interface Props {
+  expungement_result: any;
+}
+
+export default class Time extends React.Component<Props> {
```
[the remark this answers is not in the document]
SteelFace87: It is now RecordTime

src/frontend/src/components/Case/index.tsx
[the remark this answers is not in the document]
maxkwallace: @KentShikama I won't be at the meeting this week so if you think it's best to just ticket this, that's fine with me. 

I'm aware these are the field names from the JSON responses. The correct way to handle this is probably to have constructors for these objects defined next to their types. This could be plain Javascript:

    createCaseType(input) {
      return {
        balanceDue: input.balance_due
        birthYear: input.birth_year
        ...
      }
    }

Or a Typescript class:

    class Greeter {
      balanceDue: number;
      birthYear: number; 

      constructor(input: Object) {
        this.balanceDue = input.balance_due;
        this.birthYear = input.birth_year;
        ...
      }
    }

This code is straightforward to write and maintain. 

I have also used https://github.com/domchristie/humps in the past which worked well. IMO the backend should not send responses using camelCase by default because APIs generally shouldn't know that much about how they're being used.

PR #572, opened 2019-11-20
Replace 'None' with null where applicable in search JSON response

KentShikama:
This is a follow up from #566 ; issue was found by @NickSchimek .

We would receive a "Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null" error on a search result. Eventually we should validate against a JSON schema so that we don't have these kinds of errors.

A number of pre-fix refactoring commits are added. Review should be easier if done commit by commit. Supersedes #543.


src/frontend/src/service/api-service.test.ts
```diff
@@ -52,8 +52,11 @@ describe('API SERVICE TEST', () => {
     };
     apiService(mockStore.dispatch, request).then((response: any) => {
       expect(response).toEqual({
-        data: fakeRecord,
-        errors: []
+        data: {
+          data: {
+            record: fakeRecord
```
NickSchimek: The backend is being wrapped in two data objects?

KentShikama: Once explicitly by us at https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/endpoints/search.py#L50 and then once by I believe Flask.

NickSchimek: Understood.

Issue #49, opened 2019-01-03
Endpoint: /api/stats


erikjasso:
Endpoint for reporting on system statistics such as number of searches done, number of records expunged, and anything else we think is interesting

erikjasso:
See `src/backend/expungeservice/endpoints/hello.py` for an example of a Flask endpoint

erikjasso:
This endpoint could also return data for displaying in a chart

wittejm:
We've decided against keeping or using any results data

PR #576, opened 2019-11-21
Save anonymized search results to database

wittejm:
This saves some minimal search results data to the database every time the search endpoint is called: 

- Timestamp of search 
- Hash of the current user's user_id and the search params used.
- The number of charges found in the record
- The number of eligible charges found in the record
- Edit: Incuding user_id

The search parameters and user_id are concatenated and then hashed deterministically so that we can keep track of unique searches.

Edit: Michael et al. agreed that holding onto the user_id information is fine. 
~~The search result doesn't include the user_id itself. This is to make individual users' levels of activity hidden, once there is more than one user of the app. The tradeoff is that it doesn't enable a user to check only their own usage statistics. If we want to add that feature but protect that info from other users, we'll need to add an additional security mechanism.~~

I updated the search endpoint test case to mock this added feature, so that the test case doesn't have a side effect on the database.

Something I will change before merging is that `current_user.user_id` is currently being read in **search.py** but should be read from **stats.py** . I was having trouble getting the test case to work that way. Will figure it out though.

Related to #49 and #119 .

### DIFF
--- a/src/backend/expungeservice/stats/stats.py
+++ b/src/backend/expungeservice/stats/stats.py
@@ -4,9 +4,12 @@ import hashlib
 
 from expungeservice.database.db_util import rollback_errors
 from expungeservice.models.expungement_result import EligibilityStatus
+from flask_login import current_user
 
 
-def save_result(user_id, request_data, record):
+def save_result(request_data, record):
+
+    user_id = current_user.user_id
 
     search_param_string = (
         user_id +
--- a/src/backend/expungeservice/stats/stats.py
+++ b/src/backend/expungeservice/stats/stats.py
@@ -1,4 +1,5 @@
 from flask import g
+import uuid
 
 from expungeservice.database.db_util import rollback_errors
 from expungeservice.models.expungement_result import EligibilityStatus
@@ -19,15 +20,15 @@ def save_result(user_id, request_data, record):
         c.expungement_result.type_eligibility.status == EligibilityStatus.ELIGIBLE])
 
     insert_result = db_insert_result(
-        g.database, hashed_search_params, num_charges, num_eligible_charges)
+        g.database, user_id, hashed_search_params, num_charges, num_eligible_charges)
 
 
 @rollback_errors
-def db_insert_result(database, hashed_search_params, num_charges, num_eligible_charges):
+def db_insert_result(database, user_id, hashed_search_params, num_charges, num_eligible_charges):
 
     result= database.cursor.execute(
         """
-        INSERT INTO  SEARCH_RESULTS(search_result_id, hashed_search_params, num_charges, num_eligible_charges )
-        VALUES ( uuid_generate_v4(), %(params)s, %(num_charges)s, %(num_eligible_charges)s);
-        """, {'params': hashed_search_params, 'num_charges': num_charges,
+        INSERT INTO  SEARCH_RESULTS(search_result_id, user_id, hashed_search_params, num_charges, num_eligible_charges )
+        VALUES ( uuid_generate_v4(), %(user_id)s, %(params)s, %(num_charges)s, %(num_eligible_charges)s);
+        """, {'user_id': uuid.UUID(user_id).hex, 'params': hashed_search_params, 'num_charges': num_charges,
               'num_eligible_charges': num_eligible_charges})
--- a/src/backend/expungeservice/stats/stats.py
+++ b/src/backend/expungeservice/stats/stats.py
@@ -1,5 +1,6 @@
 from flask import g
 import uuid
+import hashlib
 
 from expungeservice.database.db_util import rollback_errors
 from expungeservice.models.expungement_result import EligibilityStatus
@@ -7,12 +8,14 @@ from expungeservice.models.expungement_result import EligibilityStatus
 
 def save_result(user_id, request_data, record):
 
-    hashed_search_params = hash(
+    search_param_string = (
         user_id +
         request_data["first_name"] +
         request_data["last_name"] +
         request_data["middle_name"] +
         request_data["birth_date"])
+
+    hashed_search_params = hashlib.sha256(search_param_string.encode()).hexdigest()
 
     num_charges = len(record.charges)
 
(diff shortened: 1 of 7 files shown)


src/backend/expungeservice/stats/stats.py
```diff
@@ -0,0 +1,33 @@
...
+
+    num_eligible_charges = len([ c for c in record.charges if
+        c.expungement_result.type_eligibility.status == EligibilityStatus.ELIGIBLE])
+
+    insert_result = db_insert_result(
+        g.database, hashed_search_params, num_charges, num_eligible_charges)
+
+
+@rollback_errors
+def db_insert_result(database, hashed_search_params, num_charges, num_eligible_charges):
```
KentShikama: `_db_insert_result`

src/backend/expungeservice/stats/stats.py
```diff
@@ -0,0 +1,33 @@
...
+
+    result= database.cursor.execute(
```
KentShikama: `result` is unused and should be removed

src/backend/expungeservice/stats/stats.py
```diff
@@ -0,0 +1,33 @@
...
+        request_data["last_name"] +
+        request_data["middle_name"] +
+        request_data["birth_date"])
+
+    num_charges = len(record.charges)
+
+    num_eligible_charges = len([ c for c in record.charges if
+        c.expungement_result.type_eligibility.status == EligibilityStatus.ELIGIBLE])
+
+    insert_result = db_insert_result(
```
KentShikama: `insert_result` is unused and should be removed

src/backend/tests/stats/test_stats_save_result.py
```diff
@@ -0,0 +1,57 @@
...
+            save_result(user_id, request_data, record)
+
+            expungeservice.request.after(None)
+
+            hashed_search_params = hash(
+                user_id +
+                "first_name" +
+                "last_name" +
+                "middle_name" +
+                "birth_date")
```
KentShikama: I would prefer if realistic values were used as on the initial read I got confused why you were hashing the field names (when in fact you are hashing values that just happened to be the same as the field names).

src/backend/tests/stats/test_stats_save_result.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class TestStats():
+
+    def test_save_result(self):
+
+        with expungeservice.create_app("development").app_context():
+
+            expungeservice.request.before()
+            database = get_database()
```
KentShikama: There is some redundancy in the above two lines: see https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/request/request.py#L8

wittejm:
@KentShikama thanks for the review. I made all the suggested fixes. For the hash I chose SHA256, didn't seem obvious that there was a particular best choice of algorithm. 

Additional comment: I decided to make the user_id field nullable because we could have a search result tied to a user whose account gets deleted, but we would want to retain the search information. We don't have an endpoint for deleting a user account, but when we do it will need to update the search_result table to replace all instances of that user id with `null`.

src/backend/tests/stats/test_stats_save_result.py
```diff
@@ -0,0 +1,64 @@
...
+            assert result["num_eligible_charges"] == 6
+            assert result["num_charges"] == 9
+
+            g.database.cursor.execute(
+                    """
+                    DELETE FROM SEARCH_RESULTS
+                    WHERE hashed_search_params = %(hashed_search_params)s
+                    ;
+                    """,{'hashed_search_params': hashed_search_params})
+            g.database.connection.commit()
```
KentShikama: Probably good to be consistent with the above. See https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/request/request.py#L11 ; I wouldn't even mind removing this wrapper: there isn't much purpose in just returning the passed in argument.

wittejm:
I made two more changes (final 2 commits): 
- the `save_results()` call made in the search endpoint is now wrapped with `try/except` so that it will never break the search endpoint. It catches any exception, which probably isn't great but I wanted to be cautious about adding a potential point of failure to the endpoint, especially since we don't have a deploy-staging pipeline yet.
   - this commit includes expanded test code, including a test case that jointly tests the search endpoint response and the `save_results` effect on the database. It's arguably not a unit test but I wanted to test that save_results is working correctly within its calling context.
- The second change is just to read the current_user id within the `save_results` function and not `search.py`
   - this needed a little more scaffolding in the test case so that flask's `current_user` is readable when called "directly" in the unit test. This was the best approach I could figure out :man_shrugging:

src/backend/tests/stats/test_stats_save_result.py
```diff
@@ -0,0 +1,53 @@
...
+from tests.endpoints.endpoint_util import EndpointShared
+from flask.views import MethodView
+
+
+class TestStatsView(MethodView):
+    """
+    To get the direct call to stats.save_result() to work,
+    this endpoint wrapper is needed so that the flask.current_user
+    object gets populated with user_id
+    """
```
KentShikama: Can you elaborate on how `flask.current_user` gets populated with this mock endpoint and why it doesn't without it?

wittejm: Thanks for asking! It did seem weird to require this workaround, so I looked around a bit more and found a better solution. 

Before I decided to wrap the function call in an endpoint, I was trying to use the usual testing wrapper code for flask:

```
with flaskr.app.test_client() as client:
        with flaskr.app.app_context():
                test_stuff()
````

To the best of my understanding: this would create a flask app context that was distinct from the app context created from the test's `login()` call, and the `current_user` object isn't carried over between contexts. flask_login normally handles the `current_user` data, but that underlying functionality seemed to only kick in when making an actual endpoint call on the app, and not from just using the `with` context handler. 

Makes enough sense, I guess.

I found that I

PR #577, opened 2019-11-21
Convert Charge into a dataclass

KentShikama:
~~This was not caught in the existing unit tests because we explicitly pass in a datetime object (`charge = ChargeFactory.create(date=self.THREE_YEARS_AGO)`) but the crawler returns a string. Note that we do properly convert to a datetime object when we build the Case object.~~

Edit: As pointed out by @NickSchimek below we do pass in strings in the unit tests; there is no issue here.

Also clarifies that the Disposition may be None (Optional).

### DIFF
--- a/src/backend/expungeservice/models/charge_types/charge.py
+++ b/src/backend/expungeservice/models/charge_types/charge.py
@@ -34,13 +34,13 @@ class Charge:
     def acquitted(self):
         return self.disposition and self.disposition.ruling[0:9].lower() != 'convicted'
 
-    def convicted(self):
-        return self.disposition and not self.acquitted()
+    def __convicted(self):
+        return not self.acquitted()
 
     def recent_conviction(self):
         ten_years_ago = (date_class.today() + relativedelta(years=-10))
-        if self.convicted() and self.disposition:
-            return self.disposition.date > ten_years_ago
+        if self.__convicted():
+            return self.disposition.date > ten_years_ago # type: ignore
         else:
             return False
 
--- a/src/backend/expungeservice/models/charge_types/charge.py
+++ b/src/backend/expungeservice/models/charge_types/charge.py
@@ -9,7 +9,7 @@ from dateutil.relativedelta import relativedelta
 from expungeservice.models.disposition import Disposition
 from expungeservice.models.expungement_result import ExpungementResult, TimeEligibility, EligibilityStatus
 
-@dataclass
+@dataclass(eq=False)
 class Charge:
     name: str
     statute: str
--- a/src/backend/expungeservice/models/charge_types/charge.py
+++ b/src/backend/expungeservice/models/charge_types/charge.py
@@ -1,24 +1,29 @@
 import weakref
+from dataclasses import dataclass, InitVar, field
 
-from datetime import datetime
-from datetime import date as date_class
+from datetime import date as date_class, datetime
+from typing import Optional
+
 from dateutil.relativedelta import relativedelta
+
+from expungeservice.models.disposition import Disposition
 from expungeservice.models.expungement_result import ExpungementResult, TimeEligibility, EligibilityStatus
 
-
+@dataclass
 class Charge:
+    name: str
+    statute: str
+    level: str
+    date: date_class
+    disposition: Optional[Disposition]
+    expungement_result: ExpungementResult = field(init=False)
+    _chapter: Optional[str]
+    _section: str
+    _case: weakref.ref
 
-    def __init__(self, case, name, statute, level, date, chapter, section, disposition=None):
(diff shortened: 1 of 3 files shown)


NickSchimek:
I’ve been waiting for this. I will review thoroughly in the next couple of days

src/backend/expungeservice/models/charge_types/charge.py
```diff
@@ -34,7 +39,7 @@ def convicted(self):
 
     def recent_conviction(self):
         ten_years_ago = (date_class.today() + relativedelta(years=-10))
-        if self.convicted():
+        if self.convicted() and self.disposition:
```
NickSchimek: This extra check of `self.disposition` is not needed. Likewise the self.disposition on line 38 is also not needed.

KentShikama:
@NickSchimek Thanks for the catch and apologies for the misleading commit message. The tests pass in the date as a string indeed. I got too used to seeing constants as `THREE_YEARS_AGO` as a date object. I've amended the commit message to clarify that there is no issue here.

KentShikama:
@NickSchimek PTAL

I believe https://github.com/codeforpdx/recordexpungPDX/pull/577/commits/950db0ec02ae081924dc1afc975c3a5f343a73cb addresses your last paragraph.

NickSchimek:
The thing I'm most worried about is having multiple Charges with the same attributes. If the Charge class is converted to a data-class, then we cannot differentiate between two charges with the same attributes. Converting a class to a data-class that is more apt to be a regular class has the potential to create hard to find bugs.

@KentShikama There can be instances of a Case class having multiple charges with the same attributes. I have seen them. And it is not unheard of to be arrested for multiple counts of the same charge. Hence, there can be multiple charges on the case with the same attributes.

Here is an example: A record with one case consisting of two counts of robbery.
```python
    def test_mrc_is_not_equal_to_2nd_mrc(self):
        mrc_charge = ChargeFactory.create(case=self.case)
        second_mrc_charge = ChargeFactory.create(case=self.case)

        assert mrc_charge != second_mrc_charge
```

^^^ The above code will pass with a data-class, but fail with a regular class.

Having the Charge class as a data-class is something I was considering when I first wrote this, but I felt I did not have enough information to make that decision at that time. Because of that, I created all the required classes as regular classes. Case and some other classes have now been converted to data classes (Thanks!). However, I feel the Charge class should be left alone. It is something I've given a lot of thought about and am very hesitant to allow this change based on the reasons mentioned above. Please let me know if you, @wittejm or @erikjasso feels any different.

KentShikama:
@NickSchimek I had originally assumed that multiple charges with the same attributes couldn't appear in one case. As clarified on Slack, this isn't the case, and I appreciate the catch. I'll follow up with some commits soon.

KentShikama:
@NickSchimek PTAL. I believe https://github.com/codeforpdx/recordexpungPDX/pull/577/commits/5bce931afc1e4ff422c892152ea928f799ffa772 will actually address your concerns this time.

KentShikama:
I also just rebased out the test for "Add test for when the two charges are the same besides their parent case" as it no longer makes sense to have.

NickSchimek:
What's the purpose for labeling this as a data class but not treating it as a data class? Seems like this could cause confusion later on.

KentShikama:
@NickSchimek Did you read the commit message?

NickSchimek:
Thank you. That all makes sense, however I have a couple questions:
> In the future I want to add in some generative tests and it will be easier to do so if everything under models/ is a dataclass

Why is it easier when it's a data class? Can you please explain for me, and anyone else who looks at this code in the future to see why these changes were done?

Is there an issue for generative tests? This PR should reference it.

> 4. Fundamentally, Charge exists to hold data (unlike Expunger which operates on data) and thus marking it as a dataclass to make that clear is valuable.

Fundamentally, yes it does hold data. But how do we know it will never operate on data? I guess I'm a bit more hesitant and don't want to assume how the future will look. I would like to understand more why the need to make this change is absolutely necessary. I don't claim to know everything so please talk me into why these changes are required.

KentShikama:
See https://github.com/codeforpdx/recordexpungPDX/pull/610 for the generative tests PR.

> But how do we know it will never operate on data?

Because it is marked as a dataclass. If someone tries to add, for example, expunger logic to this class, the fact that it is marked as a dataclass is a red flag during a PR review saying "wait a minute, shouldn't we create a new class for these operations"?

KentShikama:
"Dataclasses are handled natively by the inference from type hints." - https://hypothesis.readthedocs.io/en/latest/data.html?highlight=dataclass#hypothesis.strategies.builds

PR #579, opened 2019-11-21
Fix case where disposition is not set

### LINKED ISSUE #578: Fix case where disposition is not set
Originally discovered by Michael Zhang.

Sometimes there are charges without a corresponding disposition in the OECI database. This will just crash the server saying that upon attempt at serialization disposition.date didn't exist as disposition was None.

KentShikama:
Closes #578 

Sometimes there are charges without a corresponding disposition in the OECI database. This will just crash the server saying that upon attempt at serialization disposition.date didn't exist as disposition was None.

~~TODO: I still owe a non-manual test for this issue. However, I've gone ahead and pushed this to https://recordexpungpdx.herokuapp.com/ so that Michael won't be blocked from testing.~~

Edit: Tests have been added.

### DIFF
--- a/src/backend/Pipfile
+++ b/src/backend/Pipfile
@@ -21,6 +21,7 @@ cryptography = "*"
 flask-login = "==0.4.1"
 dacite = "==1.0.2"
 mypy = "==0.740"
+pytest-lazy-fixture = "==0.6.2"
 
 [dev-packages]
 
--- a/src/backend/Pipfile.lock
+++ b/src/backend/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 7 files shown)


NickSchimek:
@KentShikama It would be good to go through and hook up the record fixtures already in place to create a system test for the entire backend. Similar to what is done in `test_crawler_expunger.py`. Because the record fixtures cover a lot of different real world use cases that would've caught this issue.

NickSchimek:
>  I've gone ahead and pushed this to https://recordexpungpdx.herokuapp.com/ so that Michael won't be blocked from testing

Thanks for getting that pushed. 🎉

KentShikama:
@NickSchimek @wittejm PTAL as I've added the backend tests that I owe.

wittejm:
Cool, LGTM

KentShikama:
@NickSchimek I've addressed unused imports with https://github.com/codeforpdx/recordexpungPDX/pull/588

Issue #582, opened 2019-11-25
Create reusable error component 


ForrestLonganecker:
We are currently creating error related logic on a case by case basis. It would be helpful to have one way to handle errors.

### This task is done when:
 - There is an error component that will replace a failing component
 - The error component is connected to the store so we can move error handling into Redux/services

PR #583, opened 2019-11-25
user component now integrated with /api/users GET endpoint

ForrestLonganecker:
This PR now is pulling users from our database: #91 

Notes:

- Adjusted the initial users populated in the Developer database: "user@email.com" is no longer an Admin.  
- Adjusted the response object for the `/api/users` GET to match up with expected variable names in the front-end. changed: `user_id` to `id` and `group_name` to `group`.   
Previously I had a function that was converting the *snake_case* to the proper format but this solution allows us to pass the response object in directly.  
- Converted `<Spinner />` component to a generic `<LoadingSpinner inputString={} />` we are now able to use this loading spinner in anywhere we need to have it and pass in a string such as "Users" or "your search results".  

To-Do:

- currently I am displaying errors for this component, it would be good to isolate this logic and connect it to the redux-store so we can plug the error component in wherever we need it. I have created an issue for this: #582

### DIFF
--- a/src/frontend/src/components/NotAuthorized/index.tsx
+++ b/src/frontend/src/components/NotAuthorized/index.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+class NotAuthorized extends React.Component {
+  render() {
+    return (
+      <p className="bg-white mv4 pa4 br3 fw6 tc shadow br3">
+        You are not authorized to view this content, please contact system
+        administraitor.
+      </p>
+    );
+  }
+}
+
+export default NotAuthorized;
--- a/src/frontend/src/components/NotAuthorized/index.tsx
+++ b/src/frontend/src/components/NotAuthorized/index.tsx
@@ -5,7 +5,7 @@ class NotAuthorized extends React.Component {
     return (
       <p className="bg-white mv4 pa4 br3 fw6 tc shadow br3">
         You are not authorized to view this content, please contact system
-        administraitor.
+        administrator.
       </p>
     );
   }
--- a/src/frontend/src/components/TechnicalDifficulties/index.tsx
+++ b/src/frontend/src/components/TechnicalDifficulties/index.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+class TechnicalDifficulties extends React.Component {
+  render() {
+    return (
+      <p className="bg-white mv4 pa4 br3 fw6 tc shadow br3">
+        Something went wrong, please contact system administraitor to report
+        technical difficulties.
+      </p>
+    );
+  }
+}
+
+export default TechnicalDifficulties;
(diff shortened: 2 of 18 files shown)


src/frontend/src/components/NotAuthorized/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+class NotAuthorized extends React.Component {
+  render() {
+    return (
+      <p className="bg-white mv4 pa4 br3 fw6 tc shadow br3">
+        You are not authorized to view this content, please contact system
+        administraitor.
```
KentShikama: administrator

src/frontend/src/components/TechnicalDifficulties/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+class TechnicalDifficulties extends React.Component {
+  render() {
+    return (
+      <p className="bg-white mv4 pa4 br3 fw6 tc shadow br3">
+        Something went wrong, please contact system administraitor to report
```
KentShikama: administrator

src/frontend/src/components/UserList/index.tsx
```diff
@@ -2,32 +2,57 @@ import React from 'react';
...
 
-      return returnList;
-    }
+    return returnList;
   };
 
-  public render() {
-    return (
+  render() {
+    return this.props.users.userList.length > 0 ? (
```
KentShikama: Nitpick: I think the True case can be wrapped in its own function just like `displayNoUsers`.

src/frontend/src/components/UserList/index.tsx
```diff
@@ -2,32 +2,57 @@ import React from 'react';
...
-        return <User key={user.id} user={user} />;
-      });
+  displayNoUsers = () => {
+    return this.state.errorType === 'none' ? (
+      <LoadingSpinner inputString={'Users'} />
+    ) : this.state.errorType === 'unauthorized' ? (
+      <NotAuthorized />
+    ) : (
+      <TechnicalDifficulties />
+    );
```
KentShikama: I would prefer not chaining ternary operators

src/frontend/src/components/UserList/index.tsx
ForrestLonganecker: Simple fix! I will make that change.

ForrestLonganecker:
@KentShikama Thank you for the review! 

I realized that a non admin could see the userlist if:

- Admin logs in and views userList page
- Admin logs out
- User logs in and navigates to userList page
- The state is not cleared so the list is still visible

I added a `CLEAR_USERS` action that is now called on logout and will fix this issue.

KentShikama:
Are you referring to if the non-admin happens to use the same browser as the admin (aka they share a computer)?

ForrestLonganecker:
Exactly

Issue #584, opened 2019-11-25
"Unknown type or time eligibility" in frontend search results 


wittejm:
Some records evaluated by the Expunger have the result:
```
 "expungement_result": {
                        "date_of_eligibility": null,
                        "time_eligibility": null,
                        "time_eligibility_reason": "",
                        "type_eligibility": true,
                        "type_eligibility_reason": "..."
                    },
```

The frontend Search results page doesn't handle this case when rendering a status label for the charge , so it shows the default value, "Unknown type or time eligibility"

[src/frontend/src/components/Eligibility/index.tsx](https://github.com/codeforpdx/recordexpungPDX/blob/master/src/frontend/src/components/Eligibility/index.tsx)

@NickSchimek Are there cases where  `"type_eligibility": true` with `"time_eligibility": null` is the desired output? Or is this an expunger bug? I'm digging into the expunger logic myself to see if I can figure out an answer.

The record showing this use case is in the password-protected bug reports page. I don't want to post the search params here.

KentShikama:
We currently abort all analysis when a person still has an open case. We need to handle cases where the status is "Bankruptcy pending" as closed.

KentShikama:
Although not relevant to person who originally triggered this issue to be opened, if there is a user who really does still have an open case, we would see a similar result.

NickSchimek:
> "type_eligibility": true with "time_eligibility": null is the desired output?
Yes that is the desired output as the time_eligibility would not be ran if there is an open case.

For your information: If the record has an open case then charges cannot be expunged. Hence, time analysis is not ran.

NickSchimek:
Hopefully there are not too many more case status types that we are unaware of. Of course we could always bypass that open case check upon user request.

NickSchimek:
It seems this was causing an error on the frontend? The frontend should default to ineligible charge if the preceding conditions do not pass. @KentShikama are you able to verify this logic?

wittejm:
> For your information: If the record has an open case then charges cannot be expunged. Hence, time analysis is not ran.

If time analysis doesn't run for a given case it would be good to show the reason e.g. in the `time_eligibility_reason` field. Though if it's the same for all the cases because the record has an open case, this information could potentially go into the "summary panel" that has been mentioned as a new feature.

hmarcks:
@wittejm there's a discussion about this in #586 if you want to check it out

PR #585, opened 2019-11-26
Handle case status "Bankruptcy Pending" as closed

KentShikama:
Closes https://github.com/codeforpdx/recordexpungPDX/issues/584

We currently abort all analysis when a person still has an open case. We need to handle cases where the status is "Bankruptcy Pending" as closed.

I will follow up with a test here but for now I'm pushing to production to unblock Michael.

### DIFF
--- a/src/backend/expungeservice/models/case.py
+++ b/src/backend/expungeservice/models/case.py
@@ -69,4 +69,5 @@ class Case:
         return 'violation' in self.violation_type.lower() or 'municipal parking' == self.violation_type.lower()
 
     def _closed(self):
-        return self.current_status == 'Closed' or self.current_status == 'Inactive' or self.current_status == 'Purgable' or self.current_status == 'Bankruptcy Pending'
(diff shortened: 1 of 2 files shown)


src/backend/expungeservice/models/case.py
```diff
@@ -69,4 +69,4 @@ def _ignore_open_case(self):
         return 'violation' in self.violation_type.lower() or 'municipal parking' == self.violation_type.lower()
 
     def _closed(self):
-        return self.current_status == 'Closed' or self.current_status == 'Inactive' or self.current_status == 'Purgable'
+        return self.current_status == 'Closed' or self.current_status == 'Inactive' or self.current_status == 'Purgable' or self.current_status == 'Bankruptcy Pending'
```
NickSchimek: It might be good to extract these keywords into a tuple. And then we can do `self.current_status in CLOSED_STATUS`

KentShikama:
@NickSchimek PTAL

Issue #545, opened 2019-11-13
Fix KeyError: 'oeci_token' when user tries to search before OECI login


KentShikama:
The error message is as follows

```
2019-11-13T08:59:05.042120+00:00 app[web.1]: File "./expungeservice/endpoints/search.py", line 28, in post
2019-11-13T08:59:05.042123+00:00 app[web.1]: decrypted_credentials = cipher.decrypt(request.cookies["oeci_token"])
2019-11-13T08:59:05.042129+00:00 app[web.1]: KeyError: 'oeci_token'
```

We should never have a 500 server error. The simplest fix is to return a 40x error when the `oeci_token` is missing. Ideally we should make it impossible for the user to search before they login to OECI based on front-end flows.

NickSchimek:
How can I duplicate this error?

KentShikama:
Do a valid search without logging in to OECI.

NickSchimek:
oh I see. We need to prevent access to search page when login creds don't exist. When that occurs it should redirect user to login screen. Thanks for reporting this! One thing I haven't done yet is play around with the UI. Great catch

ForrestLonganecker:
Just opened a PR that redirects on the client side if no OECI token is present.
#587

PR #587, opened 2019-11-26
Redirect to OECI login from search page

ForrestLonganecker:
- Added check for OECI token on componentDidMount() in AllRecords component.
This should take care of the front end side of issue #545 

- Adjusted conditional logic to ensure we are showing "No search results found."

### DIFF
--- a/src/frontend/src/containers/AllRecords.tsx
+++ b/src/frontend/src/containers/AllRecords.tsx
@@ -9,6 +9,7 @@ import {
 import RecordSearch from '../components/RecordSearch';
 import SearchResults from '../components/SearchResults';
 import AllStatus from './AllStatus';
+import { checkOeciRedirect } from '../service/cookie-service';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


src/frontend/src/containers/AllRecords.tsx
```diff
@@ -17,18 +18,24 @@ type Props = {
...
+
   componentWillUnmount() {
     this.props.clearRecords();
   }
 
   render() {
-    // Fetch has now been passed down to RecordSearch componant and is triggered when submit button is pressed.
-    // Currentlly no matter what the user puts in the search fields, it still triggers a mock fetch with pre determined API response
+    // Fetch has now been passed down to RecordSearch component and is triggered when submit button is pressed.
+    // Currently no matter what the user puts in the search fields, it still triggers a mock fetch with pre determined API response
```
KentShikama: This comment is now outdated and can be removed.

Issue #586, opened 2019-11-26
Handle case in which person has a record with an open case


KentShikama:
Currently the frontend search result of a person with an open case would be to have all charges that are type eligible shown as "Unknown type or time eligibility" as reported on https://github.com/codeforpdx/recordexpungPDX/issues/584. 

Perhaps we should instead just have a the search result return a giant "there is an open case" banner at the top and show all time analyzer result areas as "Not analyzed".

hmarcks:
@KentShikama the original plan was to show the disposition and time eligibility as open, similar to this mockup. I like the idea of a banner as well to make it more clear:

![open-charge](https://user-images.githubusercontent.com/6887982/69600655-13a22480-0fd7-11ea-9cda-effe6b79e45a.png)

cc @NickSchimek

NickSchimek:
It would probably be good to have a link on the banner to run the same request that will tell the backend to ignore the open cases. This would be good for those edge cases where the user knows there are no open cases, but our app is reporting that there is. They would still be able to get their analysis that way.

hmarcks:
Ohh I like it, here's what that could look like:
![open-charge-v2](https://user-images.githubusercontent.com/6887982/69664444-03795c00-104e-11ea-8899-2e3fba35fcd1.png)

hmarcks:
I wonder if a better approach is to give the option to mark any open charges as either convicted or dismissed, and possibly add a date to give an estimate with the open case considered. I think just ignoring the charge gives an accurate picture of what could be expungeable because the case will eventually have to be considered. It's obviously more involved but thought I'd throw it out for consideration.

wittejm:
@hmarcks I think that (also applied to #652) would be an awesome feature. 

Showing a warning and skipping the open cases or missing disposition errors in the analysis seems like an ok (not great) stopgap for the issue. And in fact the expunger can always do the time analysis, and the frontend can just check whether there are errors attached to the record. If there are errors, it can show the informative warning message for the error, with the button to ignore and show analysis. That way the expunger doesn't have to run twice to get the same data, to do something a user will probably want to do anyway.

Thinking through your suggestion a little, I think the backend would need an additional endpoint that accepts the record data object back from the frontend, including the "fixed" charge dispositions, so that the crawler doesn't need to trying merging a fetched OECI record with the edited data from the user. 

[...]

PR #591, opened 2019-11-27
Moved errors fields to record object 

### LINKED ISSUE #261: Append errors to record object
## Description
Currently any errors that are recorded within the expunger are added to the error attribute on the expunger. Now that we have a record object these should be appended to error attribute on the record object. 

This is needed because the search endpoint will return a jsonified record object to the frontend and whatever data is left in the expunger will not be sent. After this the expunger should only have a private attributes. 

### This task is done when:
 - Move current test for error field from expunger test file to the record test file
 - The error field is moved to the record object

wittejm:
The Expunger no longer has an `errors` field so any errors that it generates need to be attached to the record object. Closes #261 . 

This is a requirement for #586 , since the errors field now is returned in the search endpoint. 

Expunger errors are unrelated to flask API errors, which addresses the discussion on #261.

### DIFF
--- a/src/backend/tests/functional_tests/test_crawler_expunger.py
+++ b/src/backend/tests/functional_tests/test_crawler_expunger.py
@@ -29,7 +29,7 @@ def test_expunger_with_open_case(record_with_open_case):
     expunger = Expunger(record_with_open_case)
 
     assert not expunger.run()
-    assert expunger.record.errors == ['Open cases exist']
+    assert record_with_open_case.errors == ['Open cases exist']
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


src/backend/tests/functional_tests/test_crawler_expunger.py
```diff
@@ -29,7 +29,7 @@ def test_expunger_with_open_case(record_with_open_case):
     expunger = Expunger(record_with_open_case)
 
     assert not expunger.run()
-    assert expunger.errors == ['Open cases exist']
+    assert expunger.record.errors == ['Open cases exist']
```
NickSchimek: Thanks @wittejm I'd suggest in this case the record object should be directly tested.
```suggestion
    assert record_with_open_case.errors == ['Open cases exist']
```

PR #607, opened 2019-12-06
Add New Duii charge class

### LINKED ISSUE #592: A Dismissed DUII charge should be "Further Analysis Needed"
(edited to account for charges explicitly labeled as "diverted)"
- In general, any dismissed charge is normally expungeable.
- A DUII charge can be marked with a "diverted" disposition, meaning that it was a conviction that was reclassified to a dismissal because the defendant entered a recovery program or similar. 
- In some cases a DUII charge can be marked "dismissed" even if it was diverted.
- This type of diverted charge is not expungeable. A **diverted charge** needs to be marked as **not expungeable**. 
- A dismissed DUII charge needs to be marked with `type_eligibility.reason == 'Further Analysis Needed'`

This rule is described in [Statute 137.225 (8)(b) ](https://www.oregonlaws.org/ors/137.225)

Diversion is described here: https://www.oregonlaws.org/ors/813.200

This is related to two bugs reported from the logic bug-reports page (identifying record information is omitted here) 

### This task is done when:
 - A new test case checks that a dismissed charge with "DUII" in the name is classified correctly.
 - A new Charge subclass called DUII is implemented in expungeservice.models.charge_types
 - The DUII class has `type_eligibility.reason == 'Further Analysis Needed'` if it is a dismissed charge. Otherwise (if it's a convicted charge or a diverted charge) it is ineligible.
 - the ChargeClassifier classifies a DUII charge in the `_classification_by_statute` method. The statute is 813010. This statute needs to appear in the list before `ChargeClassifier._traffic_crime(statute)` because it is a special case of a traffic crime. 
 
I'm pinging Michael and @NickSchimek to review this for correctness before I start working on this issue.

wittejm:
duii is classified in _traffic_crime along with Level800TrafficCrime

fixes https://github.com/codeforpdx/recordexpungPDX/issues/592

### DIFF
--- a/src/backend/expungeservice/models/charge_classifier.py
+++ b/src/backend/expungeservice/models/charge_classifier.py
@@ -5,6 +5,7 @@ from expungeservice.models.charge_types.felony_class_a import FelonyClassA
 from expungeservice.models.charge_types.felony_class_b import FelonyClassB
 from expungeservice.models.charge_types.felony_class_c import FelonyClassC
 from expungeservice.models.charge_types.level_800_traffic_crime import Level800TrafficCrime
+from expungeservice.models.charge_types.duii import Duii
 from expungeservice.models.charge_types.list_b import ListB
 from expungeservice.models.charge_types.marijuana_ineligible import MarijuanaIneligible
 from expungeservice.models.charge_types.misdemeanor import Misdemeanor
@@ -79,11 +80,22 @@ class ChargeClassifier:
         if section.isdigit() and any(int(section) in statute_range for statute_range in statute_ranges):
             return PersonCrime
 
+
     @staticmethod
     def _traffic_crime(statute):
-        statute_range = range(801, 826)
-        if statute[0:3].isdigit() and int(statute[0:3]) in statute_range:
-            return Level800TrafficCrime
+
+        if statute[:3].isdigit():
+            statute_range = range(801, 826)
+
+            chapter_num = int(statute[:3])
+
+            if chapter_num == 813:
+                return Duii
+
+            elif chapter_num in statute_range:
+                return Level800TrafficCrime
+
+
 
     @staticmethod
     def _parking_ticket(statute, chapter):
--- a/src/backend/expungeservice/models/charge_classifier.py
+++ b/src/backend/expungeservice/models/charge_classifier.py
@@ -84,10 +84,11 @@ class ChargeClassifier:
     @staticmethod
     def _traffic_crime(statute):
 
-        if statute[:3].isdigit():
+        chapter = statute[:3]
+        if chapter.isdigit():
             statute_range = range(801, 826)
 
-            chapter_num = int(statute[:3])
+            chapter_num = int(chapter)
 
             if chapter_num == 813:
                 return Duii
--- a/src/backend/expungeservice/models/charge_types/duii.py
+++ b/src/backend/expungeservice/models/charge_types/duii.py
@@ -9,7 +9,7 @@ class Duii(Charge):
         if not self.disposition:
             return TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Further Analysis Needed - No disposition')
 
-        elif self.disposition.ruling[0:8].lower() == 'diverted':
+        elif 'diverted' in self.disposition.ruling.lower():
             return TypeEligibility(EligibilityStatus.INELIGIBLE, reason='Ineligible under 137.225(8)(b)')
 
         elif self.acquitted():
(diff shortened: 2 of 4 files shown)


src/backend/expungeservice/models/charge_types/duii.py
```diff
@@ -0,0 +1,25 @@
...
+
+        elif self.acquitted():
+            return TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Further Analysis Needed')
+
+        else:
+            """
+            TODO: other dispositions than "convicted" are possible and are not handled here.
+            See issue: https://github.com/codeforpdx/recordexpungPDX/issues/601
+
+            TODO: reason can be expanded to "137.225(7)(a) - Traffic offenses are ineligible"
```
wittejm: The thinking behind this second todo: It would be great IMO to start providing more readable reason strings for most expungement results. This is a larger and separate issue though.

src/backend/tests/models/charge_types/test_type_analyzer_duii.py
```diff
@@ -0,0 +1,49 @@
...
+from expungeservice.models.disposition import Disposition
+
+
+def test_duii_acquitted():
+
+    last_week = (datetime.today() - timedelta(days=7)).strftime('%m/%d/%Y')
+    single_charge = ChargeFactory.build()
+
+    single_charge['statute'] = '813.010'
+    single_charge['disposition'] = Disposition(ruling='Acquitted', date=last_week)
```
KentShikama: I forgot to mention that we probably should extract these common lines out

```python
def build_charge(statute, disposition_ruling):
    last_week = (datetime.today() - timedelta(days=7)).strftime('%m/%d/%Y')
    charge = ChargeFactory.build()
    charge['statute'] = statute
    charge['disposition'] = Disposition(ruling=disposition_ruling, date=last_week)
    return charge
```

and then use like

```
single_charge = build_charge(statute='813.010', disposition_ruling='Acquitted')
```

for each of these tests.

src/backend/expungeservice/models/charge_classifier.py
```diff
@@ -79,11 +80,22 @@ def _crime_against_person(section):
...
     @staticmethod
     def _traffic_crime(statute):
-        statute_range = range(801, 826)
-        if statute[0:3].isdigit() and int(statute[0:3]) in statute_range:
-            return Level800TrafficCrime
+
+        if statute[:3].isdigit():
+            statute_range = range(801, 826)
+
+            chapter_num = int(statute[:3])
```
KentShikama: We should probably extract a local variable here for chapter like

```python
chapter = statute[:3]
```

and replace the two usages.

src/backend/expungeservice/models/charge_types/duii.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Duii(Charge):
+
+    def _default_type_eligibility(self):
+
+        if not self.disposition:
+            return TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Further Analysis Needed - No disposition')
+
+        elif self.disposition.ruling[0:8].lower() == 'diverted':
```
KentShikama:
```suggestion
        elif 'diverted' in self.disposition.ruling.lower():
```

src/backend/tests/models/charge_types/test_level_800_traffic_crime.py
```diff
@@ -102,13 +108,13 @@ def test_dismissed_violation_is_not_type_eligible(self):
         assert charge.expungement_result.type_eligibility.reason == 'Ineligible under 137.225(5)'
 
     def test_convicted_infraction_is_not_type_eligible(self):
-        charge = ChargeFactory.create(statute='811135', level='Infraction Class B', disposition=self.convicted)
+        charge = ChargeFactory.create(statute='811.135', level='Infraction Class B', disposition=self.convicted)
```
KentShikama: We should probably keep the format of the statute like before as it seems we can get both patterns in production.

src/backend/tests/models/charge_types/test_level_800_traffic_crime.py
```diff
@@ -125,11 +131,12 @@ def test_violation_level_charge_skips_analysis(self):
         assert charge.skip_analysis() is True
 
     def test_level_800_infractions_skip_analysis(self):
-        charge = ChargeFactory.create(statute='811135', level='Infraction Class B', disposition=self.convicted)
+        charge = ChargeFactory.create(statute='811.135', level='Infraction Class B', disposition=self.convicted)
 
         assert charge.skip_analysis() is True
 
 
+
```
KentShikama: We forgot to change the statute for test cases below (i.e. skip analysis). As you mentioned while we were pair programming, the meaning of these tests have now changed if we don't switch away from chapter 813.

wittejm:
@KentShikama Cool thanks, I made all of the recommended fixes. 

I just made up a new statute number for the misdemeanor traffic charge. The test case doesn't care what it is, and I still looked for a real statute that would work but I couldn't find any that qualify that aren't DUII.

PR #608, opened 2019-12-07
Convert Record into a data class

KentShikama:
See #617


NickSchimek:
@KentShikama Can you please explain why this is being done and reference the issue?

Issue #601, opened 2019-12-04
Classify a disposition more accurately


wittejm:
For the purposes of expungement, a charge can be ruled as "convicted", "dismissed", or "no complaint".

Legally speaking, "dismissed" and "acquitted" have separate definitions. But for expungement they are considered the same and we call both of those dismissals.

In the OECI database, multiple different `ruling` strings on the Disposition can be interpreted as convictions:

- "Convicted"
- "Finding - Guilty"
- "Reduced to Lesser Charge"

And multiple different `ruling` strings can be interpreted as dismissals:

- "Dismissed"
- "Finding - Not Guilty"

Also, the "Diverted" ruling is a weird case that is being addressed in #592 

Currently the Charge class has the `convicted()` method which returns True only if the `disposition.ruling.lower()` value begins with the substring "convicted". This should be updated to classify all of the listed rulings correctly. The `Charge.acquittal` should be updated also.

This addresses two bugs reported from the logic bug-reports page (identifying record information is omitted here)

Also I suggest renaming the occurrences of `acquittal` and `dismissal` in the project to to a single term since their legal definitions are distinct but we treat them as equivalent in the expunger.

wittejm:
More types that can come from OECI: "converted" and "conversion" should both be interpreted as convictions

wittejm:
This field be the one that is read by the frontend to determine convicted/not convicted status (related to PR #622 ), rather than redundnantly parse the `ruling` string

PR #609, opened 2019-12-07
Convert Disposition into a data class

KentShikama:
Note we set `eq=False` and keep the old `__init__` method to preserve previous behavior. In other words, this is primarily for `__repr__`, the types, and for the consistency with every other class under models/.

Also see #617

### DIFF
--- a/src/backend/expungeservice/models/disposition.py
+++ b/src/backend/expungeservice/models/disposition.py
@@ -1,9 +1,14 @@
-from datetime import datetime
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import Optional
 
 
+@dataclass(eq=False)
 class Disposition:
+    date: Optional[date]
+    ruling: str
 
-    def __init__(self, date=None, ruling=None):
+    def __init__(self, date: Optional[str] = None, ruling: str = ""):
         self.date = self.__set_date(date)
         self.ruling = ruling
 


src/backend/expungeservice/models/disposition.py
```diff
@@ -1,9 +1,14 @@
-from datetime import datetime
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import Optional
 
 
+@dataclass(eq=False)
 class Disposition:
+    date: Optional[date]
```
NickSchimek: I don't believe the date is optional. If the disposition is instantiated then there needs to be a date.

src/backend/expungeservice/models/disposition.py
```diff
@@ -1,9 +1,14 @@
-from datetime import datetime
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import Optional
 
 
+@dataclass(eq=False)
```
NickSchimek: Same question here as the other PR. Why label this as a data class, but treat it as a regular class?

NickSchimek:
Can you please explain why this change is being done and reference an issue?

src/backend/expungeservice/models/disposition.py
KentShikama: I am just preserving current behavior here. By nature of the previous default argument being `date=None`, the date can be optional. It doesn't matter that the default argument is always overridden.

src/backend/expungeservice/models/disposition.py
KentShikama: Hmm maybe we have a conceptual difference of what a data class is. A data class is still just a regular Python class. From a functional perspective, you can view the dataclass class wrapper as a pattern that adds `__repr__`, `__init__` (including removing boilerplate), and `__eq__` to the class. Optionally it can include `__hash__` and ordering methods. I'll also say that tells the code reader that this class is primarily about holding data.

src/backend/expungeservice/models/disposition.py
KentShikama: Also we assume this on line 18 of this file.

wittejm:
@KentShikama @NickSchimek can we merge this PR? I'd like to work on issue #601 which will touch the same file.

wittejm:
Woops misclick.

src/backend/expungeservice/models/disposition.py
NickSchimek: Thanks for pointing that out. I let my knowledge of the code base interfere with reading the code
.

NickSchimek: @KentShikama The default parameters can be removed. I think we can either put that in this PR or create a new issue. I will leave that up to you on how you want to do it. See PR #357 for more information on when the optional parameters were removed.

src/backend/expungeservice/models/disposition.py
NickSchimek: Yes, we have a different conceptual difference. I re-read PEP 557, and I now understand why they chose to do it this way rather than the way that I would expect. For me a dataclass would be or less like a tuple. 

> Such a class is called a Data Class, but there's really nothing special about the class: the decorator adds generated methods to the class and returns the same class it was given.

KentShikama:
@wittejm It requires a bit more git work but you can always work on top of this branch.

src/backend/expungeservice/models/disposition.py
KentShikama: Hmm, so to be clear, are you ok with this `@dataclass(eq=False)` change now or not?

wittejm:
That certainly is a thing I could do.

src/backend/expungeservice/models/disposition.py
KentShikama: I've created https://github.com/codeforpdx/recordexpungPDX/issues/625 as this is not as trivial of a change as I first thought (see the issue as to why).

NickSchimek: Thanks @KentShikama I thought there might still be tests that use the optional params.

src/backend/expungeservice/models/disposition.py
NickSchimek: Yes

NickSchimek:
> @KentShikama @NickSchimek can we merge this PR? I'd like to work on issue #601 which will touch the same file.

yes I was waiting to see if Kent was going to remove default params or create new issue.

src/backend/expungeservice/models/disposition.py
NickSchimek: I still don't like it, but I'm OK with it since it seems to be the python way.

src/backend/expungeservice/models/disposition.py
NickSchimek: Thanks for being patient and helping me learn!

src/backend/expungeservice/models/disposition.py
NickSchimek: Thanks for being patient and helping me learn!

KentShikama: I've learned a bunch myself and I'm sure these conversations will be useful to anyone working with this repo in the future.

PR #610, opened 2019-12-07
Add generative models for serialization tests and mock data for `dev_mock_oeci_up`

KentShikama:
Relative to #577 #608 #609

### DIFF
--- a/src/backend/expungeservice/models/generator/__init__.py
+++ b/src/backend/expungeservice/models/generator/__init__.py
@@ -2,6 +2,7 @@ from _weakref import ref
 
 from hypothesis._strategies import none, composite
 
+from expungeservice.expunger.expunger import Expunger
 from expungeservice.models.case import Case
 from expungeservice.models.record import Record
 from hypothesis.strategies import builds, just, lists, one_of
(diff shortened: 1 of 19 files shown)


wittejm:
Oh yea and if the serializer test fails it would be awesome to print the data object that broke it (I don't think the pytest stack trace will do this?) because it could be difficult to reproduce.

src/backend/expungeservice/models/generator/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+from expungeservice.models.charge_types.list_b import ListB
+from expungeservice.models.charge_types.marijuana_ineligible import MarijuanaIneligible
+from expungeservice.models.charge_types.misdemeanor import Misdemeanor
+from expungeservice.models.charge_types.non_traffic_violation import NonTrafficViolation
+from expungeservice.models.charge_types.parking_ticket import ParkingTicket
+from expungeservice.models.charge_types.person_crime import PersonCrime
+from expungeservice.models.charge_types.schedule_1_p_c_s import Schedule1PCS
+from expungeservice.models.charge_types.unclassified_charge import UnclassifiedCharge
+
+charge_types = [JuvenileCharge, FelonyClassA, FelonyClassB, FelonyClassC, Level800TrafficCrime, ListB, MarijuanaIneligible, Misdemeanor, NonTrafficViolation, ParkingTicket, PersonCrime, Schedule1PCS, UnclassifiedCharge]
```
NickSchimek: This is a dependency. If we ever add a new charge type, like the DUII that was just added, then we need to add it here? Perhaps we can come up with a way that this dynamically updates?

KentShikama: Yeah it does make it such that we would need to add a new charge types here. I did a quick attempt at making it dynamically update initially but I didn't really like the result; I'll think about it a bit more while I address @wittejm 's comments above. If someone forgets to add a charge_type here, it simply means that charge_type won't be tested in terms of serialization, so the consequences are pretty minor imo.

KentShikama:
Superseded by https://github.com/codeforpdx/recordexpungPDX/pull/628

PR #613, opened 2019-12-08
Display loading spinner while loading a second search

### LINKED ISSUE #574: Display spinner when searching
## Description

The spinner shows when doing a search, however if a subsequent search is done the results page is not cleared, and it gives the appearance of being broken.

### To do:

When a subsequent search is performed the following should occur:
 - [ ] Clear all results on the page from previous search
 - [ ] Display spinner while waiting response from backend

wittejm:
The loading spinner component is only shown if the redux `records` object is empty, which is checked here:

 https://github.com/codeforpdx/recordexpungPDX/blob/master/src/frontend/src/containers/AllRecords.tsx#L34 ~~line 34.~~
 
We could change that logic, but this is a smaller and simpler change code-wise, so I'm running with it. 

closes #574

### DIFF
--- a/src/frontend/src/redux/records/reducer.ts
+++ b/src/frontend/src/redux/records/reducer.ts
@@ -25,7 +25,7 @@ export function recordsReducer(
       // When an API call is made for records, loading state is toggled to true.
       // while loading state is true, a spinner is rendered on the screen. If loading
       // state is false, and no records were fetched, no search results found will be displayed.
-      return { ...state, loading: true };
+      return { ...state, records: {}, loading: true };
     case CLEAR_SEARCH_RECORDS:
       return { ...state, records: {}, loading: false };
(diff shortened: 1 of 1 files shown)


KentShikama:
Btw you can link directly to the line by clicking on it in Github (e.g. https://github.com/codeforpdx/recordexpungPDX/blob/master/src/frontend/src/containers/AllRecords.tsx#L34). How does this pass https://github.com/codeforpdx/recordexpungPDX/blob/master/src/frontend/src/containers/AllRecords.tsx#L35 this check though?

wittejm:
> Btw you can link directly

Ah good to know.

> How does this pass 

What do you mean? The expression short-circuits because `{}.cases` is `undefined` and evaluates to `false`. Or are you talking about something else?

PR #614, opened 2019-12-08
Remove time portion on dates in search results data 

### LINKED ISSUE #573: Remove time portion on dates
## Description

Remove the time portion whenever a date is displayed.

Screenshot 2019-11-20 08 58 14

This should not include a time, only the date needs to be included.

### To do:

- [ ] Normalize all dates to display the month day, year: `Sep 19, 2026`

wittejm:
Closes #573.

This is a backend change because the data serializer is the source of the extended datetime format. 

An alternative to the expression I'm using here is `o.strftime("%b %d, %Y")` but this pads the day of month with a leading 0 which is ugly. The date formatter doesn't provide a directive to show day of month without the leading 0.

### DIFF
--- a/src/backend/expungeservice/serializer/serializer.py
+++ b/src/backend/expungeservice/serializer/serializer.py
@@ -1,6 +1,7 @@
 import flask
 import expungeservice
 from expungeservice.models.expungement_result import EligibilityStatus
+from datetime import date
 
 
 class ExpungeModelEncoder(flask.json.JSONEncoder):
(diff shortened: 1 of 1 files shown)


NickSchimek:
I guess I could pull this down and spin it up to see the results. It's convenient for other's to add a picture if you can.

Screen Shot 2019-12-10 at 9 28 35 PM

To remove the padding on a digit you can add a minus sign
```
>>> datetime.datetime.now().strftime("%b %d, %Y")
'Dec 08, 2019'
>>> datetime.datetime.now().strftime("%b %-d, %Y")
'Dec 8, 2019'
```

PR #622, opened 2019-12-10
Show correct date based on conviction vs arrest

### LINKED ISSUE #621: Date of a convicted charge should show conviction date
This is a correction related to #605 . The [mockup](http://dev.huntermarcks.net/search/) shows the date of conviction if the charge resulted in conviction, whereas it shows the date of arrest for other charges. This makes sense because eligibility dates are computed based on these dates, rather than always depending on one or the other.

wittejm:
I'm not crazy about the approach I took here because it looks kind of clunky, but I couldn't figure out a more "reacty" style. PTAL

Closes #621

### DIFF
--- a/src/frontend/src/components/Charge/index.tsx
+++ b/src/frontend/src/components/Charge/index.tsx
@@ -18,18 +18,27 @@ export default class Charge extends React.Component<Props> {
       expungement_result
     } = this.props.charge;
 
-    const knownDisposition = (disposition: any) => (
-      <>
+
+    const knownDisposition = (disposition: any, date: any) => {
+      let dispositionEvent;
+      if (disposition.ruling.toLowerCase().includes("convicted")) {
+        dispositionEvent = "Convicted: " ;
+        date = disposition.date;
+      } else {
+        dispositionEvent = "Arrested: " ;
+      }
+
+      return <>
         <li className="mb2">
           <span className="fw7">Disposition: </span> {disposition.ruling}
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/Charge/index.tsx
```diff
@@ -18,26 +18,35 @@ export default class Charge extends React.Component<Props> {
       expungement_result
     } = this.props.charge;
 
-    const knownDisposition = (disposition: any) => (
-      <>
+
+    const knownDisposition = (disposition: any, date: any) => {
+      let dispositionEvent;
+      if (disposition.ruling.toLowerCase().includes("convicted")) {
```
NickSchimek: I assume this is based on the logic in the backend. If the backend is ever changed then we have to remember to change the logic here. Do you think we should update the JSON response so that the charge object will have a field that states wether it is convicted or arrested so that we can remove this logic?

KentShikama: Why don't we just display the disposition ruling as is (Convicted, Acquitted, Dismissed, etc)?

wittejm: @NickSchimek Yes, I like the idea of having this be simpler to read from the JSON response object. It should be part of #601 . I can add a note there.

@KentShikama The date only comes from the disposition if it was a conviction. If the disposition was an acquittal etc., the arrest date is what matters instead

Issue #620, opened 2019-12-10
Unclassified charges should always have type eligibility NEEDS_MORE_ANALYSIS


wittejm:
An acquitted unclassified charge is inaccurately marked as eligible. 

### This task is done when:
 - Unclassified charge has type eligibility  `NEEDS_MORE_ANALYSIS` even when it is acquitted.
 - Test case updated

wittejm:
closed with #623

PR #623, opened 2019-12-10
Unclassified charge always needs more analysis

wittejm:
closes issue #620 

Reason: Certain known charge types are ineligible or need more analysis even if they are not convicted. A charge can be unclassified as a result of OECI data error and might be ineligible, so it should be marked as needing more analysis.


wittejm:
@NickSchimek Oh yes good idea.  I added the reason to this PR and to the issue.

PR #626, opened 2019-12-11
Classify additional statute as PCS

### LINKED ISSUE #624: Add a repealed statute that qualifies as a PCS charge
The statute [475.992 ](https://www.oregonlaws.org/ors/475.992) used to describe a PCS charge but has been repealed/renamed. Since it still exists in OECI records applying to PCS, we should classify it accordingly.

The OECI records actually list, for example, 4759924A and 4759924B under this statute, which refer to its subsections. So when we check to see if a charge matches the statute we could examine only the first 6 digits rather than doing a string equality 

### This task is done when:
 - update [charge_classifier.py](https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/charge_classifier.py)
 - a new test for this use case

wittejm:
closes #624

### DIFF
--- a/src/backend/expungeservice/models/charge_classifier.py
+++ b/src/backend/expungeservice/models/charge_classifier.py
@@ -109,7 +109,7 @@ class ChargeClassifier:
 
     @staticmethod
     def _schedule_1_pcs(section):
-        if section in ['475854', '475874', '475884', '475894']:
+        if section in ['475854', '475874', '475884', '475894'] or section[:6] == '475992':
             return Schedule1PCS
 
     @staticmethod
--- a/src/backend/expungeservice/models/charge_classifier.py
+++ b/src/backend/expungeservice/models/charge_classifier.py
@@ -109,7 +109,7 @@ class ChargeClassifier:
 
     @staticmethod
     def _schedule_1_pcs(section):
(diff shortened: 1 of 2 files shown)


src/backend/expungeservice/models/charge_classifier.py
```diff
@@ -109,7 +109,7 @@ def _parking_ticket(statute, chapter):
 
     @staticmethod
     def _schedule_1_pcs(section):
-        if section in ['475854', '475874', '475884', '475894']:
+        if section in ['475854', '475874', '475884', '475894'] or section[:6] == '475992':
```
KentShikama:
```suggestion
        if section[:6] in ['475854', '475874', '475884', '475894', '475992']
```

wittejm: I considered that but I wanted to be super cautious about changing the legal logic here. Your suggestion would cause something like `4758544A` to be classified as PCS too. That's probably fine though, because the subsequent digits are subsections.

KentShikama: If you're worried, I would add a comment that you're not sure. Otherwise it looks like the subsections actually matter for the previous 4.

PR #628, opened 2019-12-11
Add generative models for serialization tests and mock data for `dev_mock_oeci_up`

KentShikama:
Note that https://github.com/ automatically closed https://github.com/codeforpdx/recordexpungPDX/pull/610 when I deleted the develop branch the older PR was based on. This PR is now based on the master branch and is no longer relative to any other PRs.

I am currently addressing @wittejm 's comments from the earlier PR.


KentShikama:
@wittejm @NickSchimek PTAL. Note that the generated results still may not be realistic but I'll save that for a separate PR (see https://github.com/codeforpdx/recordexpungPDX/issues/634). The PR as-is is useful for ensuring the integrity of the serializer.

> if the serializer test fails it would be awesome to print the data object that broke it

Indeed, `pytest` does do this. The output will look like the following:

```
Falsifying example: test_round_trip_various_records(
    record=Record(cases=[Case(name='', birth_year=None, case_number='', citation_number='', location='', date=datetime.date(2000, 1, 1), violation_type='', current_status='', charges=[Duii(name='', statute='', level='', date=datetime.date(2000, 1, 1), disposition=None, expungement_result=ExpungementResult(type_eligibility=TypeEligibility(status=<EligibilityStatus.NEEDS_MORE_ANALYSIS: 'Needs more analysis'>, reason='Further Analysis Needed - No disposition'), time_eligibility=None), _chapter=None, _section='', _case=<weakref at 0x7f2f5fc04138; to 'Case' at 0x7f2f5fc457f0>)], case_detail_link='', _Case__balance_due_in_cents=0, _Case__probation_revoked=False)], errors=[]),
)
```

src/backend/expungeservice/models/helpers/charge_creator.py
```diff
@@ -4,7 +4,7 @@
 
 from dacite import from_dict
 
-from expungeservice.models.charge_classifier import ChargeClassifier
+from expungeservice.models.helpers.charge_classifier import ChargeClassifier
```
NickSchimek: I'm not quite sure if I agree entirely with this name change, but that's not really important. What I think is important is that the `generator.py`, as far as I can tell, is only used for tests. Shouldn't the `generator.py` file be in the test directory?

KentShikama: Feel free to suggest a different name. It is also used for the [mock backend](https://github.com/codeforpdx/recordexpungPDX/pull/628/files#diff-d84ab9e795dfeb073aa4f52331a51b6eR5).

NickSchimek: I'm honestly not familiar with the mock backend. Personally, I feel any code that does not contribute to the production app should be in its own directory, such as test, or whatever directory name is suitable. 

As far as the naming goes, when I see a directory named `helper` then I tend to think the files in the helper classes are to dry up code between different modules. This is a factory class, and it does help/facilitate the creation of a charge class. Although putting it in a directory with a class that is not used as part of the main app doesn't seem quite right.

Anyway, this is probably not worth talking about much. We can discuss more in person. Feel free to merge. I didn't see anything else that stood out on me. Thanks for the PR 💫

PR #637, opened 2019-12-12
DispositionStatus enum and related refactor

### LINKED ISSUE #601: Classify a disposition more accurately

wittejm:
This classifies a variety of disposition rulings into a standardized set of types. 

I still want to keep poking at this and see if the refactor of the  `acquitted` method is a good idea or could use some more work.

I could also add a test that various ruling strings get classified correctly. 

The check that I added for `DispositionStatus.UNKNOWN` (or call it `UNRECOGNIZED`?) is also something that I think should be added to type eligibility for every charge type. That should probably mean checking it once in the `Charge` class.
 
Closes #601

### DIFF
--- a/src/backend/expungeservice/models/charge_types/duii.py
+++ b/src/backend/expungeservice/models/charge_types/duii.py
@@ -31,10 +31,10 @@ class Duii(Charge):
                     reason="137.225(8)(b) - Diverted DUII charges are ineligible")
 
             cases = {
-                DispositionStatus.CONVICTED: convicted_type_eligibility
-                DispositionStatus.DISMISSED: dismissed_type_eligibility
-                DispositionStatus.NO_COMPLAINT: dismissed_type_eligibility
-                DispositionStatus.DIVERTED: diverted_type_eligibility
+                DispositionStatus.CONVICTED: convicted_type_eligibility,
+                DispositionStatus.DISMISSED: dismissed_type_eligibility,
+                DispositionStatus.NO_COMPLAINT: dismissed_type_eligibility,
+                DispositionStatus.DIVERTED: diverted_type_eligibility,
                 DispositionStatus.UNKNOWN: unknown_type_eligibility
             }
             assert len(cases) == len(DispositionStatus)
--- a/src/backend/expungeservice/models/charge_types/duii.py
+++ b/src/backend/expungeservice/models/charge_types/duii.py
@@ -1,5 +1,6 @@
 from expungeservice.models.charge import Charge
 from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+from expungeservice.models.disposition import DispositionStatus
 
 
 class Duii(Charge):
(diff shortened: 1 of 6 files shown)


src/backend/expungeservice/models/charge_types/charge.py
```diff
@@ -32,14 +32,14 @@ def case(self):
         return self._case
 
     def acquitted(self):
-        return self.disposition and self.disposition.ruling[0:9].lower() != 'convicted'
+        return self.disposition and (self.disposition.status == DispositionStatus.DISMISSED or self.disposition.status == DispositionStatus.NO_COMPLAINT)
```
KentShikama: We should add a TODO to change the method name acquitted to dismissed.

src/backend/expungeservice/models/disposition.py
```diff
@@ -1,18 +1,60 @@
...
 @dataclass(eq=False)
 class Disposition:
     date: Optional[date]
     ruling: str
+    status: DispositionStatus
 
     def __init__(self, date: Optional[str] = None, ruling: str = ""):
         self.date = self.__set_date(date)
         self.ruling = ruling
+        self.status = self.__set_status()
```
KentShikama: Use `__post_init__` to set `self.status` as we eventually just want to remove this `__init__` field. See https://docs.python.org/3/library/dataclasses.html#post-init-processing.

src/backend/expungeservice/models/disposition.py
```diff
@@ -1,18 +1,60 @@
...
         return None
+
+    def __set_status(self):
+        ruling = self.ruling.lower()
+        if any([conviction_ruling in ruling for conviction_ruling in [
+                "convicted",
+                "conviction",
+                "finding - guilty",
+                "conversion",
+                "converted"]]):
```
KentShikama: I would prefer if we extracted this list into a variable `conviction_rulings` so it only takes up one line without going over line character limit. A similar thing with the below.

src/backend/expungeservice/models/disposition.py
```diff
@@ -1,18 +1,60 @@
...
+
+            return  DispositionStatus.DIVERTED
+
+        elif "no complaint" in ruling:
+
+            return  DispositionStatus.NO_COMPLAINT
+
+        else:
+
+            return  DispositionStatus.UNKNOWN
```
KentShikama: I prefer without the unnecessary line spacing.

src/backend/expungeservice/models/charge_types/duii.py
```diff
@@ -9,17 +10,18 @@ def _default_type_eligibility(self):
...
             """
-            TODO: other dispositions than "convicted" are possible and are not handled here.
-            See issue: https://github.com/codeforpdx/recordexpungPDX/issues/601
-
             TODO: reason can be expanded to "137.225(7)(a) - Traffic offenses are ineligible"
             """
-            return TypeEligibility(EligibilityStatus.INELIGIBLE, reason='Ineligible under 137.225(7)(a)')
+            return TypeEligibility(EligibilityStatus.INELIGIBLE, reason='137.225(7)(a) - Traffic offenses are ineligible')
+
+        elif self.acquitted():
```
KentShikama: I think this case requires more of a comment as the naive reader may otherwise rebut that "wait a minute, isn't the diverted case handled above"?

src/backend/expungeservice/models/charge_types/duii.py
```diff
@@ -9,17 +10,18 @@ def _default_type_eligibility(self):
...
+            return TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Further Analysis Needed - Dismissed charge may have been Diverted')
+
+        else:
+            return TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Further Analysis Needed - Unrecognized ruling')
```
KentShikama: For now, it looks like the only reason we are `__convicted`/`acquitted` in `charge.py` is because of changes in these files. This may have side-effects on the relationship between other charge subclasses and the expunger (and in fact most likely they are "good"/bug-fixing side-effects). However, I think we need to be more explicit about those side-effects as any future reader of this PR may think that `__convicted`/`acquitted` changes are solely for this `duii.py` file.

Since we have "full coverage" of the DispositionStatus Enum here, we should just do a "switch/case" imitation like the following:

```python
if self.disposition:
    dismissed_type_eligibility = TypeEligibility(EligibilityStatus.NEEDS_MORE_ANALYSIS, reason='Further Analysis Needed - Dismissed charge may have been Diverted')
    unknown_type_eligibility = ...
    cases = {
        DispositionStatus.CONVICTED: TypeEligibility(EligibilityStatus.INELIGIBLE, reason='137.225(7)(a) - Traffic offenses are ineligible')
        DispositionStatus.DISMISSED: dismissed_type_eligibility
        DispositionStatus.NO_COMPLAINT: dismissed_type_eligibility
        DispositionStatus.DIVERTED:  ...
        DispositionStatus.UNKNOWN: unknown_type_eligibility
    }
    assert len(cases) == len(DispositionStatus) # So that if someone adds something to DispositionStatus, it won't automatically go to the Unknown case.
    return cases.get(self.disposition.status, unknown_type_eligibility) 
else:
     ...
```

And then I suggest a separate PR for the `__convicted`/`acquitted` changes that adds comments how it affects the expunger (or how it doesn't).

PR #646, opened 2019-12-14
Skip <td> tags and non-breaking whitespace in <th> tags

### LINKED ISSUE #647: Case Parser crashing due to non-standard charge info table
## Description
Issue found by @ForrestLonganecker and reported by @wittejm 

Reported that the parser was crashing. Upon further inspection the case parser was crashing due to a non-standard charge info table. 

The table has extra <th> tags and empty <td> cells.

### Todo:
 - [ ] Update the state parser for the charge info table to skip <th> tags and <td> tags containing whitespace

Will create another ticket to import the logging statements from the CLI to the parser so that further issues can be caught and logged.

NickSchimek:
## Description
This fixes an issue with the case parser when a table contains extra header tags and rows with empty data tags. Issue reported by @ForrestLonganecker 

Fixes #647


NickSchimek:
> How about if this checks for anything that's purely whitespace?

This is the HTML for the blank row `<tr><td>&nbsp;</td></tr>` and it is read by python as `\xa0`. I currently don't see the need to add that check, because the line `<tr><td>&nbsp;</td></tr>` is probably automatically added by their system when charges are entered for a new name. Feel free to add more commits to this PR if you feel the extra check is needed.

PR #648, opened 2019-12-14
Convicted/acquitted methods on Charge use disposition status

wittejm:
This incorporates the new DispositionStatus enum from #637 to evaluate `acquitted()` and `convicted()` for a charge.

Here is where `acquitted()` gets called in time analysis:
https://github.com/codeforpdx/recordexpungPDX/blob/67c4bcc745cb5d4cbbb0c1179a36bf5b7617b5a3/src/backend/expungeservice/expunger/expunger.py#L86

Should we skip UNKNOWN dispositions in time analysis, like we do dispositionless charges? If so, then by the time we reach that `acquitted()` call we will have already purged UNKNOWN dispositions:

Since the new test case ensures everything else is either convicted or acquitted, the `if / else` in the expunger here works the same way as checking `acquitted()` and `convicted()` separately. 

So if that's the case, `convicted()` does nothing for time analysis. Right now I only renamed it from `__convicted()` so that it would be inherited on subclasses and be exposed in the test case (and that's not strictly necessary) . 

That seems fine as a place to leave it. We can also imagine incorporating this more after #645, because this is another way to check missing/unknown dispositions. Do something like:

```
def _categorize_charges(self):
        for charge in self.charges:
            if charge.acquitted():
                self.acquittals.append(charge)
            elif charge.convicted():
                self.convictions.append(charge)
            else:
                ... handle and/or skip charge ...
```

and some refactoring because the expunger already removed dispositionless charges. It's just that now we can handle dispositionless and disposition status unknown at the same time because they both yield  `(acquitted()==False` and `convicted==False)`.

### DIFF
--- a/src/backend/tests/models/test_disposition_status.py
+++ b/src/backend/tests/models/test_disposition_status.py
@@ -1,6 +1,7 @@
 from datetime import date
 
 from expungeservice.models.disposition import Disposition, DispositionStatus
+from expungeservice.models.expungement_result import EligibilityStatus
 
 from tests.factories.charge_factory import ChargeFactory
 
@@ -46,7 +47,15 @@ def test_all_disposition_statuses_are_either_convicted_or_acquitted():
             assert charge.convicted() or charge.acquitted()
 
 def test_dispositionless_charge_is_not_convicted_nor_acquitted():
-
     charge = ChargeFactory.create()
     assert not charge.convicted()
     assert not charge.acquitted()
+    assert charge.expungement_result.type_eligibility.status is EligibilityStatus.NEEDS_MORE_ANALYSIS
+    assert charge.expungement_result.type_eligibility.reason == "Disposition not found. Needs further analysis"
+
+def test_charge_with_unknown_disposition_eligibility():
+    charge = ChargeFactory.create(disposition=["What am I", "1/1/0001"])
+    assert not charge.convicted()
+    assert not charge.acquitted()
+    assert charge.expungement_result.type_eligibility.status is EligibilityStatus.NEEDS_MORE_ANALYSIS
+    assert charge.expungement_result.type_eligibility.reason == "Disposition not recognized. Needs further analysis"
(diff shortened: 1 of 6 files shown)


src/backend/tests/models/test_disposition_status.py
```diff
@@ -0,0 +1,52 @@
...
+    for status in DispositionStatus:
+
+        #Use the status.value to create the disposition,
+        #which happens to always be a valid string for that dispo status.
+        charge.disposition = Disposition(today, status.value)
+
+        if status == DispositionStatus.UNKNOWN:
+            assert not charge.convicted()
+            assert not charge.acquitted()
+            continue
```
KentShikama: I would prefer `else:`

KentShikama:
> Diverted ruling is being considered a conviction here atm because in DUII type eligibility it's treated like one. Am checking for confirmation on this.

Here is my reasoning.

DUII currently has a type eligibility of either NEEDS_MORE_ANALYSIS or INELIGIBLE so it will never be expungeable itself (based on the existing codebase).

So then we need to consider how a DUII affects the eligibility of other charges.

https://www.oregonlaws.org/ors/137.225 (7)(b) only excludes motor vehicle violations and a single violation. https://www.oregonlaws.org/ors/813.010 (4) says that DUII a Class A misdemeanor, which by definition is not a violation. Is it possible that it may be converted down to a violation somehow? https://www.oregonlaws.org/ors/137.225 (8)(b) says that diverted DUIIs can't be treated as a non-conviction so it would make sense to include it under (7)(b) as a conviction. Is that indeed the case?

I guess we should ask Michael.

KentShikama:
Answer from Michael:

- Yes - it is definitely possible that a case is marked as closed if even if it is No-Complainted
- A diverted DUI dismissal counts as an arrest, but not a conviction
- It is not possible that a DUI would be reduced to a violation

KentShikama:
So to translate that in the context of our codebase, a diverted DUI should be grouped with the non-convictions.

wittejm:
This PR is finished but still needs unknown Disposition to be handled in the expunger, related to #645, before it should be merged.

PR #649, opened 2019-12-14
Run time analyzer with all closed cases

KentShikama:
~~Relative to #645~~

As suggested by @wittejm earlier today, we should always run the time analyzer regardless of whether there are open cases. The open cases are treated as if they don't exist. Whether the analysis is shown depends on how the frontend wants to handle it; the frontend can look at `record.errors` to see whether there were any open cases.

Related to https://github.com/codeforpdx/recordexpungPDX/issues/586

### DIFF
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -1,9 +1,8 @@
-from more_itertools import padnone, take
 from typing import Set
+from more_itertools import padnone, take, flatten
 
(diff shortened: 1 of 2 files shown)


KentShikama:
We probably want some minimal warning message on the front-end in the situation there are open cases. Otherwise this change will just silently show through.

KentShikama:
Ready now thanks to https://github.com/codeforpdx/recordexpungPDX/pull/685

PR #653, opened 2019-12-16
Class B felony ineligible if any subsequent arrests or convictions

### LINKED ISSUE #651: B felonies are ineligible if there are any subsequent arrests/convictions
The currently expunger logic incorrectly applies time eligibility to Class B felonies.
The rule is: If a conviction is a B felony and doesn' qualify for expungement under any other rule, it is eligible only if 20 years have passed from the date of conviction AND the record has no other arrests or convictions following the B felony conviction

The logic fix should happen here in the time analyzer:

https://github.com/codeforpdx/recordexpungPDX/blob/67c4bcc745cb5d4cbbb0c1179a36bf5b7617b5a3/src/backend/expungeservice/expunger/analyzers/time_analyzer.py#L62

Comparing the date of the B felony conviction against the dates of the MRC and MRD is insufficient to meet this condition, because those only get assigned they are less than 10 years or 3 years old respectively. For example if the most recent conviction is 15 years old it doesn't appear as the MRC because it doesn't block other expungements. But it will still block B felony expungement, so it needs to be checked for in this analysis.

The relevant statute (to be included for ineligibility reason) is 137.225(5)(a)(A)
### This task is done when:
 - The method ` _evaluate_class_b_felonies(expunger)` is updated to reflect this rule and new test cases added.
 -

wittejm:
Class B felony are only eligible if they're at least 20 years old, and only if the record has no subsequent charges.

The test cases here check that it doesn't affect the similar but different cases: If the B felony was acquitted, if there was a prior conviction, etc. 

This includes a slight refactor of the two preexisting class B felony tests, to clean up some repeated code. 

Closes #651


src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -60,11 +62,37 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
+                    #if it's an aquittal, the arrest date is what matters
+                    if other_charge.acquitted():
+                        date_of_charge = other_charge.date
+                    #if it's a conviction, the conviction date is what matters
+                    else:
+                        date_of_charge = other_charge.disposition.date
+
+                    if date_of_charge > b_felony.disposition.date:
+                        has_subsequent_charge = True
+                        break
```
KentShikama: We can extract a method here: `def calculate_has_subsequent_charge(...) -> bool`

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -60,11 +62,37 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
-                                           expunger.most_recent_charge.disposition.date + relativedelta(
-                                               years=TimeAnalyzer.TWENTY_YEARS))
+        for b_felony in expunger.class_b_felonies:
+
+            # if the B felony is acquitted, it is affected the other time eligibility rules,
+            # so leave it untouched here.
+            # if it's convicted, this restriction overrides any of those rules.
+            if not b_felony.acquitted():
+                has_subsequent_charge = False
+                for other_charge in expunger.charges:
```
KentShikama: Doesn't `expunger.charges` include `expunger.class_b_felonies`? It looks like this fails the `date_of_charge > b_felony.disposition.date` check (as desired) because `date_of_charge == b_felony.disposition.date` when comparing the same charge, but it doesn't feel intentional.

A couple options include renaming `other_charge` to `charge`, removing `b_felony` from the charges like
`other_charges = [charge for charge in expunger.charges if charge != b_felony]`, or putting an explicit check for when `other_charge == b_felony`.

NickSchimek:
Thanks Jordan! Please give me a couple days to review this.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: Yes there is a self-comparison here, but a class B felony should only become ineligible if the compared date is strictly greater than that of the B felony. The self-equality here makes this expression false.  I suppose that explicitly pulling out the self-comparison could improve readability.

KentShikama: Indeed, that is why I said "as desired".

wittejm: Ah, woops, yes good. I'm fine with changing it as a matter of style/readability.

wittejm: It was intentional though, slash I initially decided it wasn't worth acting on :)

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -58,13 +60,39 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
+            # If the B felony is acquitted, it is affected by the other time eligibility rules,
+            # so leave it untouched here.
+            # If it's convicted, this restriction overrides any of those rules.
+            if not class_b_felony.acquitted():
+                other_charges = [ charge for charge in expunger.charges if charge != class_b_felony]
+                if TimeAnalyzer._calculate_has_subsequent_charge(class_b_felony.disposition.date, other_charges):
+                    type_eligibility = TypeEligibility(
+                        EligibilityStatus.INELIGIBLE,
+                        reason='137.225(5)(a)(A)(ii) - Class B felony can have no subsequent arrests or convictions')
+                    class_b_felony.expungement_result = ExpungementResult(type_eligibility=type_eligibility, time_eligibility=None)
```
KentShikama: This should still be `TimeEligibility`. Maybe we need to rename the classes but my understanding is that `TypeEligibility` is what can be determined with only the charge and `TimeEligibility` is what can be determined only in the context of the entire record.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -58,13 +60,39 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
-                charge.set_time_ineligible('Time-ineligible under 137.225(5)(a)(A)(i)',
-                                           expunger.most_recent_charge.disposition.date + relativedelta(
-                                               years=TimeAnalyzer.TWENTY_YEARS))
+        for class_b_felony in expunger.class_b_felonies:
+            # If the B felony is acquitted, it is affected by the other time eligibility rules,
+            # so leave it untouched here.
+            # If it's convicted, this restriction overrides any of those rules.
+            if not class_b_felony.acquitted():
+                other_charges = [ charge for charge in expunger.charges if charge != class_b_felony]
+                if TimeAnalyzer._calculate_has_subsequent_charge(class_b_felony.disposition.date, other_charges):
```
KentShikama: I would prefer if we passed in `_calculate_has_subsequent_charge(class_b_felony, expunger.charges)` as `other_charges` is only relevant in calculating whether there is a subsequent charge or not.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -58,13 +60,39 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
     def _evaluate_class_b_felonies(expunger):
-        if expunger.most_recent_charge and expunger.most_recent_charge.disposition.date > TimeAnalyzer.TWENTY_YEARS_AGO:
-            for charge in expunger.class_b_felonies:
-                charge.set_time_ineligible('Time-ineligible under 137.225(5)(a)(A)(i)',
-                                           expunger.most_recent_charge.disposition.date + relativedelta(
-                                               years=TimeAnalyzer.TWENTY_YEARS))
+        for class_b_felony in expunger.class_b_felonies:
+            # If the B felony is acquitted, it is affected by the other time eligibility rules,
+            # so leave it untouched here.
+            # If it's convicted, this restriction overrides any of those rules.
```
KentShikama: Maybe just say "If the class B felony is acquitted, then we have already handled its time eligibility."

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: That's true. I was thinking about how the frontend interprets the difference between the two: if a charge is TypeEligible but not TimeEligible, that means it will become eligible at some point in the future, which is not what happens here. This definitely feels a bit off, semantically, though.

KentShikama: You could use [`datetime.MAXYEAR`](https://docs.python.org/3/library/datetime.html#datetime.MAXYEAR).

wittejm: Sorry missed this comment earlier. Wouldn't that just result in the frontend displaying, "will be eligible in a billion years"? Ideally it should say Ineligible.

wittejm: Could do something like having time eligibiltiy `NEVER`. I agree though that we should keep the TypeEligibility and reason intact to stay clear to the user.

KentShikama: A slightly long but accurate name for the two fields are "Default type eligibility" and "Eligibility in context of record and date". For Class B Felonies with subsequent arrests, we can use the value of `None` for the `date_will_be_eligible`. Records that are eligible in the context of record and date should have the date they became eligible in the `date_will_be_eligible` instead of the current value of `None`. This way `None` always means never in this context, which I think is an acceptable use of this often abused value.

We should then have a separate serialized field for the largest display label that we can build up right before the serializer using the DefaultEligibility/EligibilityInRecordContext (TypeEligibility/TimeEligibility). [We were looking to push this logic from the frontend to the backend anyways](https://github.com/codeforpdx/recordexpungPDX/pull/557#issuecomment-564383275).

I think it is fine to do this in a separate PR.

src/backend/tests/expunger/test_time_analyzer.py
```diff
@@ -81,40 +81,6 @@ def test_less_than_three_year_rule_conviction(self):
...
-        self.expunger.class_b_felonies = [charge]
-        self.expunger.most_recent_charge = charge
-
-        self.expunger.charges = [charge]
-        TimeAnalyzer.evaluate(self.expunger)
-
-        assert charge.expungement_result.time_eligibility.status is EligibilityStatus.INELIGIBLE
-        assert charge.expungement_result.time_eligibility.reason == 'Time-ineligible under 137.225(5)(a)(A)(i)'
-        assert charge.expungement_result.time_eligibility.date_will_be_eligible == Time.TOMORROW
-
```
NickSchimek: Creating a new file to put these tests in doesn't fit with the current style of tests.
We're testing the time_analyzer, hence the tests should stay in the time_analyzer test file to conform to the style of the project layout.<- if this style is not followed then we could potentially end up with a test file for each method.

If the desire is to separate these tests into their own section then I would suggest creating a class. Examples of this can be found in this file.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -58,13 +60,37 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
+                    type_eligibility = TypeEligibility(
+                        EligibilityStatus.INELIGIBLE,
+                        reason='137.225(5)(a)(A)(ii) - Class B felony can have no subsequent arrests or convictions')
+                    class_b_felony.expungement_result = ExpungementResult(type_eligibility=type_eligibility, time_eligibility=None)
+
+                else:
+                    if class_b_felony.disposition.date > TimeAnalyzer.TWENTY_YEARS_AGO:
+                        class_b_felony.set_time_ineligible('137.225(5)(a)(A)(i) - Twenty years from class B felony conviction',
+                                                       class_b_felony.disposition.date + relativedelta(
+                                                           years=TimeAnalyzer.TWENTY_YEARS))
```
NickSchimek: It seems like this is missing an else statement. If it doesn't have subsequent charges and the disposition date is less than 20 years what do we do? Otherwise perhaps we can do `elif` here.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
NickSchimek: A class b felony can be type eligible; however, if there is a subsequent charge then it invalidates the eligibility forever. This is definitely out of place and confusing here. Have we asked Michael how he would like to display this scenario?

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -58,13 +60,37 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
     def _evaluate_class_b_felonies(expunger):
-        if expunger.most_recent_charge and expunger.most_recent_charge.disposition.date > TimeAnalyzer.TWENTY_YEARS_AGO:
-            for charge in expunger.class_b_felonies:
-                charge.set_time_ineligible('Time-ineligible under 137.225(5)(a)(A)(i)',
-                                           expunger.most_recent_charge.disposition.date + relativedelta(
-                                               years=TimeAnalyzer.TWENTY_YEARS))
+        for class_b_felony in expunger.class_b_felonies:
+            # If the class B felony is acquitted, then we have already handled its time eligibility.
+            # If it's convicted, this restriction overrides any of those rules.
+            if not class_b_felony.acquitted():
```
NickSchimek: It's usually recommended to use positive expressions if available. Perhaps we can say `if class_b_felony.convicted()` or if we also need to handle unknown then the expression can be if convicted or unknown.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -58,13 +60,37 @@ def _mark_all_acquittals_ineligible_using_mrd_date(expunger):
...
             charge.set_time_ineligible('Recommend sequential expungement', eligibility_date)
 
+    @staticmethod
+    def _calculate_has_subsequent_charge(class_b_felony, charges):
+        other_charges = [charge for charge in charges if charge != class_b_felony]
+        for other_charge in other_charges:
+            if other_charge.acquitted():
+                date_of_other_charge = other_charge.date
+            else:
+                date_of_other_charge = other_charge.disposition.date
```
NickSchimek: I believe this logic is used in a lot of places.
```python
            if other_charge.acquitted():
                date_of_other_charge = other_charge.date
            else:
                date_of_other_charge = other_charge.disposition.date
```

It's probably a good time to extract this logic into a method and move it to the charge class. Or we can create a refactoring ticket to do that so that we can clean up the other places this similar logic is used. I'm pretty sure its used elsewhere although I haven't verified. -> even if its not used in other places, there is a method here that belongs on the charge class.

src/backend/tests/utilities/time.py
```diff
@@ -4,6 +4,7 @@
 
 class Time:
 
+    OVER_TWENTY_YEARS_AGO = (date.today() + relativedelta(years=-21)).strftime('%m/%d/%Y')
```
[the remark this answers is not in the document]
wittejm: Sure, makes sense

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: Sure, refactoring this and also using a shared method for when we apply MRC and MRD rules, and checking a charge's own time eligibility, makes sense (and putting it on a different ticket sounds good), calling it something like ```Charge.date_that_applies_for_time_eligibility(self)```

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: The main part that matters for the displayed result is: The top line on the charge is the red "Ineligible" label, and the displayed reason whether it's from the Type eligibilty for the Time eligibility, is clear that it's ineligible because there was a subsequent arrest or conviction. As Kent said, TypeEligibility is a function of the charge analyzed by itself whereas time eligibility is a function of the record as a whole, and I'm inclined to stick with that. The way we render the result can be a function of that underlying specification when we present to the user.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: Sure. we don't do anything here the disposition is missing or unknown (they're not subject to time analysis), so this check can just become `convicted()`.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: Yea, gotta fix this.

The comparison if `class_b_felony.disposition.date > TimeAnalyzer.TWENTY_YEARS_AGO` is checking that the date was after the date of 20 years ago, not that the conviction was "more than 20 years" old. The logic is correct by relying on the fact that if the felony is more than 20 years old and it has no subsequent charges, then it has already been set as eligible by the other parts of the time analyzer. 

I can rewrite this logic because it's difficult to interpret. The TimeAnalyzer can just have the constant `TWENTY_YEARS` and not `TWENTY_YEARS_AGO`, in keeping with the other time analysis implementation

src/backend/tests/utilities/time.py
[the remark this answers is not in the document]
wittejm: In fact the charge becomes eligible at 20 years on the dot, so I can use that to match what the other time eligibility tests do.

src/backend/tests/expunger/test_time_analyzer.py
wittejm: Sure.

wittejm:
Incorporated most of @NickSchimek 's comments. What I changed:

- Move b felony test cases back into `test_time_analyzer.py` . They have a shared utility function if`create_class_b_felony_charge` but no setUp() or internal variables, so I’m inclined to leave them as standalone test cases. 
- `OVER_TWENTY_YEARS_AGO` is 20 years +1 day instead of 21 years.
- Updated the logic for the 20-year restriction for readability
- `if not acquitted()` => `if convicted()`

Also I replaced the type ineligibility from subsequent arrests with time ineligibility. Getting this to render accurate results will take changes on the frontend that is out of scope here. With this update the result looks like this, misleading but not incorrect:


time_ineligibility_is_misleading_but_not_wrong


If we overwrite type eligibility we get a nice looking red label but we break the integrity of what type eligbility means (also the text spacing is busted, but whatever):


b_felony_correctly_ineligible_if_expressed_with_type_eligibility
  

Both are much better than what's currently in production, which is totally wrong:


b_felony_logic_is_totally_wrong_in_v 1 1

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
NickSchimek: yeah, you may be right.

NickSchimek: I feel there is some logic in this private method that can be extracted, but maybe this is not the right time to do it.

PR #660, opened 2019-12-18
Two non-traffic violations required to block time eligibility

### LINKED ISSUE #654: A single violation conviction isn't enough to block time eligibility
This is to fix a reported expunger logic bug.

The rule is: a prior conviction within 10 years normally blocks eligibility of other expungements, but this only counts convictions of misdemeanor level or above. It takes two violations within the last 10 years to block eligibility of other expungements, and the date of eligibility is 10 years following the date of the older conviction.

This implements a clause of the subsection 137.255(7)(b): "A single violation, other than a motor vehicle violation, within the last 10 years is not a conviction under this subsection". 

I've already written the test cases on a separate branch here: 

https://github.com/wittejm/recordexpungPDX/blob/two-violations-required-to-block-time-eligibility/src/backend/tests/expunger/test_time_analyzer.py

The open PR #645 does some refactoring of the time analyzer, so this issue should depend on that change.

wittejm:
closes #654

### DIFF
--- a/src/backend/tests/expunger/test_time_analyzer.py
+++ b/src/backend/tests/expunger/test_time_analyzer.py
@@ -238,3 +238,80 @@ class TestSecondMRCLogic(unittest.TestCase):
         self.run_expunger(one_year_old_conviction, nine_year_old_conviction)
 
         assert one_year_old_conviction.expungement_result.time_eligibility.date_will_be_eligible == eligibility_date
+
+
+class TestViolations(unittest.TestCase):
+
+    def setUp(self):
+        self.expunger = ExpungerFactory.create()
+
+    @staticmethod
+    def violation(disposition):
+        return ChargeFactory.create(
+            name='Failure to Send Child to or Maintain Child in School',
+            statute='339.020',
+            level='Violation Class C',
+            disposition=disposition)
+
+    @staticmethod
+    def traffic_violation(disposition):
+        return ChargeFactory.create(
+            name='Made-up traffic violation',
+            statute='825.999',
+            level='Class C traffic violation',
+            disposition=disposition)
+
+    def test_one_recent_violation_does_not_block_eligibility(self):
+        violation_two_years_ago = TestViolations.violation(['Convicted', Time.TWO_YEARS_AGO])
+        misdemeanor_four_years_ago = ChargeFactory.create(disposition=['Convicted', Time.FOUR_YEARS_AGO])
+
+        self.expunger.charges = [violation_two_years_ago, misdemeanor_four_years_ago]
+        self.expunger.run()
+        TimeAnalyzer.evaluate(self.expunger)
+
+        assert misdemeanor_four_years_ago.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE
+
+    def test_one_prior_violation_does_not_block_eligibility(self):
+        violation_seven_years_ago = TestViolations.violation(['Convicted', Time.SEVEN_YEARS_AGO])
+        misdemeanor_four_years_ago = ChargeFactory.create(disposition=['Convicted', Time.FOUR_YEARS_AGO])
+
+        self.expunger.charges = [misdemeanor_four_years_ago, violation_seven_years_ago]
+        self.expunger.run()
+        TimeAnalyzer.evaluate(self.expunger)
+
+        assert misdemeanor_four_years_ago.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE
+
+    def test_two_violations_block_eligibility(self):
+        violation_seven_years_ago = TestViolations.violation(['Convicted', Time.SEVEN_YEARS_AGO])
+        violation_less_than_ten_years_ago = TestViolations.violation(['Convicted', Time.LESS_THAN_TEN_YEARS_AGO])
+        misdemeanor_four_years_ago = ChargeFactory.create(disposition=['Convicted', Time.FOUR_YEARS_AGO])
+
+        self.expunger.charges = [misdemeanor_four_years_ago, violation_seven_years_ago, violation_less_than_ten_years_ago]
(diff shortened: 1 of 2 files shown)


src/backend/tests/expunger/test_time_analyzer.py
```diff
@@ -238,3 +238,80 @@ def test_mrc_is_eligible_in_two_years(self):
         self.run_expunger(one_year_old_conviction, nine_year_old_conviction)
 
         assert one_year_old_conviction.expungement_result.time_eligibility.date_will_be_eligible == eligibility_date
+
+
+class TestViolations(unittest.TestCase):
```
KentShikama: I think this effectively duplicates a number of tests but I don't have the time to double check right now. Since the bug is about fragile string comparisons, I think it would be enough to adjust the string of an existing test.

wittejm: You're right, there are some tests for this. I'll take another look and update

wittejm:
So there are a bunch of tests for how the expunger handles violations. I had overlooked them because they test the expunger and not the time eligibility results. 

They’re not behavioral tests, and after looking at them I realized that I really like behavioral tests.

The existing tests all look at the implementation of the expunger and verify that the expunger logic, here,

https://github.com/codeforpdx/recordexpungPDX/blob/244a8d947098ce310f4dd52b0e9a673678e8f979/src/backend/expungeservice/expunger/expunger.py#L103

does what it's supposed to and results in various combos of violations/non-violations being assigned as the MRC and second-MRC variables "as expected" according to the way those variables are supposed to work within the expunger. 

"MRC" has a semantically incorrect name because, for example, if the record has nothing but a violation conviction, the MRC is None. That makes sense for the existing expunger logic because a single violation doesn't affect other charges' eligibility. But there is this and several other circumstances where MRC and second-MRC mean surprising things, which is why there are so many different expunger tests for this. 

None of it is a behavioral test on how the expunger will produce time eligibility results on a record. If the underlying expunger logic changes, the tests that verify that logic will break and need to be rewritten, even if the expungement results on any given record never change. But we don’t have any behavioral tests that could survive an expungement refactor unchanged. Shouldn’t we?

Pinging @NickSchimek again because it’s his baby

NickSchimek:
> So there are a bunch of tests for how the expunger handles violations. I had overlooked them because they test the expunger and not the time eligibility results.
> 
> They’re not behavioral tests, and after looking at them I realized that I really like behavioral tests.
> 
> The existing tests all look at the implementation of the expunger and verify that the expunger logic, here,
> 
> https://github.com/codeforpdx/recordexpungPDX/blob/244a8d947098ce310f4dd52b0e9a673678e8f979/src/backend/expungeservice/expunger/expunger.py#L103
> 
> does what it's supposed to and results in various combos of violations/non-violations being assigned as the MRC and second-MRC variables "as expected" according to the way those variables are supposed to work within the expunger.
> 
> "MRC" has a semantically incorrect name because, for example, if the record has nothing but a violation conviction, the MRC is None. That makes sense for the existing expunger logic because a single violation doesn't affect other charges' eligibility. But there is this and several other circumstances where MRC and second-MRC mean surprising things, which is why there are so many different expunger tests for this.
> 
> None of it is a behavioral test on how the expunger will produce time eligibility results on a record. If the underlying expunger logic changes, the tests that verify that logic will break and need to be rewritten, even if the expungement results on any given record never change. But we don’t have any behavioral tests that could survive an expungement refactor unchanged. Shouldn’t we?
> 
> Pinging @NickSchimek again because it’s his baby

@wittejm The expunger is basically a middle man. we don't actually need it. All it does is setup data, data that should belong in the record object, so that the time analyzer can do its job.

In the beginning there was the expunger that fed setup data to a type analyzer, and a time analyzer. I refactored the type analyzer so that we no longer needed it as all of the code in the type analyzer belonged in the charge class. And now that the type analyzer is gone, we are left with an expunger class that has code that belongs in the record class. The record class should know what its MRC is or 2nd MRC by sending a message to the Record object. Hence, the expunger is not needed, and all that should be needed is the time analyzer. 

Regarding the tests you referred to, I believe they just check the data needed for the time analyzer is setup correctly, for example: it checks the mrc and 2nd mrc variable are set correctly in the expunger class. They are unit tests, and not meant to be behavioral tests. There are some functional tests that are meant to test the behavior here: `test_crawler_expunger.py`

MRC I believe is short for most recent conviction. I used that terminology because that is how michael refers to it. What would you suggest calling it?

wittejm:
I did see the tests in `test_crawler_expunger.py` and I should have acknowledged those. The tests I wrote for this PR were the level of behavioral tests that I have in mind, and I imagine there’s a better name for them. 

But since this PR was for just a string compare fix, I can take those out and maybe we could talk more about testing strategy elsewhere.

I didn’t mean that to end up as a rant about how the expunger works and I think it came out that way. My best stab at variable names might be `most_recent_conviction_that_blocks_eligibility` and  `second_most_recent_conviction_that_blocks_eligibility`. I’m not sure those are quite accurate either. Sorry for derailing here.

I'm going to replace this PR without the test cases, and with a new title and branch that are more accurate.

NickSchimek:
I didn't see it as a rant, but rather an opportunity to explain my feelings about the current state of the expunger. It's a work in progress.

Behavioral testing is applied to the entire system or portions of the whole system. Unit tests are essential to ensure the implementation is correct, and behavioral testing is vital to ensure the application runs correctly. Both tests have their strengths and weaknesses; hence, both kinds of tests are needed to test an application thoroughly.

Behavioral testing can be viewed as when a user pushes a button on the browser, then we expect a specific behavior to occur in the application. Behavioral tests are unconcerned with how it was accomplished, and are only concerned with what was accomplished.

Here is a good link for a more detailed explanation: http://softwaretestingfundamentals.com/black-box-testing/

PR #664, opened 2019-12-20
Check 401 error message match before app logout

### LINKED ISSUE #560: Ensure that 401 from OECI login doesn't logout user on frontend

wittejm:
A 401 error gets thrown for both an app auth failure and an OECI auth failure, and the forntend logs out from the app in both cases. The app should only log out if the error is from the app login attempt. 

The only visible difference between an OECI error and an app error is the attached message, so we do a direct string comparison here.

Closes #560


wittejm:
Going to merge for the release tonight this because it's such a useful fix and I've already tested it out it at our test url, https://recordexpungpdx-test.herokuapp.com

PR #683, opened 2020-01-03
Disposition result always shows arrest date

wittejm:
Fixes #666 

Results display fix. 

Here's a screencap of the new format:
 
always_showing_arrest_date

vs the prior version:

old_date_without_arrest

PR #685, opened 2020-01-03
Display error banner for any error messages on record

### LINKED ISSUE #684: Display an error banner with message if Record object has an error
The Record object returned by /api/search has an error field which has the message text for the error, if there is one. The frontend can just check for this message and display it in a banner per the mockup linked in #586

This issue is a small code change, and is a required piece for #652, #586. Those issues describe larger feature enhancements but this is an incremental improvement.

wittejm:
Closes #684 

This is a small amount of code added mainly just to the SearchResults component. An approach that would more closely fit the existing style of the frontend code would be to create a new RecordErrors component that wraps the set of RecordError components. 

It will definitely be worthwhile later to make a RecordError component when they get more sophisticated, namely with the added button described in issue #586

Here's what the displayed banner looks like:
example_error_banner

### DIFF
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -46,7 +46,7 @@ class Expunger:
         :return: True if there are no open cases; otherwise False
         """
         if self._open_cases():
-            self.record.errors.append('Open cases exist')
+            self.record.errors.append('All charges are ineligible because there is one or more open case.')
             return False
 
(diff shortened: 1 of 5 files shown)


hmarcks:
Style/layout looks good, thanks. Someone else will need to review the code.

maxkwallace:
@KentShikama thanks for the review here and in https://github.com/codeforpdx/recordexpungPDX/pull/683/ !

Issue #358, opened 2019-09-05
Add error to record when there are charges without a disposition


NickSchimek:
## Description

Currently when there is a charge without a disposition a message is written to `type_eligibility_reason`, however type eligibility is determined upon charge creation not during expungement, and it would only affect time eligibility. Need to move the message from type_reason to time_reason, and add a message to the record so that the user can be informed that errors may exist in the records time analysis.

This task requires #261 to be completed first. Could also combine the two tasks into one PR.

Also it requires PR #357 to be merged in order to avoid conflicts.

### This task is done when:
 - `record.errors` has error message appended to its array stating that closed cases containing charges exist without a disposition, and that time analysis may not be correct for the record. The only exception to this if the charge would normally skip time analysis (some level 800 charges are like that). If that's the case then there shouldn't be an error message on the record object.
 -

NickSchimek:
Further clarification. This is needed because the case may have a disposition, but OECI failed to properly log it.

wittejm:
Type eligibility based on a missing or unrecognized dispo should still be Needs More Analysis, and is set here:
https://github.com/codeforpdx/recordexpungPDX/blob/a40b678af5ef523996f7a3cc81d93b339d24def2/src/backend/expungeservice/models/charge.py#L28

Although the rest of this issue will be resolved by #686

PR #686, opened 2020-01-04
Append record errors for missing or unknown dispositions

### LINKED ISSUE #652: Charges with missing or unknown dispositions are record errors
A case with a missing or unknown disposition on a charge can be closed but still affect time eligibility with unknown impact. The charge could be either convicted or dismissed which can make a big difference for expungement eligibility of other records. This error should be attached to the record and reported in the frontend, with an informative warning message, like: "This record has charges with missing or unknown dispositions, which may affect the time eligibility of other charges." 

This issue depends on open PRs #645 and #648, which already impact the way that missing/unknown dispositioned charges are being handled by the expunger. 

### This task is done when:
 - The expunger attaches charges with missing and unknown dispositions to `record.errors`

A separate PR should expand the warning message on the frontend, which is related to issue #586

wittejm:
A missing or unknown disposition generates a record error unless the charge is "skippable" because it never affects time analysis, e.g. a traffic violation.
 
The test cases here (expanding the functional test case file) check for examples where a single error message is generated. 

Two more tests that would be useful:
- Test the case of multiple missing/unknown dispos and that the correct error messages are all present.
- Test that a traffic violation charge with a missing / unknown disposition doesn't generate a record error. 

Related to PR #684, which will show these error messages in the search results.

Closes #652 and #358 (which are mostly the same issue)

### DIFF
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -1,8 +1,10 @@
 from more_itertools import padnone, take
+from typing import Set
 
 from expungeservice.expunger.analyzers.time_analyzer import TimeAnalyzer
 from expungeservice.models.charge_types.felony_class_b import FelonyClassB
 from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+from expungeservice.models.disposition import DispositionStatus
 
 
 class Expunger:
@@ -48,6 +50,7 @@ class Expunger:
         if self._open_cases():
             self.record.errors.append('All charges are ineligible because there is one or more open case.')
             return False
+        self.record.errors += self._build_disposition_errors(self.charges)
 
         self.charges = Expunger._without_skippable_charges(self.charges)
         self.acquittals, self.convictions, _ = Expunger._categorize_charges(self.charges)
@@ -124,3 +127,42 @@ class Expunger:
             if isinstance(charge, FelonyClassB):
                 class_b_felonies.append(charge)
         return class_b_felonies
+
+    @staticmethod
+    def _build_disposition_errors(charges):
+        record_errors = []
+        cases_with_missing_disposition, cases_with_unknown_disposition = Expunger._filter_cases_with_errors(charges)
+        if cases_with_missing_disposition:
+            record_errors.append(Expunger._build_disposition_error_message(
+                cases_with_missing_disposition, "a missing"))
+        if cases_with_unknown_disposition:
+            record_errors.append(Expunger._build_disposition_error_message(
+                cases_with_unknown_disposition, "an unrecognized"))
+        return record_errors
+
+    @staticmethod
+    def _filter_cases_with_errors(charges):
+        cases_with_missing_disposition : Set[str] = set()
+        cases_with_unknown_disposition : Set[str] = set()
+        for charge in charges:
+            if not charge.skip_analysis():
+                case_number = charge.case()().case_number
+                if not charge.disposition:
+                    cases_with_missing_disposition.add(case_number)
+                elif charge.disposition.status == DispositionStatus.UNKNOWN:
+                    cases_with_unknown_disposition.add(case_number)
+        return cases_with_missing_disposition, cases_with_unknown_disposition
+
+    @staticmethod
+    def _build_disposition_error_message(error_cases, disposition_error_name):
+        if len(error_cases) == 1:
+                error_message = (
+f"""Case {error_cases.pop()} has a charge with {disposition_error_name} disposition.
+This is likely an error in the OECI database. Time analysis is ignoring this charge and may be inaccurate for other charges.""")
+        else:
+            cases_list_string = ", ".join(error_cases)
+            error_message = (
+f"""The following cases have charges with {disposition_error_name} disposition.
+This is likely an error in the OECI database. Time analysis is ignoring these charges and may be inaccurate for other charges.
+Case numbers: {cases_list_string}""")
+        return error_message
--- a/src/backend/tests/functional_tests/test_crawler_expunger.py
+++ b/src/backend/tests/functional_tests/test_crawler_expunger.py
@@ -56,7 +56,6 @@ def test_partial_dispos(partial_dispos_record):
     expunger = Expunger(partial_dispos_record)
     assert expunger.run()
 
-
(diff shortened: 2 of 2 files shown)


src/backend/expungeservice/expunger/expunger.py
```diff
@@ -123,4 +126,46 @@ def _class_b_felonies(charges):
...
+                    unknown_dispo_error_message = (
+                        f"Case {cases_with_unknown_disposition[0]} has a charge with an unrecognized disposition. " +
+                        message_shared_content)
+                else:
+                    cases_list_string = ", ".join(cases_with_unknown_disposition)
+                    unknown_dispo_error_message = (
+                        "The following cases have charges with an unrecognized disposition." +
+                        message_shared_content +
+                        f"\nCase numbers: {cases_list_string}")
+                record.errors.append(unknown_dispo_error_message)
```
KentShikama: I would prefer to pass in `self.charges`, return a list of record errors (`List[str]`), and then `+=` the result to record.errors in the `run` method. We want to keep all private functions in this class side-effect free.

src/backend/expungeservice/expunger/expunger.py
```diff
@@ -123,4 +126,46 @@ def _class_b_felonies(charges):
...
+                if len(cases_with_unknown_disposition) == 1:
+                    unknown_dispo_error_message = (
+                        f"Case {cases_with_unknown_disposition[0]} has a charge with an unrecognized disposition. " +
+                        message_shared_content)
+                else:
+                    cases_list_string = ", ".join(cases_with_unknown_disposition)
+                    unknown_dispo_error_message = (
+                        "The following cases have charges with an unrecognized disposition." +
+                        message_shared_content +
+                        f"\nCase numbers: {cases_list_string}")
```
KentShikama: Use 

```
f"""This
is a multiline
string with {variable}"""
```

for multi-line strings

src/backend/expungeservice/expunger/expunger.py
```diff
@@ -123,4 +126,46 @@ def _class_b_felonies(charges):
...
+                if not charge.disposition:
+                    if not case_number in cases_with_missing_disposition:
+                        cases_with_missing_disposition.append(case_number)
+                elif charge.disposition.status == DispositionStatus.UNKNOWN:
+                    if not case_number in cases_with_unknown_disposition:
+                        cases_with_unknown_disposition.append(case_number)
+            message_shared_content = (
+                "This is likely an error in the OECI database. " +
+                "Time analysis is ignoring these charges and may be inaccurate for other charges.")
+            if cases_with_missing_disposition:
```
KentShikama: Will this ever happen if `charge.skip_analysis()` is `True`?

src/backend/expungeservice/expunger/expunger.py
```diff
@@ -123,4 +126,46 @@ def _class_b_felonies(charges):
...
-        return class_b_felonies
\ No newline at end of file
+        return class_b_felonies
+
+    @staticmethod
+    def _mark_missing_or_unknown_dispositions_as_record_errors(record):
+        cases_with_missing_disposition : List[str] = []
+        cases_with_unknown_disposition : List[str] = []
+        for charge in record.charges:
+            if not charge.skip_analysis():
```
KentShikama: You may want to extract a method to handle each non-skipped charge.

src/backend/expungeservice/expunger/expunger.py
wittejm: Will do. I thought about that and decided to keep the `run` method looking clean, but I think you're right that the functional style is more important here.

src/backend/expungeservice/expunger/expunger.py
wittejm: Not sure what you mean. The result of `skip_analysis()` is always unaffected by the disposition, so the charges with `skip_analysis() == True` never should and never do get collected here.

KentShikama: I was just asking a question. Now that I know the answer, I would suggest to indent this but that should naturally happen if you extract a method to handle each non-skipped charge.

wittejm:
> You may want to extract a method to handle each non-skipped charge.

I looked for a meaningful way to do this, but there's not much to extract because the method still needs to accumulate case numbers and then translate them jointly into the error message(s).

Though I did manage to extract a method to build the error message to shave off a few lines.

I also just added the test case that multiple record errors get added due to multiple disposition errors.

KentShikama:
I think the indentation before threw me off because it suggested you were doing something with the error message for each charge. Now it is clear that you are handling the error message after all charges have been looped over.

src/backend/expungeservice/expunger/expunger.py
```diff
@@ -123,4 +126,38 @@ def _class_b_felonies(charges):
...
-        return class_b_felonies
\ No newline at end of file
+        return class_b_felonies
+
+    @staticmethod
+    def _build_record_errors_list_for_missing_or_unknown_dispositions(charges):
+        record_errors = []
+        cases_with_missing_disposition : Set[str] = set()
+        cases_with_unknown_disposition : Set[str] = set()
+        for charge in charges:
```
KentShikama: You can now extract the entire loop.

```
cases_with_missing_disposition, cases_with_unknown_disposition = Expunger.filter_cases_with_errors(...)
if cases_with_missing_disposition:
    ...
```

src/backend/expungeservice/expunger/expunger.py
```diff
@@ -48,7 +50,8 @@ def run(self):
         if self._open_cases():
             self.record.errors.append('All charges are ineligible because there is one or more open case.')
             return False
-
+        disposition_errors = self._build_record_errors_list_for_missing_or_unknown_dispositions(self.charges)
+        self.record.errors += disposition_errors
```
KentShikama: You can inline this

`self.record.errors += self._build_record_errors_list_for_missing_or_unknown_dispositions(self.charges)`

But perhaps your reluctance suggests that maybe the function name itself should just be

`self.record.errors += self._build_disposition_errors(self.charges)`

wittejm: Yea that sounds good. I walked around in circles a few times here and got dizzy I guess, it happens.

PR #689, opened 2020-01-06
Force frontend to logout state if not invalid OECI auth

KentShikama:
Follow up from https://github.com/codeforpdx/recordexpungPDX/pull/664

`error.response.message` doesn't exist (now or previously). Previously the error message was embedded in the HTML response body. Now the jsonified error message is at `error.response.data.message`.


wittejm:
Nice nice. This works for me locally now. I want to try pushing it to our test url on heroku before we merge, which I can do in a few hours, unless you want to do it.

PR #692, opened 2020-01-07
Extract ChargesWithSummary from Expunger

KentShikama:
We ultimately aim to make the TimeAnalyzer more functional. This commit is a step towards that. Instead of passing in the entire Expunger, we extract a dataclass ChargesWithSummary that contains the fields that the TimeAnalyzer operates on.


KentShikama:
I'll take care of those in a following PR as this one is already large enough.

Issue #555, opened 2019-11-15
Cleanup of files that don't end in line feed


keppel2:
Running this `find . -type f -exec echo -n {}, \; -exec hexdump -ve '/1 "%02X"' {}  \; -exec echo \; | grep -v 0A$` shows a few files that don't end in a line feed. This is in a category of text problems that can lead to issues later. There's no carriage returns in the entire repo--so that's good. This should only be considered complete when it's part of the automated tests.

keppel2:
It's fine to just clean them up occasionally like #693--it might be weird for beginners to see check-in test errors.

keppel2:
We have scripts that appear to lint the frontend(husky). Fixing LF there should be fine.

wittejm:
AFAICT the only python files without a newline are __init__ files.

PR #693, opened 2020-01-07
Add Black: the opinionated auto-formatter

KentShikama:
Note you can run `pip install pre-commit` followed by `pre-commit install` once to have all future git commits be auto-formatted by black. 

Addresses https://github.com/codeforpdx/recordexpungPDX/issues/555 for the Python source files that we control.


wittejm:
Love it. I'm able to apply the reformatter on new code but
> You should be able to reproduce this commit with the command  `black .`

This command is not recognized for me. Maybe we can go over it when we meet up shortly.

keppel2:
Does this have an option to just check for some bad cases (mixed spaces/tabs) and reject in our checkin tests?

KentShikama:
[flake8](http://flake8.pycqa.org/en/latest/#) seems to be the most popular tool for that in the Python ecosystem.

PR #713, opened 2020-01-09
Minor wording and format changes to rendered results

wittejm:
* if the disposition is missing, it's shown as Missing, not Unknown
 * the 'reason' string wraps around to not overlap other text
 * adjusted the Charge detail column widths to 40/60%, not 30/70%, to make more room for our lengthier (and more informative) reason strings

 The text wrap fixes the following example:
b_felony_correctly_ineligible_if_expressed_with_type_eligibility

And here's the effect of the width change. I tried 50/50 but that added way too much awkward whitespace. 

old:

narrow

vs new:

extra_space_for_reason_string

I'm tagging @hmarcks since it includes a design change.

### DIFF
--- a/src/frontend/src/components/Charge/index.tsx
+++ b/src/frontend/src/components/Charge/index.tsx
@@ -41,7 +41,7 @@ export default class Charge extends React.Component<Props> {
       if (disposition === null) {
         return (
           <li className="mb2">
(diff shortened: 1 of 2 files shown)


hmarcks:
Looks good :-)

PR #714, opened 2020-01-10
show charge as 'ineligible' if it's time ineligible and has no eligibility date

### LINKED ISSUE #702: Time-ineligible Class B Felony should display as "ineligible"
The backend correctly says that this charge is "always time-ineligible", but the frontend should render this with a red "Ineligible" label, while the type-eligibility remains unchanged.
b_felony_with_subsequent_charge

wittejm:
In theory, any charge that is time-ineligible and has no date should be interpreted as "never eligible". In practice, this only ever happens in one case which is a Class B Felony that has subsequent convictions on the record. 

So this might seem a bit hacky to handle this single case in the frontend. But at least B felonies look nice now:

fixed_b_felony

The top-level eligibility label on a charge is just an interpretation of the type and time eligibility. It could be generated in the backend, which also eliminates the need for the frontend to handle "unexpected data from backend" issues.

Closes #702

### DIFF
--- a/src/frontend/src/components/Eligibility/index.tsx
+++ b/src/frontend/src/components/Eligibility/index.tsx
@@ -51,6 +51,12 @@ export default class Eligibility extends React.Component<Props> {
     };
 
     const eligibility = () => {
+
+      // Time ineligibility without a date (meaning "never eligible") beats all other rules for eligibility
+      if (time_eligibility && time_eligibility.status === 'Ineligible' && time_eligibility.date_will_be_eligible == null) {
+        return ineligible;
+      }
+
       switch (type_eligibility.status) {
         case 'Eligible':
           return handleWhenTypeEligibile();
--- a/src/frontend/src/components/Eligibility/index.tsx
+++ b/src/frontend/src/components/Eligibility/index.tsx
@@ -53,6 +53,7 @@ export default class Eligibility extends React.Component<Props> {
     const eligibility = () => {
 
       // Time ineligibility without a date (meaning "never eligible") beats all other rules for eligibility
+      // Currently this only occurs on Class B felonies, but the rule theoretically applies always, so it is checked first.
       if (time_eligibility && time_eligibility.status === 'Ineligible' && time_eligibility.date_will_be_eligible == null) {
         return ineligible;
       }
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/Eligibility/index.tsx
```diff
@@ -51,6 +51,12 @@ export default class Eligibility extends React.Component<Props> {
     };
 
     const eligibility = () => {
+
+      // Time ineligibility without a date (meaning "never eligible") beats all other rules for eligibility
+      if (time_eligibility && time_eligibility.status === 'Ineligible' && time_eligibility.date_will_be_eligible == null) {
+        return ineligible;
+      }
+
```
KentShikama: How does this affect line 46 of this file (hit the expand arrow above on Github to see this)?

wittejm: Excellent question. That line will no longer ever run. I considered taking it out now, but my thought was to make that part of getting rid of most of the case-checking logic from this section in the frontend, as described in the PR description, in a future PR.

KentShikama:
Is this for Class B Felonies only?

src/frontend/src/components/Eligibility/index.tsx
KentShikama: I believe it was never run before too, unless there was a bug in the backend logic. I'm trying to think how this can be integrated into the switch statement below.

wittejm:
Yep, only occurs on Class B felonies for the moment. This can potentially change, e.g. Michael just suggested the idea of identifying a subclass of B felonies that would still have the same "never" rule. So I think the frontend logic should remain blind to charge types.

KentShikama:
Are convicted Class B Felonies always processed as "Needs more analysis" for type eligibility?

wittejm:
Yes. Improving that is the hope with subclassing B felony with new charge types. So we may end up with such a charge type that is Type Eligible but still can suffer from the Never rule of a B felony.

KentShikama:
I see. In that case, I would just insert this statement inside the existing switch statement. When you introduce the subclasses, you can simultaneously extract out as this commit does. That way it is clearer to anyone who only has the git log as to what has happened.

wittejm:
I kept this condition above the switch statement because the rule beats everything for eligibility on the charge: applying for `type_eligibility` of either `eligible` or `needs more analysis` (and, trivially, `ineligible`).
 
If it only gets checked in the  `needs more analysis` case, then , the only reason it's written that way is because that's how the Class B felony happens to work. Without that implicit knowledge, this code doesn't make sense. I don't think we should rely on someone just remembering or stumbling into the discovery that they will need to change it later?

wittejm:
Yea, that all make sense. I can add a comment line providing context. It definitely seems to be a style issue, and fairly minor in this case. Along with merging this bugfix I'll make a new issue to push all of this logic into the backend.

PR #715, opened 2020-01-10
Missing disposition is not an error if the case is open

### LINKED ISSUE #701: Make the "open cases" error banner more informative.
A small change: add "Time analysis is ignoring these cases and may be inaccurate for other charges."

### This task is done when:
 -
 -

### LINKED ISSUE #712: If a case is "open" and it has missing dispositions, it shows redundant error messages
Sometimes a case is identified as "open", but it has all the required disposition information so the expunger can still run analysis on it. This occurs sometimes as an OECI data error, but the inconsistency should still be pointed out to the user. 

If a case is actually open and there is no disposition on the charges yet, it will report the "Open Case" error, and also the "missing dispositions" error. This has weird / confusing results like this:

open_case_and_missing_dispo_error

In this case there is definitely not a "likely error in the OECI database", 

So, any time there is an open case, and it has missing dispositions, don't report the missing dispositions as an error because we can safely assume that it is not.

Also, let's rephrase the "likely error in the OECI database", because for now there are fairly common edge cases that we just haven't addressed yet, e.g. #705

wittejm:
closes #712

This also updates "likely an error with OECI", to "might be", because it's so accusatory and unfair. 

also closes #701

### DIFF
--- a/src/backend/expungeservice/expunger/expunger.py
+++ b/src/backend/expungeservice/expunger/expunger.py
@@ -32,7 +32,7 @@ class Expunger:
         if len(open_cases) > 0:
             case_numbers = ",".join([case.case_number for case in open_cases])
             self.record.errors += [
-                f"All charges are ineligible because there is one or more open case: {case_numbers}. Open cases with valid dispositions are still included in time analysis."
+                f"All charges are ineligible because there is one or more open case: {case_numbers}. Open cases with valid dispositions are still included in time analysis. Otherwise they are ignored, so time analysis may be inaccurate for other charges."
             ]
         self.record.errors += self._build_disposition_errors(self.record.charges)
         TimeAnalyzer.evaluate(self.charges_with_summary)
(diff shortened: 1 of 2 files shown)


src/backend/expungeservice/expunger/expunger.py
```diff
@@ -64,7 +64,9 @@ def _filter_cases_with_errors(charges: List[Charge]):
             if not charge.skip_analysis():
                 case_number = charge.case()().case_number
                 if not charge.disposition:
-                    cases_with_missing_disposition.add(case_number)
+                    if charge.case()().closed():
```
KentShikama:
```
if not charge.disposition and charge.case()().closed():
```

wittejm: NayIf there is a charge disposition, line 70 should run. If this condition checked jointly, it's possible for there to be  charge disposition but line 70 doesn't run. I tried writing it as an `and` expression and it caused some unrelated test to fail.

KentShikama: `elif charge.disposition and charge.disposition.status == DispositionStatus.UNRECOGNIZED`

KentShikama:
```
if not charge.disposition and charge.case()().closed():
  ...
elif charge.disposition and charge.disposition.status == DispositionStatus.UNRECOGNIZED:
  ...
```

wittejm: Ok sounds good

Issue #681, opened 2020-01-03
Bug: A single violation doesn't receive the 3-year time eligibility constraint


wittejm:
The time analyzer sets time eligibility based on assigning the MRC (most recent conviction) and second-most recent conviction. A record with only a single violation conviction doesn't get assigned an MRC, and the violation is never given the correct time eligibility constraint. 

The open PR (#682 ) shows the test case that fails without the fix. (The PR includes the bugfix to the expunger as well)

Issue #619, opened 2019-12-09
Time eligibility reasons are inconsistent


wittejm:
The time eligibility reasons are inconsistent between the mockup, the logic flowchart, and the existing output from the TimeAnalyzer.

Namely, in the flowchart some of the time-constraint labels include  "Time-ineligible under [statute-subsection]", while some others say , "eligible at [date]".

The mockup and time analyzer both use some inconsistent / confusing wording IMO. For example the reason "Most recent conviction is less than three years old" (from the time analyzer): 

1. is not explicit that this is referring to itself as the conviction 
2. is misleading about what the actual rule is for time constraints generally: for example if the most recent conviction is 4 years old, it still renders all other expungements ineligible.

The reason "no more than 1 arrest in 3 years" (from the mockup) makes sense if it's attached to a dismissed charge, but is incorrect if it's attached to a convicted charge. "No conviction within 3 years" is a valid rule if it is attached to a conviction but is incorrect if it is attached to a dismissal.

There are 5 different time-constraint rules, and I propose using the following wordings:

 - Convictions are ineligible for 3 years
 - No-complaint arrests are ineligible for 1 year
 - Must have no other convictions for 10 years
 - Must have no other arrests for 3 years
 - B felony convictions are ineligible for 20 years 

Since the flowchart and time analyzer sometimes refer to the statute subsection that applies:  if we want to we could add this as another line in the result. 

Pings for @NickSchimek and @hmarcks but any opinions welcome
 
### This task is done when:
 - the `reason` values in `time_analyzer.py` have been updated.

wittejm:
So, just because there are a bunch of ways to write this and I want to shoot for consistency more than anything, and partly to demonstrate that I think trying to make it really clear is a pain in the butt, here's a different set of reasons. Arguably, these phrases fit more naturally into reading the entry as  "Eligible at [date] because that date is...[reason]:

- Ten years from most recent conviction
- Ten years from second-most recent conviction
- Three years from most recent arrest
- Three years from most recent arrest - Recommend sequential expungement
- Three years from second-most recent arrest
- One year from No Complaint arrest
- Three years from conviction
- Twenty years from B felony conviction

The reason of "Recommend sequential expungement" occurs if there is exactly one arrest from the last 3 years. That's because the most recent arrest is time-eligible (if there are no convictions within 10 years), and once you expunge it, any other arrests become time-eligible because they are no longer blocked by the arrest that just got expunged.

[...]

PR #716, opened 2020-01-10
Single conviction doesn't block arrests on same case

aemrich:
Addresses #387. 

Tested for some cases in the issue. Fixes two of the reported bugs for real examples. Other reported bugs relevant to issue may be errors. There may be extra edge cases we can add tests for.

### DIFF
--- a/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
+++ b/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
@@ -149,10 +149,10 @@ class TimeAnalyzer:
                         charge.set_time_eligible()
         elif mrc and second_mrc:
                 for charge in mrc.case()().charges:
-                    if second_mrc.disposition.date and charge.acquitted(): # type: ignore
+                    if charge.acquitted():
                             charge.set_time_ineligible(
                                 'Multiple convictions within last ten years',
-                                charges_with_summary.second_most_recent_conviction.disposition.date # type: ignore
+                                second_mrc.disposition.date # type: ignore
                                 + relativedelta(years=+TimeAnalyzer.TEN_YEARS)
                                 )
 
--- a/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
+++ b/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
@@ -141,18 +141,16 @@ class TimeAnalyzer:
         # AND all other charges in the same case are arrests
         # Then make all other charges in the case time eligible
         mrc = charges_with_summary.most_recent_conviction
-        if not mrc is None and not mrc.expungement_result.time_eligibility is None and mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE:
-            if charges_with_summary.second_most_recent_conviction:
-                pass
-            else:
+        second_mrc = charges_with_summary.second_most_recent_conviction
+        if mrc and mrc.expungement_result.time_eligibility and mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE:
+            if not second_mrc:
                 for charge in mrc.case()().charges:
                     if charge.acquitted():
                         charge.set_time_eligible()
         else:
             if charges_with_summary.second_most_recent_conviction and not mrc is None:
                 for charge in mrc.case()().charges:
-                    if not charges_with_summary.second_most_recent_conviction.disposition is None:
-                        if charge.acquitted():
+                    if second_mrc.disposition and charge.acquitted():
                             charge.set_time_ineligible(
                                 'Multiple convictions within last ten years',
                                 charges_with_summary.second_most_recent_conviction.disposition.date
--- a/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
+++ b/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
@@ -2,6 +2,7 @@ from datetime import date
 from typing import List
 
 from dateutil.relativedelta import relativedelta
+from expungeservice.models.expungement_result import EligibilityStatus
 
 from expungeservice.expunger.charges_summarizer import ChargesWithSummary
 from expungeservice.models.charge import Charge
@@ -23,6 +24,7 @@ class TimeAnalyzer:
                 charges_with_summary.charges, "Time-ineligible under 137.225(7)(b)", elig_date
             )
             TimeAnalyzer._check_mrc_time_eligibility(charges_with_summary)
+            TimeAnalyzer._set_arrests_as_eligible(charges_with_summary)
         elif charges_with_summary.most_recent_dismissal:
             TimeAnalyzer._mark_all_acquittals_ineligible_using_mrd_date(charges_with_summary)
             TimeAnalyzer._mark_as_time_eligible(charges_with_summary.most_recent_dismissal.case()().charges)
@@ -132,6 +134,30 @@ class TimeAnalyzer:
     def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_date: date):
         for charge in charges:
             charge.set_time_ineligible(reason, eligibility_date)
+
+    @staticmethod
+    def _set_arrests_as_eligible(charges_with_summary: ChargesWithSummary):
+        # If most recent conviction is time eligible
+        # AND all other charges in the same case are arrests
+        # Then make all other charges in the case time eligible
+        mrc = charges_with_summary.most_recent_conviction
+        if charges_with_summary.most_recent_conviction.expungement_result.time_eligibility.status \
+            is EligibilityStatus.ELIGIBLE:
+            if charges_with_summary.second_most_recent_conviction:
+                pass
+            else:
+                for charge in mrc.case()().charges:
+                    if charge.acquitted():
+                        charge.set_time_eligible()
+        else:
+            if charges_with_summary.second_most_recent_conviction:
+                for charge in mrc.case()().charges:
+                    if charge.acquitted():
+                        charge.set_time_ineligible(
+                            'Multiple convictions within last ten years',
+                            charges_with_summary.second_most_recent_conviction.disposition.date\
+                            + relativedelta(years=+TimeAnalyzer.TEN_YEARS)
+                            )
 
     @staticmethod
     def _mark_as_time_eligible(charges: List[Charge]):
--- a/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
+++ b/src/backend/expungeservice/expunger/analyzers/time_analyzer.py
@@ -141,8 +141,7 @@ class TimeAnalyzer:
         # AND all other charges in the same case are arrests
         # Then make all other charges in the case time eligible
         mrc = charges_with_summary.most_recent_conviction
-        if charges_with_summary.most_recent_conviction.expungement_result.time_eligibility.status \
-            is EligibilityStatus.ELIGIBLE:
+        if not mrc is None and not mrc.expungement_result.time_eligibility is None and mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE:
             if charges_with_summary.second_most_recent_conviction:
                 pass
             else:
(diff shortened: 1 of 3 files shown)


src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,30 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
         for charge in charges:
             charge.set_time_ineligible(reason, eligibility_date)
 
+    @staticmethod
+    def _set_arrests_as_eligible(charges_with_summary: ChargesWithSummary):
+        # If most recent conviction is time eligible
+        # AND all other charges in the same case are arrests
+        # Then make all other charges in the case time eligible
+        mrc = charges_with_summary.most_recent_conviction
```
KentShikama: Might as well extract out 

`second_mrc = charges_with_summary.second_most_recent_conviction`

to keep it consistent with `mrc`.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,30 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+        if not mrc is None and not mrc.expungement_result.time_eligibility is None and mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE:
```
KentShikama: Does

`if mrc and mrc.expungement_result.time_eligibility and mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE:`

work?

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,30 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+                pass
+            else:
+                for charge in mrc.case()().charges:
+                    if charge.acquitted():
+                        charge.set_time_eligible()
+        else:
+            if charges_with_summary.second_most_recent_conviction and not mrc is None:
+                for charge in mrc.case()().charges:
+                    if not charges_with_summary.second_most_recent_conviction.disposition is None:
+                        if charge.acquitted():
```
KentShikama: I think you can combine this to

`if second_mrc.disposition and charge.acquitted()`

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,30 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+    @staticmethod
+    def _set_arrests_as_eligible(charges_with_summary: ChargesWithSummary):
+        # If most recent conviction is time eligible
+        # AND all other charges in the same case are arrests
+        # Then make all other charges in the case time eligible
+        mrc = charges_with_summary.most_recent_conviction
+        if not mrc is None and not mrc.expungement_result.time_eligibility is None and mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE:
+            if charges_with_summary.second_most_recent_conviction:
+                pass
+            else:
```
KentShikama: You can just do

`if not second_mrc:`

and delete the `pass` clause.

src/backend/tests/expunger/test_time_analyzer.py
```diff
@@ -16,6 +16,118 @@ class TestSingleChargeAcquittals(unittest.TestCase):
...
+        assert three_yr_conviction.expungement_result.time_eligibility.date_will_be_eligible == date.today() + relativedelta(years=7, days=1)
+
+        assert arrest.expungement_result.time_eligibility.status is EligibilityStatus.INELIGIBLE
+        assert arrest.expungement_result.time_eligibility.date_will_be_eligible == date.today() + relativedelta(years=7)
+
+        # assert violation.expungement_result.time_eligibility.status is EligibilityStatus.INELIGIBLE
+        # assert violation.expungement_result.time_eligibility.date_will_be_eligible == date.today() + relativedelta(years=7)
+        #
+        # assert violation_2.expungement_result.time_eligibility.status is EligibilityStatus.INELIGIBLE
+        # assert violation_2.expungement_result.time_eligibility.date_will_be_eligible == date.today() + relativedelta(years=7)
```
KentShikama: Looks like these comments need to be cleaned up?

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,28 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+        # AND all other charges in the same case are arrests
+        # Then make all other charges in the case time eligible
+        mrc = charges_with_summary.most_recent_conviction
+        second_mrc = charges_with_summary.second_most_recent_conviction
+        if mrc and mrc.expungement_result.time_eligibility and mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE:
+            if not second_mrc:
+                for charge in mrc.case()().charges:
+                    if charge.acquitted():
+                        charge.set_time_eligible()
+        else:
```
KentShikama: What happens without this else block?

aemrich: Without the else block, arrests on a case with a second mrc don't have the correct eligibility date

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,28 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+            if charges_with_summary.second_most_recent_conviction and not mrc is None:
```
KentShikama: elif mrc and second_mrc:

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
KentShikama: Isn't that what line 23/24 of this file is for?

KentShikama: I see. I feel like there must be a better way to codify the combination of this and line 23/24 but I think this is fine for now.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,28 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+                for charge in mrc.case()().charges:
+                    if second_mrc.disposition and charge.acquitted():
```
KentShikama: Does `second_mrc.disposition` exist because of `mypy`? If so, you can just add a comment `# type: ignore` on line 156.

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,28 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+                            charge.set_time_ineligible(
+                                'Multiple convictions within last ten years',
+                                charges_with_summary.second_most_recent_conviction.disposition.date
```
KentShikama: `second_mrc.disposition.date`

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
```diff
@@ -133,6 +135,28 @@ def _mark_as_time_ineligible(charges: List[Charge], reason: str, eligibility_dat
...
+            if not second_mrc:
+                for charge in mrc.case()().charges:
+                    if charge.acquitted():
+                        charge.set_time_eligible()
+        else:
+            if charges_with_summary.second_most_recent_conviction and not mrc is None:
+                for charge in mrc.case()().charges:
+                    if second_mrc.disposition and charge.acquitted():
+                            charge.set_time_ineligible(
+                                'Multiple convictions within last ten years',
```
KentShikama: Please reference `137.225 7(b)` somehow here.

src/backend/tests/expunger/test_time_analyzer.py
```diff
@@ -16,6 +16,112 @@ class TestSingleChargeAcquittals(unittest.TestCase):
...
+            level='Violation',
+            case=case,
+            disposition=['Convicted', Time.THREE_YEARS_AGO])
+
+        case.charges = [three_yr_mrc, arrest, violation]
+        record = Record([case])
+        expunger = Expunger(record)
+
+        expunger.run()
+        # import pdb; pdb.set_trace()
```
KentShikama: Looks like a debug comment.

KentShikama:
As to the build errors, feel free to just add the line comment `# type: ignore` to both

```
expungeservice/expunger/analyzers/time_analyzer.py:154: error: Item "None" of "Optional[Charge]" has no attribute "disposition"
expungeservice/expunger/analyzers/time_analyzer.py:157: error: Item "None" of "Optional[Disposition]" has no attribute "date"
```

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
KentShikama: So based on what I read of 7(b), I think the charges should be set time ineligible until the mrc is 10 years old (not until the second mrc is 10 years old as is written now). @wittejm What do you think?

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
wittejm: These reason strings are all inconsistent but they can be fixed in a separate PR addressing the issue: #619

src/backend/expungeservice/expunger/analyzers/time_analyzer.py
KentShikama: After a bit more thinking, I think this else block is unnecessary. The eligibility for everything that is not the mrc is 10 years after the mrc date, which is taken care of in line 23/24. @wittejm Any input here?

wittejm: If the mrc becomes eligible after the second_mrc is 10 years old, then the arrests on the same case as the mrc fall under this rule, and have the same eligibility. The reason for the `else` here is the same as the reason for having this whole fix: the arrests on this case are not blocked by the mrc, but instead only by the second_mrc

wittejm: Though I think this  actually needs another tweak: the eligibility of the arrests on the same case as the mrc get pushed out by 10 years from the date of the second_mrc, and also by 3 years from the date of the mrc, whichever is farther out.

So for example if the mrc is 1 year old and the second_mrc is 9 years old, the arrests on the mrc's case become eligible in 2 years, not 1. 

It's confusing as hell. But I'm pretty sure there's a test case here that would be failing given the current logic.

KentShikama: Ah yes, I was thinking about 7(b) too strictly again. Here my high-level overview of the TimeAnalyzer without Class B Felonies as of now.

```
if mrc and second_mrc:
    # mrc and arrests in same case is eligible the max of mrc + 3 years and second_mrc + 10 years
    # everything else is eligible mrc + 10 years
elif mrc:
    # mrc and arrests in same case is eligible mrc + 3 years
    # everything else is eligible mrc + 10 years
elif mrd and second_mrd:
    # mrd is eligible second_mrd + 3 years
    # everything else is eligible mrd + 3 years
elif mrd:
    # mrd is eligible now
else:
    # no recent convictions or dismissals
```

KentShikama: The code also needs a check to see whether `mrc.case != second_mrc.case`. It feels like this logic should be grouped together with `_check_mrc_time_eligibility`, otherwise there will be a lot of duplicated code.

wittejm: I think refactoring this whole thing is probably a good idea, but let's definitely get this fix in so we can include it in a new version push this week. 

While at it there's probably a way to overhaul the way violations are handled with the mrc/second_mrc so we don't need edge case fixes for these new/old/old bugs: #718 #681 #654

Issue #631, opened 2019-12-11
Add search results summary data to /api/search/ endpoint response


wittejm:
This is the backend feature required to support the new frontend SearchResultsSummary component described in #630 .

The new data fields look like this: 
```
{
  total_cases: num_total_cases,
  total_charges: num_total_cases,
  case_summary: {
    eligible_now: [
      case_id_1,
      case_id_2
    ],
    ineligible: [
      case_id_3,
      case_id_4
    ],
  },
  charge_summary: {
    eligible_now: [
      charge_name_1,
      charge_name_2,
      
    ],
    will_be_eligible: [
      charge_name_3,
      charge_name_4
    ],
    ineligible: [
      charge_name_5,
      charge_name_6
    ]
  },
  balance_due_summary: [
    {
      county_name: county_name_1,
      balance_due: balance_due_1,
    }, 
    {
      county_name: county_name_2,
      balance_due: balance_due_2,
    }, 
  ]
}
 ```
### This task is done when:
 - The JSON response data includes these new data fields. The fields could be wrapped in a `results_summary` object which exists adjacent to the `record` field on the endpoint response object, so that the `record` data is unchanged from what is returned by the expunger. The summary data construction should occur in (or called from) [search.py](https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/endpoints/search.py)

wittejm:
Closed with #721

Issue #630, opened 2019-12-11
Add SearchResultsSummary Component in React


wittejm:
A new component will provide summary data about search results, appearing above search results details. 

The mockup for this feature is below. Use the HTML from http://dev.huntermarcks.net/search // [here's the relevant commit](https://github.com/hmarcks/expunge-pdx-frontend/commit/e62e80ee7ff9dac2c71c8d30679f7dc0513c1efc)

### This task is done when:
 - SearchResultsSummary component is implemented in React
 - The existing SearchResults Component should be renamed to SearchResultsDetail. These two components should be wrapped together into a single component called SearchResults, which can remain a member of AllRecords.
 - The component is integrated with `/api/search` endpoint, which must be expanded to provide the `case_summary` and `charge_summary`, and `balance_due_summary` data that get rendered in this component (see issue #631  for endpoint implementation). 
 

![Record Expunge Search Summary Copy](https://user-images.githubusercontent.com/6887982/71208008-c0399400-226d-11ea-8e8e-d225d1644be3.png)

hmarcks:
Created the HTML/CSS and updated the description

wittejm:
@hmarcks Awesome, thank you!

wittejm:
completed with #756

PR #720, opened 2020-01-14
new charge_eligibility value computed from type and time eligibility

### LINKED ISSUE #717: Determine overall charge Eligibility in the backend
There are 5 different kinds of overall charge eligibility, which is a function of both the time and type eligibility:

- Eligible now
- Will be eligible (date) 
- Possibly eligible now
- Possibly will be eligible (date)
- Ineligible

In the interest of moving expunger logic to the backend, and reducing risk of data errors that can be received by the frontend, we can tag charges with this overall eligibility variable.

This should probably use a new ChargeEligibility enum that has 5 values corresponding to the above. It can replace most of the logic in this frontend component: https://github.com/codeforpdx/recordexpungPDX/blob/master/src/frontend/src/components/Eligibility/index.tsx , including the error-checking that is performed there.

The charge eligibility provided to the frontend should contain both the enum value, so that the frontend can determine which color the charge label should have, and the eligibility label string (incuding the date, if necessary) so that the frontend doesn't need to conditionally generate the label.

wittejm:
ExpungementResult now computes a stateless Charge Eligibility value based on the charge's type and time eligibility.   

This replaces and simplifies the corresponding frontend logic for visualizing eligibility results. It will also make it easier to generate a record Summary component (#630, #631).

Closes #717

### DIFF
--- a/src/backend/expungeservice/models/expungement_result.py
+++ b/src/backend/expungeservice/models/expungement_result.py
@@ -10,6 +10,15 @@ class EligibilityStatus(str, Enum):
     INELIGIBLE = "Ineligible"
 
 
+class ChargeEligibilityStatus(str, Enum):
+    UNKNOWN = "Unknown"
+    ELIGIBLE_NOW = "Eligible now"
+    POSSIBLY_ELIGIBILE = "Possibly eligible"
+    WILL_BE_ELIGIBLE = "Will be eligible"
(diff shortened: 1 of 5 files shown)


wittejm:
Ah dang. Agreed. I'd hoped I could stop paying attention entirely to spacing and leaning fully on `black`. Didn't go back and check.

PR #721, opened 2020-01-15
Include results summary in search endpoint

wittejm:
Provides a new data object for populating the Summary panel that will lead the search results #630.

This summarizer sorts case numbers into 4 disjoint sets: `fully_eligible`, `fully_ineligible`, `partially_eligible`, and `other` . It's probably not useful to include `other` in the summary panel, but this way the summary object includes the complete set of case numbers, in case that is useful. 

County balances are computed across cases as a sum of floats, which accumulates a floating point error, but the sums get rounded off again and that seems fine to me.

### DIFF
--- a/src/backend/expungeservice/models/helpers/record_summarizer.py
+++ b/src/backend/expungeservice/models/helpers/record_summarizer.py
@@ -48,6 +48,8 @@ class RecordSummarizer:
             "partially_eligible": partially_eligible_cases,
             "other": other_cases,
         }
+        for county in county_balances.keys():
+            county_balances[county] = round(county_balances[county], 2)
         total_charges = len(record.charges)
         eligible_charges = [
             c.name
--- a/src/backend/expungeservice/models/helpers/record_summarizer.py
+++ b/src/backend/expungeservice/models/helpers/record_summarizer.py
@@ -1,0 +1,62 @@
+from expungeservice.models.expungement_result import ChargeEligibilityStatus
+from expungeservice.models.record_summary import RecordSummary
+
+
+class RecordSummarizer:
+    @staticmethod
+    def summarize(record):
+
+        fully_eligible_cases = []
+        fully_ineligible_cases = []
+        partially_eligible_cases = []
+        other_cases = []
+
+        county_balances = {}
+        for case in record.cases:
+            if not case.location in county_balances.keys():
+                county_balances[case.location] = case.get_balance_due()
+            else:
+                county_balances[case.location] += case.get_balance_due()
+
+            if all(
+                [
+                    c.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.ELIGIBLE_NOW
+                    for c in case.charges
+                ]
+            ):
+                fully_eligible_cases.append(case.case_number)
+            elif any(
+                [
+                    c.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.ELIGIBLE_NOW
+                    for c in case.charges
+                ]
+            ):
+                partially_eligible_cases.append(case.case_number)
+            elif all(
+                [
+                    c.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.INELIGIBLE
+                    for c in case.charges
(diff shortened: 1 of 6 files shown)


src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -0,0 +1,65 @@
+from expungeservice.models.expungement_result import ChargeEligibilityStatus
+from expungeservice.models.record_summary import RecordSummary
+from typing import Dict
+
+
+class RecordSummarizer:
+    @staticmethod
+    def summarize(record):
+
```
KentShikama: space

src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -0,0 +1,65 @@
...
+                other_cases.append(case.case_number)
+
+        cases_sorted = {
+            "fully_eligible": fully_eligible_cases,
+            "fully_ineligible": fully_ineligible_cases,
+            "partially_eligible": partially_eligible_cases,
+            "other": other_cases,
+        }
+        for county in county_balances.keys():
+            county_balances[county] = round(county_balances[county], 2)
```
KentShikama: `for key, value in dict.items():`

src/backend/expungeservice/models/record_summary.py
```diff
@@ -0,0 +1,24 @@
...
+    cases_sorted: Dict[str, List[str]]
+    eligible_charges: List[str]
+    county_balances: Dict[str, float]
+
+    @property
+    def total_balance_due(self):
+        total_in_cents = 0
+        for county in self.county_balances.keys():
+            total_in_cents += round(self.county_balances[county] * 100)
+        return round(total_in_cents / 100.0, 2)
```
KentShikama: Note that at https://github.com/codeforpdx/recordexpungPDX/pull/721/files#diff-739da94a6cb453db19351da6939fa0d5R20, you're adding floats. I've had this discussion with @NickSchimek before but since we're adding floats with 2 decimal places, we will never lose accuracy to the point it matters. The increase in complexity for doing this "multiply, add, and divide/round routine" is probably more likely to cause bugs.

src/backend/expungeservice/models/record_summary.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @property
+    def total_cases(self):
+        return sum([len(self.cases_sorted[key]) for key in self.cases_sorted.keys()])
```
KentShikama:
```suggestion
        return sum([len(cases) for cases in self.cases_sorted.values()])
```

src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -0,0 +1,65 @@
+from expungeservice.models.expungement_result import ChargeEligibilityStatus
```
KentShikama:
```suggestion
from expungeservice.models.expungement_result.ChargeEligibilityStatus import *
```

So that `black` will allow you to keep it all inline. It currently looks unnecessarily spaced out. For example,

```
[c.expungement_result.charge_eligibility.status == ELIGIBLE_NOW for c in case.charges]
```

src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -0,0 +1,65 @@
...
+from expungeservice.models.record_summary import RecordSummary
+from typing import Dict
+
+
+class RecordSummarizer:
+    @staticmethod
+    def summarize(record):
```
KentShikama: We should probably start a habit of adding types to public methods in classes `def summarize(record) -> RecordSummary:`. This allows you to jump to the return value class definition on most IDEs.

src/backend/expungeservice/models/helpers/record_summarizer.py
wittejm: oh, nice. Didn't know that was possible with enums

Issue #364, opened 2019-09-10
Build the backend functionality for GROUPS app component 


wittejm:
This is a fairly low-priority feature and not part of our MVP.

The GROUP field on a User data entry for both API and the database is currently just string field. The app [prototype](http://dev.huntermarcks.net/) includes a Group component for managing "groups" like legal firms or public defenders services. This frontend feature is is part of PR #445 

To expand the Groups feature of the app, we will want to keep some associated data with a Group entity and use a group_id as a foreign key for the database USER.group field.

The GROUP table also needs its own CRUD operations.

Endpoints for the Users resource will need refactoring to refer to Groups via group_id rather than group_name. /api/users/ GET should return both the group_name and group_id. The JSON request data for /api/users/ POST and PUT should identify the user's group by group_id rather than group_name. 

### This task requires:
 - Expand design documentation to clarify the features described here.
 - Create a Group table in the database. 
 - Identify what columns belong in the table. group_id, group_name; anything else? description?
 - Refactor the Users table to use a foreign key for the Group field.
 - Refactor the User endpoints: GET (the db users.read operation), POST (the db users.create op), and PUT (the db users.update op) all need to be refactored to read or update the GROUPS table, i.e. using SQL join operations. 
 - Create endpoints to create, read, and alter Group entries, according to the required usage in the frontend.

wittejm:
I would identify this issue as lower priority than #360 , including the endpoints that would enable Login and User Creation

wittejm:
User account functionality no longer exists hooray!

PR #729, opened 2020-01-16
Add user edit user

grbarker:
A redo of adding two new components (AddUser and EditUser).

Added two new files: AddUser/index.tsx and EditUser/index.tsx

Edited three existing files:

AppRouter - Added AuthenticatedRoutes for the two new components.
User - Added button to take you to the EditUser component when the user clicks on the user's name.
UserList - Added button to take you to the AddUser component.


ForrestLonganecker:
This is an excellent addition to the app! 

Thank you for taking the time to put this together.

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+                    : false
+                }
+                onChange={this.handleChange}
+              />
+            </div>
+            <div className="mb4">
+              <label htmlFor="group" className="db mb2 fw6">
+                Group
+              </label>
+              <input
```
ForrestLonganecker: In the prototype this field looks more like the way you have created it in the `EditUser` component. I am not sure which one makes more sense for the end goal of the application, this may be a good thing to talk about in more detail at the next meetup. 

I would say just make them both the same for now, either `text` or `select`.

hmarcks: I don't think we need to require group for now, but yeah Michael should weigh in. It should be a `select` offering the current groups.

wittejm: Michael isn't concerned with groups as a feature for now. So we can just leave it as is, as a field that exists but is optional.

hmarcks:
I just added a new placeholder issue with a wireframe of the entire process of adding and editing accounts if you'd like more context: #731

src/frontend/src/components/AddUser/index.tsx
wittejm: And @hmarcks is right: when we make groups an actual feature it needs to have its own defined entity rather than a free string input field, including the backend support: #364

src/frontend/src/components/User/index.tsx
```diff
@@ -11,7 +12,9 @@ class User extends React.Component<Props> {
       <tr className="bt b--black-20">
         <td className="pa3">
           <a href="/admin" className="underline">
-            {this.props.user.name}
+            <button onClick={() => history.push('edit-user')}>
```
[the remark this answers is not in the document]
grbarker: Thanks for catching that. I'll remove the `/admin` from the a tag.

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -0,0 +1,248 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { Link } from 'react-router-dom';
+import { AppState } from '../../redux/store';
+import { UserState } from '../../redux/users/types';
+import validateEmail from '../../service/email-validation';
+import Logo from '../Logo';
```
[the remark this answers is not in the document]
grbarker: I removed the logo from the AddUser to match up with the prototype.

wittejm:
Awesome! Just be sure to sync and rebase with master again with `git pull -r upstream master`

wittejm:
hey @grbarker how's this going? Could you get it merged soon?

wittejm:
merged. Thanks for finishing this @grbarker!

PR #732, opened 2020-01-17
Add Make targets to deploy to our test url on Heroku

wittejm:
URL: https://recordexpungpdx-test.herokuapp.com

### DIFF
--- a/config/nginx/recordexpunge-nginx.staging.conf
+++ b/config/nginx/recordexpunge-nginx.staging.conf
@@ -1,0 +1,18 @@
+server {
+    listen $PORT default_server;
+    charset c.utf-8; #matches what is defined for the flask service env
+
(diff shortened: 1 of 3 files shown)


config/nginx/recordexpunge-nginx.staging.conf
```diff
@@ -0,0 +1,18 @@
+server {
+    listen $PORT default_server;
+    charset c.utf-8; #matches what is defined for the flask service env
+
```
KentShikama: Please put a redirect block for https. I believe the code is as follows:

```
    if ($scheme = http) {
        return 301 https://$server_name$request_uri;
    }
```

wittejm: It is not, but this shit shall be solved.

Issue #680, opened 2020-01-02
Transfer landing page and move login


hmarcks:
The landing page to promote our technology is already built here: https://codepen.io/hmarcks/full/GRgvaga. Take the HTML & CSS and transfer it to our project. The new css is highlighted in the comments.

Our current login view is at https://recordexpungpdx.herokuapp.com - the new landing page should be located here, and the login page moved to /login.

### This task is done when:
 - The landing page is on https://recordexpungpdx.herokuapp.com
 - The login view is at /login

PR #741, opened 2020-01-19
Added landing page

ForrestLonganecker:
Added the landing page, reference #680


wittejm:
thanks @hmarcks for making the landing page, it looks great.

hmarcks:
Thanks for taking this on Forrest!

The footer with "Terms of use" is showing when viewing the landing page - I'm thinking we should hide this when not logged in, especially since we don't have the content for that yet.

ForrestLonganecker:
Thanks for catching that @hmarcks!

PR #742, opened 2020-01-19
Major refactor of Expunger and Time Analyzer

KentShikama:
Paired with @wittejm and @aemrich

We are changing the logic of the Time Analyzer to more closely reflect how the law would be analyzed by a human. Instead of trying to analyze all charges at once, we loop over each charge and see how other charges in the record affect the charge in question. This allows us to remove concepts such as "second most recent conviction/dismissal" that used to be difficult to analyze. Additionally, now we build up eligibility dates such that each charge now has an set of associated eligibility dates. As of now, we end up taking the `max` and discarding the rest, but in the future this information could be passed through to the frontend, which would make our users more confident about the logic of the program.

Note that the test_expunger.py has been deleted as tested internal properties that no longer exist. A lot of what test_expunger.py covered is duplicated in test_time_analyzer.py. However, I've added a TODO of tests that I believe are not duplicated.

While I am still receiving confirmation from @wittejm, I believe this refactor has caught 2 bugs: https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/tests/expunger/test_time_analyzer.py#L64 and https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/tests/functional_tests/test_crawler_expunger.py#L192. While there is an argument to be made that this refactor should not be merged until the above TODO of tests are completed, I believe the pros of the increased readability, the 2 bugs mentioned, and the probability that this change will have less unknown issues than before justify the merge.


wittejm:
Re @KentShikama  's test case note on https://github.com/codeforpdx/recordexpungPDX/blob/f5ac32be0cbe55b56734e15651cdba9ad8da6165/src/backend/tests/functional_tests/test_crawler_expunger.py#L192 
The test case `test_expunger_runs_time_analyzer(record_with_specific_dates):` is ... also totally confusing and busted? `CaseDetails.case_x()` gives us a record of 800-level traffic charges which are B felonies. The statute and charge level (both used to determine charge type) are either function args for the fixture or hard-coded in the xml string. But as far as I can tell they're all the same: 800-level + B felony.

In the test case, all of these are non-convictions except one:, `cases[0].charges[1]`. On case 0, the dismissals should be type-eligible and the conviction is type- ineligible. (under 137.225 (7) (b)

But everything here should be time-ineligible because there is an arrest from under 3 years ago on the record.

Does this look right?

KentShikama:
@wittejm All charges are 800 level + B Felony but the way we handle our charge classification (https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/helpers/charge_classifier.py#L37-L39) makes it such that it just is an 800 level charge (with level as felony but not class b), which might be a issue, but is out of scope of this PR. Hence, just think all charges are different 800 level charges.

Case 0, charge 1 is time eligible because 8a doesn't apply to convictions. Case 0, charge 0 is also time eligible because case 0 has only one conviction, and thus is time eligible at the same time by the "friendly rule". So the arrests from case 1 end up blocking nothing in case 0. 

Case 2 should all be time ineligible because of 8a applying from case 1.

Issue #705, opened 2020-01-08
Classify a disposition using additional "Disposition events" when necessary


wittejm:
There are a few dispositions like "Conditional Discharge" that do not specify whether the charge was convicted or dismissed. Sometimes the dismissal/conviction info is available instead in additional  "Disposition events" that are tied to the charge, such as  "Amended Disposition"

### This task is done when:
 - When the disposition on a charge gets classified, and the main Disposition event is unrecognized, parse the additional events on the record to find a Dismissal or Conviction event, and use that to classify the disposition.
 -

wittejm:
Completed with PR #758  ; issue #761 describes a further enhancement for this issue.

PR #758, opened 2020-01-24
Amended Disposition gets parsed as a disposition

wittejm:
This fixes the common bug that 'conditional discharge', et al., are not readable as a final disposition

If both "disposition" and "amended disposition" events are on a record for a single charge, the last one in order is the only one returned by the crawler. Amended dispos come later in the OECI page of events, so they take precedence.

The frontend now shows "Amended" if it's showing the disposition result of an amended event

### DIFF
--- a/src/backend/expungeservice/crawler/parsers/case_parser.py
+++ b/src/backend/expungeservice/crawler/parsers/case_parser.py
@@ -108,13 +108,13 @@ class CaseParser:
             else:
                 raise ValueError("len(event_table_contents) should always be 4 or 5.")
 
-            if CaseParser.__normalize_text(event_type.text) == "disposition":
+            if CaseParser.__normalize_text(event_type.text) in ["disposition", "amendeddisposition"]:
                 disposition_data = {}
                 for row, next_row in zip(event_inner_table_parse, event_inner_table_parse[1:]):
                     if CaseParser._valid_data(row.split(".\xa0")):
                         charge_id_string, charge = row.split(".\xa0")
                         charge_id = int(charge_id_string)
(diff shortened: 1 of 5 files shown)


wittejm:
This addresses #705 . It might be a full solution but I'll confirm with Michael if there are other disposition events that need to be inspected other than "Amended Disposition"

src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -108,13 +108,13 @@ def __parse_event_table(event) -> Optional[Dict[int, Dict[str, str]]]:
...
 
-            if CaseParser.__normalize_text(event_type.text) == "disposition":
+            if CaseParser.__normalize_text(event_type.text) in ["disposition", "amendeddisposition"]:
                 disposition_data = {}
                 for row, next_row in zip(event_inner_table_parse, event_inner_table_parse[1:]):
                     if CaseParser._valid_data(row.split(".\xa0")):
                         charge_id_string, charge = row.split(".\xa0")
                         charge_id = int(charge_id_string)
-                        disposition_data[charge_id] = {"date": date.text, "charge": charge, "ruling": next_row}
+                        disposition_data[charge_id] = {"date": date.text, "charge": charge, "ruling": next_row, "event": event_type.text}
```
KentShikama: Does this rely on the fact that the amended disposition always comes after the disposition entry?

wittejm: Yep. It seems like this reliably happens because the events are listed in chronological order. This approach fixes the existing bug reports for this issue, though we could make something more sophisticated later.

Issue #719, opened 2020-01-13
Update case header in results


wittejm:
The case header information in the search results page needs to be updated to reflect the current app design:
http://dev.huntermarcks.net/search/

Namely, field names and values for each case should appear on separate lines. Also this header should include the County name provided for the charge, which is in the `case.location` field of the json object provided by the API.

hmarcks:
<img width="1043" alt="Screen Shot 2020-01-14 at 8 40 36 PM" src="https://user-images.githubusercontent.com/6887982/72400231-3961c500-370e-11ea-8d1d-bb0a6af64d59.png">

keppel2:
Closing to redo branch.

keppel2:
I closed issue instead of pr, oops.

PR #766, opened 2020-01-25
Update case header

keppel2:
Issue #719 update case header to have two lines and display county.


keppel2:
The `span` was replaced with `div` to directly match the static design.

keppel2:
Having a `div` makes the two lines block level, and thus on two lines.

PR #773, opened 2020-01-27
Add extension to friendly rule

KentShikama:
When there are 2 violations, we want to set the arrests to have the time eligibility of the older violation because that will always become eligible first. Similarly when there are 3 violations, we want to set the arrests to have the time eligibility of the second oldest violation. Note we only set the time eligibility to the attractor if the attractor has a earlier time eligibility than the arrests.

### DIFF
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -86,12 +86,22 @@ class Expunger:
 
             charge.set_time_eligibility(eligibility_dates)
         for case in self.record.cases:
-            convictions_in_case = [charge for charge in case.charges if charge.convicted()]
-            if len(convictions_in_case) == 1:
+            non_violation_convictions_in_case = []
+            violations_in_case = []
+            for charge in case.charges:
+                if charge.convicted():
+                    if "violation" in charge.level.lower():
+                        violations_in_case.append(charge)
+                    else:
+                        non_violation_convictions_in_case.append(charge)
+            violations_in_case.sort(key=lambda charge: charge.disposition.date, reverse=True)
+            convictions = non_violation_convictions_in_case + violations_in_case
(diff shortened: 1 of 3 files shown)


aemrich:
Looks good to me. Also, do we have anything that'll stop us from accidentally merging WIP PRs?

KentShikama:
Not really but at the same time I'm not so concerned as in general the PR creator does the merge.

wittejm:
The friendly rule cases (here, a "conviction" refers to a non-violation conviction) : if a case contains a single conviction; or a conviction and a violation; or one, two, or three violations; then the arrests become eligible according to the following:

- a single Conviction: arrests are eligible when that conviction is eligible
- C and a single violation: arrests are eligible when that conviction is eligible
- one violation:  arrests are eligible when that violation is eligible
- two violations:  arrests are eligible when the older one is eligible is eligible
- three violations:  arrests are eligible when the second-oldest one is eligible is eligible

KentShikama:
This PR still requires a reason string of "The friendly rule" to be inserted.

src/backend/expungeservice/expunger.py
```diff
@@ -86,14 +87,57 @@ def run(self) -> bool:
...
+            non_violation_convictions_in_case = []
+            violations_in_case = []
+            for charge in case.charges:
+                if charge.convicted():
+                    if "violation" in charge.level.lower():
+                        violations_in_case.append(charge)
+                    else:
+                        non_violation_convictions_in_case.append(charge)
+            violations_in_case.sort(key=lambda charge: charge.disposition.date, reverse=True)
+            if len(non_violation_convictions_in_case) == 1:
```
wittejm: include `and len(violations_in_case) <= 1`
don't worry about having a test.

wittejm: but  post a new issue for the test case

PR #775, opened 2020-01-29
Add logic for 137.225(1)(c) to expunger

KentShikama:
~~Assuming that convictions within the same case are processed around the same time, if we just assume that the probation revoked applies to all convictions, then the eligibility date of all charges are roughly the same as if the probation revoked only applied to a single conviction: because of the 10 year blocking rule for multiple convictions, the eligibility date of some charges will only be off by the difference in processing times.~~

### DIFF
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -12,13 +12,6 @@ from expungeservice.models.record import Record
 
 
 class Expunger:
-    """
-    The TimeAnalyzer is probably the last major chunk of non-functional code.
-    We mutate the charges in the record directly to add time eligibility information.
-    Hence, for example, it is unsafe to deepcopy any elements in the "chain" stemming from record
-    including closed_charges, charges, self.charges_with_summary.
-    """
-
     def __init__(self, record: Record):
         self.record = record
         self.analyzable_charges = Expunger._without_skippable_charges(self.record.charges)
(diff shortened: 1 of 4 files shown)


src/backend/expungeservice/expunger.py
```diff
@@ -60,6 +60,14 @@ def run(self) -> bool:
                     (charge.disposition.date + relativedelta(years=1), "Time-ineligible under 137.225(1)(b)")
                 )
 
+            if charge.convicted() and charge.case()().get_probation_revoked():
+                eligibility_dates.append(
+                    (
+                        charge.disposition.date + relativedelta(years=10),
```
wittejm: Why does this use conviction date instead of the parsed revocation date?

Relatedly, looking at the fixtures and this parser code it's not clear where revoked probation date actually comes from in the case detail page. Can only a single revocation apply to an entire case? In which case the eligibility dates of all convictions would be the same and the extra reason comment below is not applicable

src/backend/expungeservice/expunger.py
```diff
@@ -60,6 +60,14 @@ def run(self) -> bool:
...
+                        "Time-ineligible under 137.225(1)(c). See comments for when there are multiple convictions in a case.",
```
wittejm: This looks like a placeholder, but a reason string should be readable by a user who doesn't need/want to consult in-code content.

Issue #749, opened 2020-01-21
Conditionally render "Admin" link in navbar


ForrestLonganecker:
This issue is done when:
* We keep track of user admin status in the front end.
* Render the `Admin` link when `isAdmin = true`.
* Upon an `unauthorized` response from an admin protected endpoint the variable is set back to `isAdmin = false`.

See discussion in #731

ARodriguezHacks:
cc @JuanCaicedo

JuanCaicedo:
@ForrestLonganecker a couple questions for you 😀

1. @ARodriguezHacks and I noticed that the `Search` nav bar link leads to an error if  you are not authorized (which we took to mean if the user is not an admin?). Should we hide that link for non-admins as well?

2. This ticket only mentions hiding the link, but we were thinking that the admin route would still be accessible if you type it directly into the url. Should we try to fix that? If so, here's my thinking on how we could do it
  - Create a function that will check if a user `isAdmin` and allow us to redirect them if they are not
  - Apply this function at the router level so that page views will get automatically rerouted

ForrestLonganecker:
Thanks for reaching out @JuanCaicedo & @ARodriguezHacks!

1. The `Search` link is watching for the `OECI` authentication so it is currently acting the way that it should. There will be two different types of users at this point: `admin` or `search` everyone can search, but only `admins` can create new users or edit existing users. Thank you for clarifying this behavior!

2. Adding a redirect at the router level is a good idea, however there is another security concern that we will need to address at the page level as well. A react savvy person could use the react developer tools and flip the `isAdmin` flag to `true` in the dev-tools. having the redirect at the router level would be nice but we will also need to have something in the "protected" components or pages themselves that catches a `unauthorized` response from the backend and flips the `isAdmin` flag back to false/ redirect to a more appropriate page.

The most secure/seamless option would be including an endpoint that specifically returns a response stating admin/not-admin upon attempting to navigate to a protected page but at this point we are trying to limit the number of requests we are sending to the backend.

Long story short, if you would like to include a redirect at the router level based on `isAdmin` flag that would be a good improvement for now! At some point in the future we will need to implement something that will flip the flag back to false upon an unauthorized request/response.

If you have any additional questions please reach out, I really appreciate you two looking into this issue!!

ForrestLonganecker:
There is also a little more information/discussion in issue #731

PR #783, opened 2020-01-31
Fix Navbar to conditionally render Admin link

ARodriguezHacks:
PR for Issue #749 Conditionally render "Admin" link in navbar
https://github.com/codeforpdx/recordexpungPDX/issues/749


src/frontend/src/components/Header/index.tsx
```diff
@@ -7,10 +7,12 @@ import { connect } from 'react-redux';
 
 interface Props {
   isAuthenticated: boolean;
+  isAdmin: boolean; // Added 'isAdmin' for conditional rendering to work (line 35/36)
 }
 
 class Header extends React.Component<Props> {
   public render() {
+    console.log(this.props);
```
KentShikama: This looks like left over debugging statements.

src/frontend/src/redux/system/reducer.ts
```diff
@@ -1,22 +1,26 @@
...
-  loggedIn: false
+  loggedIn: false,
+  isAdmin: false
 };
 
 export function systemReducer(
   state = initialState,
   action: SystemActionTypes
 ): SystemState {
+  console.log('action', action);
```
KentShikama: This looks like left over debugging statements.

src/frontend/package.json
```diff
@@ -20,13 +20,13 @@
...
     "moment": "^2.24.0",
-    "node-sass": "^4.13.0",
+    "node-sass": "^4.13.1",
     "prettier": "^1.19.1",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-redux": "^7.1.3",
     "react-router-dom": "^5.1.2",
-    "react-scripts": "3.2.0",
+    "react-scripts": "^3.3.0",
```
KentShikama: While these probably need to be updated, I would prefer if they are done in a separate PR. You can use 

```
git reset HEAD~1 -- src/frontend/package.json
git reset HEAD~1 -- src/frontend/package-lock.json
```

and commit that.

src/frontend/src/components/Header/index.tsx
```diff
@@ -30,12 +32,15 @@ class Header extends React.Component<Props> {
             >
               Stats
             </button>
-            <button
-              onClick={() => history.push('/admin')}
-              className="link mid-gray hover-blue f6 f5-ns dib pa3"
-            >
-              Admin
-            </button>
+            {/* Conditionally render the admin button based on isAdmin from the props */}
```
KentShikama: In general I prefer locally scoped comments like this one only if the code does something that is surprising giving the context or unusually unreadable.

src/frontend/src/components/Header/index.tsx
```diff
@@ -62,7 +67,9 @@ class Header extends React.Component<Props> {
 }
 
 const mapStateToProps = (state: AppState) => ({
-  isAuthenticated: state.system.loggedIn
+  isAuthenticated: state.system.loggedIn,
+  isAdmin: state.system.isAdmin
+  //Do the same as above but with 'isAdmin'
```
KentShikama: What does "above" refer to here?

src/frontend/src/redux/system/reducer.ts
```diff
@@ -1,22 +1,26 @@
...
-        loggedIn: true
+        loggedIn: true,
+        isAdmin: action.isAdmin
       };
     }
     case LOG_OUT: {
       return {
-        loggedIn: false
+        loggedIn: false,
+        isAdmin: false
```
KentShikama: `isAdmin` should not exist for log out.

src/frontend/src/components/Header/index.tsx
```diff
@@ -7,10 +7,12 @@ import { connect } from 'react-redux';
 
 interface Props {
   isAuthenticated: boolean;
+  isAdmin: boolean; // Added 'isAdmin' for conditional rendering to work (line 35/36)
```
KentShikama: Comments like these become quickly outdated (especially line numbers). The commit message is probably a better place for this comment.

src/frontend/src/redux/system/reducer.ts
ARodriguezHacks: I'm not sure how to resolve this issue. Tried to remove it bu received an error: "Property 'isAdmin' is missing in type '{ loggedIn: false; }' but required in type 'SystemState'."

KentShikama: https://github.com/codeforpdx/recordexpungPDX/pull/783/files#diff-076570d447b7ed9a8929b014215fae08R3 makes it required. You could change it to `isAdmin?: boolean;`. If for whatever reason that doesn't work, then keeping line 22 above is fine but we probably should add it to the types at https://github.com/codeforpdx/recordexpungPDX/pull/783/files#diff-076570d447b7ed9a8929b014215fae08R14 too.

wittejm: It looks to me like `isAdmin` can't be made optional here so long as it's a required field in SystemState (which is good) , so apart from Kent's alternate suggestion here this PR looks good.

ARodriguezHacks: If only I saw this comment three minutes ago. Oh well, I already changed it to reflect Kent's suggestion, and I checked that the log in and log out were still working properly.

Issue #746, opened 2020-01-21
Connect "AddUser" component to the backend


ForrestLonganecker:
This task is done when:

* Admin user is able to create a new user by submitting a new user form to the back end.
* Form validates that the new user form contains:
  * Email
  * Password
* Non-Admins are unable to create a new user.
* Error messages match required fields

hmarcks:
![account-flow](https://user-images.githubusercontent.com/6887982/72775070-96f48680-3bd2-11ea-8a1a-ea2bc733a1d6.png)

Accessible validation technique: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19

grbarker:
@ForrestLonganecker I wanted to take a go at adding form validation to the AddUser component. I was going to open a new issue for it. I wanted to clarify what the validation should do. 

You mentioned that email and password need to be confirmed to exist. I spoke with @wittejm about the validation and he wanted to include the name field and a new confirm password field in the validation.

You also mentioned that the error messages should match the required fields. Does this mean that you want each field to have its own separate alert if they are missing, or can we use one alert to inform the user that one of the required fields is missing?

ForrestLonganecker:
Thanks for taking this on @grbarker!
Looking at the error messages here:
![image](https://user-images.githubusercontent.com/42503418/73112567-c1408f80-3ec3-11ea-9e0a-47fbee9b8bcc.png)
I think we would only need to add validation messages about password being required and the password/password confirmation needing to match. You can use the same error message format that is used elsewhere but just swap out the text.

Please reach out with any additional questions!

grbarker:
@wittejm wanted to validate that the password is 8 characters long so I put that in as well.

The email validation function catches an empty email and gives a "Invalid Email" alert. Do we want to explicitly say that it is missing like the other fields or just keep on going with the "Invalid Email" catch all alert?

ForrestLonganecker:
I feel `Invalid email address` should suffice for both: No email present, and incorrect format

grbarker:
I'll leave the email validation alone then. 
Just to make 100% sure I we are on the same page, are you wanting the validation in the screenshot as well as as the validation of email and password?

ForrestLonganecker:
Yes please!

wittejm:
completed with #889

PR #791, opened 2020-02-04
Add form validation and matching error fields to satisfy part of issue #746

grbarker:
This PR addresses part of issue #746. It adds form validation to the component and validation error alerts that match the corresponding field.


src/frontend/src/components/AddUser/index.tsx
```diff
@@ -46,18 +82,56 @@ class AddUser extends React.Component<Props, State> {
     event.preventDefault();
 
     // Standardize email-lowerCase and trim both form inputs
+        // Standardize name-lowerCase
     this.setState(
       {
+        name: this.state.name.toLowerCase().trim(),
```
wittejm: We should leave the `name` field as case-sensitive, so this `toLowerCase()` can be removed.

grbarker: Ok I blew away the master adn synced with upstream. I made new branch, added the changes, and pushed it to my repo.  I'll make a pr from there. Should I close this pr?

grbarker: And thanks for the help

wittejm: Yep, you can close this PR.

PR #796, opened 2020-02-05
Add form validation to AddUser component.

grbarker:
This PR addresses part of issue #746. It adds form validation to the component and validation error alerts that match the corresponding field.

### DIFF
--- a/src/frontend/src/components/AddUser/index.tsx
+++ b/src/frontend/src/components/AddUser/index.tsx
@@ -12,26 +12,32 @@ interface Props {
 interface State {
   email: string;
   password: string;
+  confirmPassword: string;
   name: string;
   group: string;
   role: string;
-  invalidCredentials: boolean;
   invalidResponse: boolean;
+  missingName: boolean;
   invalidEmail: boolean;
   missingPassword: boolean;
+  invalidPassword: boolean;
+  mismatchPasswords: boolean;
 }
 
 class AddUser extends React.Component<Props, State> {
   public state: State = {
     email: '',
     password: '',
+    confirmPassword: '',
     name: '',
     group: '',
     role: '',
-    invalidCredentials: false,
     invalidResponse: false,
+    missingName: false,
     invalidEmail: false,
-    missingPassword: false
+    missingPassword: false,
+    invalidPassword: false,
+    mismatchPasswords: false,
   };
 
   public handleChange = (e: React.BaseSyntheticEvent) => {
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/AddUser/index.tsx
```diff
@@ -48,16 +56,31 @@ class AddUser extends React.Component<Props, State> {
     // Standardize email-lowerCase and trim both form inputs
```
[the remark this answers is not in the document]
grbarker: The comments are not mine. I'll remove this one, but I'm sure there are others.

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -71,20 +94,20 @@ class AddUser extends React.Component<Props, State> {
                 Name
               </label>
               <input
-                id="username"
+                id="name"
                 name="username"
```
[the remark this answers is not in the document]
grbarker: Ok great, I wanted to change them so they matched.

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -127,14 +152,18 @@ class AddUser extends React.Component<Props, State> {
                 className="w-100 pa3 br2 b--black-20"
                 required={true}
                 aria-describedby={
-                  this.state.invalidCredentials
+                  this.state.missingPassword
+                    ? 'passwd_input_msg'
```
ForrestLonganecker: I think using `password` instead of `passwd` in all instances of `passwd` would be a better practice. I am realizing that we are also using `msg` instead of `message` and am thinking we should do the same for that. These fields are being set for accessibility so it would be good to have them as readable as possible.

We will eventually want to change the `..._msg` entries in other components using a similar wording, but that will be saved for a different issue.

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -12,26 +12,38 @@ interface Props {
...
 }
 
 class AddUser extends React.Component<Props, State> {
   public state: State = {
     email: '',
     password: '',
+    confirmPassword: '',
     name: '',
     group: '',
+    search: '',
```
ForrestLonganecker: `search` and `admin` are possible values for the `role` variable, ie: `role = 'search'` or `role = 'admin'` so they can be removed from state.

src/frontend/src/components/AddUser/index.tsx
grbarker: Thanks for the feedback. I'll make the requested changes.

Issue #722, opened 2020-01-15
Provide multiple search criteria in one search request


wittejm:
A single person can have records that only appear in OECI under distinct aliases, and we'd like to have all of that person's records retrievable into a combined search result.

The frontend should have the option to repeatedly add additional search criteria, e.g. with a small [add] button, that expands the search panel with a copy of the input fields on a new line.

The backend will need changes in the /api/search endpoint and/or in the crawler to:
- crawl for each of the different search criteria
- accumulate all the returned cases into a single record without duplicates
- run the expunger on the entire record

The displayed results could somehow convey which cases are tied with which search criteria, either in their case headers or by being grouped and labeled together

wittejm:
Whoever works on the frontend portion of this will need to sync with any changes reflected in #724

wittejm:
@hmarcks With the new middle name field added and the search params bumped out to two lines, I'm not sure what this new feature should look like. What do you think?

hmarcks:
@wittejm I've been meaning to create a design for this, I'll have one soon.

wittejm:
sounds good.

hmarcks:
@wittejm what do you think of something simple like this?

![multiple-searches](https://user-images.githubusercontent.com/6887982/73131262-832f9280-3fcd-11ea-9ed0-229539304a21.png)

wittejm:
This looks really good, though I was thinking of a slightly simpler flow for now especially since we want to get this feature into the March milestone release. What I had in mind was to optionally put in extra sets of search params when beginning the search, and still run it in one  button click. If someone searches once and wants to add results from another search, they would just have to put in all their desired params and hit the search button again.

An example of adding optional additional params for one request is the Slack send invitations UI, like here (mobile lawl)

![Screenshot_20200126-131058](https://user-images.githubusercontent.com/2104990/73142057-4ac89c80-403f-11ea-9350-fad1cfc3725c.png)

I think this approach would be the simplest new feature and require very little change in how we handle frontend state changes.

Sorry, I could have been more clear on this especially with the descriptionless related issue #724 that I posted.

I think having UI features that are more interactive for adjusting the analyzed results like this and your other feature mockup #687 would potentially be really great if more involved to build, and we'll have to see how much benefit Michael or our other forthcoming users would see in them.

hmarcks:
Ah got it, I’ll post a design with this approach soon

hmarcks:
[...]

PR #802, opened 2020-02-06
Add the middle name field for record search

### LINKED ISSUE #726: Add the middle name search field to frontend
And include it in the endpoint call. Endpoint code already handles it as an optional field.

wittejm:
I chose to keep all the search fields on one line despite the mockup change, since the fields are still plenty large enough when set to 25% width.

The button elements for add/remove extra aliases are included but hidden, since the formatting I used here is meant to account for the aliases feature.

A refactor to abstract out a `Field` component for the four fields is forthcoming.

Closes #726. Is a dependency for #722

### DIFF
--- a/src/frontend/src/components/RecordSearch/index.tsx
+++ b/src/frontend/src/components/RecordSearch/index.tsx
@@ -82,7 +82,7 @@ class RecordSearch extends React.Component<Props, State> {
         <h1 className="mb4 f4 fw6">Record Search</h1>
         <form onSubmit={this.handleSubmit} noValidate>
           <div className="flex flex-wrap items-end">
-            <div className="w-100 w-25-ns mb3 pr2-ns">
+            <div className="w-100 w-50-m w-25-ns mb3 pr2-ns">
               <label htmlFor="firstName" className="db mb1 fw6">
                 First Name
               </label>
@@ -98,7 +98,7 @@ class RecordSearch extends React.Component<Props, State> {
                 onChange={this.handleChange}
               />
             </div>
-            <div className="w-100 w-25-ns mb3 pr2-ns">
+            <div className="w-100 w-50-m w-25-ns mb3 pr2-ns">
               <label htmlFor="middleName" className="db mb1 ">
                 <span className= "fw6"> Middle Name </span> <span className= "fw1">(Optional) </span>
               </label>
@@ -110,7 +110,7 @@ class RecordSearch extends React.Component<Props, State> {
                 onChange={this.handleChange}
               />
             </div>
-            <div className="w-100 w-25-ns mb3 pr2-ns">
+            <div className="w-100 w-50-m w-25-ns mb3 pr2-ns">
               <label htmlFor="lastName" className="db mb1 fw6">
                 Last Name
               </label>
@@ -126,7 +126,7 @@ class RecordSearch extends React.Component<Props, State> {
                 onChange={this.handleChange}
               />
             </div>
-            <div className="w-100 w-25-ns mb3 pr2-ns">
+            <div className="w-100 w-50-m w-25-ns mb3 pr2-ns">
               <label htmlFor="dateOfBirth" className="db mb1 fw6">
                 Date of Birth <span className="fw2 f6">MM/DD/YYYY</span>
               </label>
--- a/src/frontend/src/components/RecordSearch/index.tsx
+++ b/src/frontend/src/components/RecordSearch/index.tsx
@@ -78,99 +78,109 @@ class RecordSearch extends React.Component<Props, State> {
 
   public render() {
     return (
+      // TODO: verify Tachyon differences btwn this section tag and <main> in prototype
       <section className="cf mt4 mb3 pa3 pa4-l bg-white shadow br3">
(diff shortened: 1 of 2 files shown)


wittejm:
I didn't include the `aria` property that is present on the first/last names, since middle name is not required. I'm not totally sure if there's anything else that should be different between the middle name and the others. A question for @ForrestLonganecker in particular!

hmarcks:
@wittejm , it gets pretty tight if all the fields are on a single line for narrower screen widths. I don't think it's a huge deal, but ask that you consider using the markup provided in PR #789 for #722 (multiple searches).

Screen Shot 2020-02-06 at 8 35 12 PM

And yeah if middle doesn't have an error associated with it then the aria can be removed.

wittejm:
@hmarcks gotcha, I wasn't considering narrower screens. I will update accordingly!

wittejm:
@hmarcks so I'm trying to work around the dangling field on the second row, and working with @ForrestLonganecker I found this alternate solution of applying a responsive "middle-width" rule which will bump the input fields to two rows. Varying screen widths it now looks like this:

Wide screen:
wide_fields

Narrowest resolution that allows a single row of inputs:
narrow-full-wdith

Narrowest resolution that allows for two rows of inputs:

narrow-middle-width

Small screens: (from the the pre-existing `w-25-ns` rule) 
narrow

What do you think?

wittejm:
And here's what it looks like with the buttons to add / remove aliases, bumping them down to new lines. I also renamed the "+ Search" button to "+ Add Alias" which I think is more intuitive.


1 alias with alias button

After adding an alias:

with more aliases

Same thing, on n a middling-width screen:

medium 2 aliases

Showing multiple aliases on a narrow screen:

narrow-more-aliases

wittejm:
Actually The 2x2 arrangement looks bad because the "name" fields are split on the two rows. I think you had the right idea after all with dropping just birthday down, especially once the "+Add alias" and " X Remove" buttons snuggle in there. 

I still lean toward keeping the input fields on a single row if the screen is wide enough to comfortably hold them.  What do you think of this, if the narrow-squeeze issue is handled by bumping to two rows (just moving the birth date) in the "medium-screen size" responsive configuration?

hmarcks:
I agree it's best if we can get all fields on one line on wider screens. I think it was a bit tricky with the breakpoints that Tachyons provides but I can create a custom one. Let me take a look...

wittejm:
Sounds good!

hmarcks:
@wittejm I updated the search layout to be two rows total with the alias button, see it at https://github.com/hmarcks/expunge-pdx-frontend/blob/master/search/index.html

HTML: https://github.com/hmarcks/expunge-pdx-frontend/blob/master/search/index.html

wittejm:
@hmarcks This looks great! thank you!

hmarcks:
@wittejm thanks! Feel free to pass it off to me after you implement the functionality. I restructured things a bit and there were a lot of class name updates...

wittejm:
Ok! I @hmarcks applied your markup and have the middle name field functional. This current commit includes the Alias, # Results, and Remove buttons though they will be hidden before the final merge, since their functionality is part of #722 

Looks like:
wide_middle_name


squeezed_middle_name

I think I got all your style changes copied in but let me know if you see any issues.

Re-Pinging @KentShikama @ForrestLonganecker @nickprender12 for another pair of review eyeballs.

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -75,71 +78,113 @@ class RecordSearch extends React.Component<Props, State> {
...
+              </div>
+              <div className="w-100 w-third-ns w-25-l mb3">
+                <label htmlFor="middleName" className="db mb1 fw6">
+                  Middle Name<span className= "fw2 f6">(Optional) </span>
+                </label>
+                <input
+                  id="middleName"
+                  type="text"
+                  className="w-100 pa3 br2 b--black-20"
+                  required
```
KentShikama: What does required mean here?

wittejm:
@KentShikama ah good catch. Looks like that provides a pure-html version of form validation, but putting `noValidate` on the `<form>` tag makes them not do anything. I just tried it with all the `required` and `noValidate` attributes removed and it works the same way. Seems to me like they should stay gone. @ForrestLonganecker what do you think?

hmarcks:
Taking a look now. Regarding required attributes, even though we are using noValidate I think we still want them for screen reader support if I understand this article correctly: https://developer.paciellogroup.com/blog/2019/02/required-attribute-requirements/

wittejm:
gotcha

hmarcks:
@wittejm do you know if I could push directly to this pull request? If not, [my updates are here](https://github.com/hmarcks/recordexpungPDX/commit/94b33d1144e84dd13dc43a4c48c3c3c58c6a026e).

wittejm:
Yea I'm pretty sure you'll be able to push directly to this branch. That's my default setting, I'm not 100% sure but you should give it a shot

hmarcks:
Looks like that worked. I didn't add the required attributes back but looks like we should, right?

wittejm:
Awesome. Yea, I'll re-add them.

Issue #797, opened 2020-02-06
Show expanded charge type documentation in search results


wittejm:
Edit: Renaming and updating this issue for the next phase, of adding the frontend feature.

See below for mockups. 

To handle charge types that are essentially split between two classes (`RomeoAndJulietIneligibleSexCrime` and `DUII`), we'll need to address then in a way that both the Manual's charge types page and the entries included in the search results page both look "good" to the user, while the manual page is ideally still easily generated from the text written in the backend.  

---
older, completed phase of writing the expanded documentation.

This will need to be a work in progress as we figure out exactly how to handle detailed explanatory documentation. We want to:
- keep it tightly synced with the source code, since the documentation on our set of Charge Types should purely give explanation for why those charge types work the way they do.
- expose documentation to the user programmatically, and have no redundant doc files for the same content. On search results, there could be e.g. a mouse-over tooltip, or a "show more" button, on the charge type field. We could also have pages in the website's user manual that compile the complete rules of all charge types into one place.  

For now, we can achieve this by writing the explanation for a charge type as a string field on the charge type class object. The python doc module Phoenix relies on comment strings attached to classes and methods (`obj.__doc__`) but we should use a named string field because `__doc__` strings are meant to be developer-facing, not user-facing.

The doc for each charge type can include the following: 
- Reference the expungement statute's subsections that provide eligibility reasons: for its type eligibility, for both conviction and dismissal when there is a difference, and time eligibility if there are special rules like with the Class B felony
- Whenever skip_analysis is true, say why.
- Why do the particular classification conditions apply within the [`ChargeClassifier`](https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/helpers/charge_classifier.py) for that charge type, especially when a charge seems like it could qualify as multiple charge types? A charge always gets exactly one type, and the order of rules getting applied is very specific. It might be useful to provide any extra context on why a charge type can be applied with higher precedence than other types (e.g., Subsection12 names some Felony statutes, so the classifier needs to check Subsection12 with higher priority than the Felony charge types).
- Include anything else generally interesting about charge type that wasn’t already covered.
- There’s some “special knowledge” that Michael has only imparted in Slack either to the general channel, or to some folks like Jordan or Kent. We’ll need to transcribe anything that applies.

The string field on each Charge type can be called `expungement_rules`

wittejm:
[...]

Issue #627, opened 2019-12-11
Refactor the unit tests for the charge models


NickSchimek:
## Description
When the type analyzer was removed/refactored the tests for the type analyzer were not refactored too. Some previous PR's have refactored the tests into their on files for unit testing each of the different charge types, but there are still some charge types remaining.

## Task
Finish refactoring the unit tests for the models

### Additionally
Update the schedule1PCS tests to include the new statute in PR #626

wittejm:
Hi @nickbeaird , I see you've started working on a few issues besides this one as well. How is this one going?

wittejm:
Hi @nickbeaird I'm working on #795 and part of that effort is to update their unit tests which would overlap somewhat with this issue. Do you have progress you could push for this issue? This is just to check if you've already done work that I wouldn't need to repeat for these types.

wittejm:
Ah, I just came across the branch you're working on https://github.com/codeforpdx/recordexpungPDX/tree/627_refactor_tests Looks like you've made a lot of progress! Want to file a PR with what you have so far? I recommend you file PRs incrementally to help us keep our work roughly in sync, and reduce the risk of merge conflicts. 

My PR #813 for Subsection12 rips some old code out of `test_type_analyzer_convicted_charges.py` but other than that I don't think it should cause too much conflict with your progress here.

nickbeaird:
I am posting here to keep everyone up-to-date. I met up with @wittejm today. I had a PR that was open for too long, and had some potential issues, so that branch was deleted from the remote repo, and I am opening a new one now. I will submit the new PR by EOD.

nickbeaird:
Here is the PR for this project. https://github.com/codeforpdx/recordexpungPDX/pull/850. Please feel free to send me any notes in the meantime.

wittejm:
Completed

PR #813, opened 2020-02-09
Fix Subsection12 and add Subsection6 charge types

### LINKED ISSUE #795: Fix charge type Subsection12 and add the new type Subsection6
Our definition for Subsection12 is based on an old list that does not completely match the actual subsection 12 in the statute. The list of statutes in ChargeClassifier should be updated to match. The class can also be labeled as Type Eligible (with reason: Eligible under 137.225(12))

And, we need to define and classify a new charge type, Subsection6 , which is a collection of statutes which are Needs Further Analysis. (with reason: Eligibility conditions are described in 137.225(6))

This is a combined issue because the existing subsection 12 list covers some of subsection 6. It's partially an issue to split an existing charge type into two.  

### This task is done when:
 - New passing unit tests to classify every statute listed in these two sets.

wittejm:
* Subsection12 is fixed to match the set of statutes listed in the expungement law. It has a new expungement_rules string to explain its weirdness. Also the type is now properly set to ELIGIBLE.

 * Subsection6 is introduced as a charge type to capture our out-of-date "List B" type and reflect the actual statutes given in Subsection 6 of the law.

 * Test cases are created for each of the statutes named in these two charge types.

Closes #795 . Makes some progress in extended docstrings for #797 and reorganized tests for charge types #627


src/backend/expungeservice/models/helpers/charge_classifier.py
```diff
@@ -71,25 +73,41 @@ def _marijuana_ineligible(statute, section):
...
+            "164405",  # (Robbery in the second degree)
+            "164395",  # (Robbery in the third degree)
+            "162185",  # (Supplying contraband)
+            "166220",  # (Unlawful use of weapon)
+        ]
+        conditionally_eligible_statutes = [
+            "163525"  # (Incest), conditional that the victim was at least 18 years of age.
         ]
-        if section in ineligible_statutes:
+        if section in conditionally_eligible_statutes or section in eligible_statutes:
```
KentShikama:
```suggestion
        if section in conditionally_eligible_statutes + eligible_statutes:
```

src/backend/expungeservice/models/charge_types/subsection_6.py
```diff
@@ -0,0 +1,30 @@
...
+from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+
+
+@dataclass(eq=False)
+class Subsection6(Charge):
+    type_name: str = "Subsection 6"
+
+    expungement_rules: str = (
+        """
+Applicable subsections: 137.225(6) for convictions; 137.225(1)(b) for dismissals."
```
KentShikama: We should just turn on double line spacing formatting on the resulting doc if we want to have this effect. In fact maybe you should turn on the double line spacing on your editor :P

PR #823, opened 2020-02-10
Provide multiple aliases in search

wittejm:
(edited for latest commit)
This abstracts the main component in RecordSearch into the Field and Alias subcomponents .

The Object array of aliases data is still held in RecordSearch state, and it is synced with the input fields via nested react Props and chained clickHandler calls.

The +Alias and -Remove buttons are fully functional. Multiple alias search works end-to-end.

The backend Crawler().search() has interior side effects that make a second search() call break with a single crawler instance. Rather than fix this issue, I instance a new crawler for each alias search. The crawler needs a general overhaul for multiple alias but this works as is.

for testing: Good names to search for screwing around with the UI are 'a a' (firstname a, lastname a) and 'e e' (firstname e, lastname e) . They have short nonempty records. Convenient.

Todo: insert the `<hr>` element between Aliases when more than one is rendered.

Progress for #722

### DIFF
--- a/src/frontend/src/components/RecordSearch/Field.tsx
+++ b/src/frontend/src/components/RecordSearch/Field.tsx
@@ -5,6 +5,8 @@ interface Props {
   label: string;
   content: string;
   coda: string;
+  divMarkup: string;
+  inputMarkup: string;
   onChange: Function; // arg: value
   errorMessage: string;
   required: boolean;
@@ -18,6 +20,11 @@ export default class Field extends React.Component<Props, State> {
   state: State = {
     hasInput: false
   }
+  public static defaultProps = {
+    coda: "",
+    divMarkup: "",
+    inputMarkup: ""
+    };
 
   handleInputChange = (e: React.BaseSyntheticEvent) => {
     let fieldContent : string = e.target.value
@@ -32,7 +39,7 @@ export default class Field extends React.Component<Props, State> {
 
   render() {
     return(
-      <div className="w-100 w-third-ns w-25-l mb3">
+      <div className={"w-100 w-third-ns w-25-l mb3 " + this.props.divMarkup}>
         <label htmlFor={this.props.name} className="db mb1 fw6">
           {this.props.label} <span className= "fw2 f6">{this.props.coda}</span>
         </label>
@@ -40,7 +47,7 @@ export default class Field extends React.Component<Props, State> {
           value={this.props.content}
           id={this.props.name}
           type="text"
-          className="w-100 b--black-20 br2 br-0-ns br--left-ns pa3"
+          className={"w-100 br2 b--black-20 pa3 " + this.props.inputMarkup}
           required
           aria-invalid={false} // this.state.firstNameHasInput}
           aria-describedby={
--- a/src/frontend/src/components/RecordSearch/Field.tsx
+++ b/src/frontend/src/components/RecordSearch/Field.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+
+interface Props {
+  name: string;
+  label: string;
+  content: string;
+  coda: string;
+  onChange: Function; // arg: value
+  errorMessage: string;
+  required: boolean;
+}
+
+interface State {
+  hasInput: boolean;
+}
+
+export default class Field extends React.Component<Props, State> {
+  state: State = {
+    hasInput: false
+  }
+
+  handleInputChange = (e: React.BaseSyntheticEvent) => {
+    let fieldContent : string = e.target.value
+    this.setState(
+      {
+        hasInput: fieldContent.length > 0
+      }
+      )
+
+    this.props.onChange(e.target.value);
+  };
+
+  render() {
+    return(
+      <div className="w-100 w-third-ns w-25-l mb3">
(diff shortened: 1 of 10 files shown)


src/frontend/src/components/RecordSearch/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface AliasType {
+  statute: string;
```
KentShikama: What is statute?

src/frontend/src/components/RecordSearch/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface AliasType {
```
KentShikama: Why not just Alias?

src/frontend/src/components/RecordSearch/index.tsx
```diff
@@ -53,7 +53,14 @@ class RecordSearch extends React.Component<Props, State> {
...
         let dateOfBirth = state.dateOfBirth.length > 0 ? state.dateOfBirth : '';
-        this.props.fetchRecords(firstName, middleName, lastName, dateOfBirth);
+        this.props.fetchRecords(
+          [{
+            first_name: firstName,
+            middle_name: middleName,
+            last_name: lastName,
+            birth_date: dateOfBirth
+          }]
+            );
```
KentShikama: Was this how husky auto-formatted?

src/frontend/src/components/RecordSearch/types.ts
wittejm: We call every Typescript class a `SomethingType` which seems not great but it's because there are React Components we would want to also call by the same, unless we call those `SomethingComponent` or such

wittejm: I'd be excited about a nicer approach

wittejm:
Updated the PR comment above for latest changes, ready for major review

src/frontend/src/components/RecordSearch/Field.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return(
+      <div className="w-100 w-third-ns w-25-l mb3">
+        <label htmlFor={this.props.name} className="db mb1 fw6">
+          {this.props.label} <span className= "fw2 f6">{this.props.coda}</span>
+        </label>
+        <input
+          value={this.props.content}
+          id={this.props.name}
+          type="text"
+          className="w-100 b--black-20 br2 br-0-ns br--left-ns pa3"
```
hmarcks: Each of the inputs have unique classes to give them the appearance of being connected:

First Name: `w-100 b--black-20 br2 br-0-ns br--left-ns pa3`
Middle Name: `w-100 br2 br0-ns b--black-20 pa3`
Last Name: `w-100 b--black-20 br2 bl-0-ns br--right-ns pa3`
DOB wrapper div: `w-100 w-third-ns w-25-l pl2-l mb3`
DOB: `w-100 pa3 br2 b--black-20`

Screen_Shot_2020-02-14_at_4_43_03_PM-2


Otherwise looking great!

wittejm: ah, gotcha!

src/frontend/src/components/RecordSearch/types.ts
wittejm: removed.

wittejm:
@hmarcks I updated the input fields and enclosing div for bd and it doesn't look quite like yours (though it's a tad closer). 

alias_with_newest_markup
 
Tried fiddling with the tachyon classnames and was not enlightened, though it may be a platform issue. I notice the standalone input field DOB on mine looks different from yours, in the left vs right edge colors.

I haven't done any web compat before, I could take a look at those tools but I'm going to step away for the evening.

wittejm:
oh yea and I'm using ubuntu firefox. chrome looks the same for me

hmarcks:
Yeah browsers / OS will render a bit differently, all the borders are there though 👍

wittejm:
Ok, I don't see any remaining TODOs (still hoping for react review @ForrestLonganecker @KentShikama et al), but @hmarcks I wanted to see if I'm using `required` correctly on input fields. 

I made it `false` for DOB, though there's no behavioral difference between it being `true` or `false`, suggesting it only affects the aria. Your [prototype ](https://github.com/codeforpdx/recordexpungPDX/pull/789/files#diff-c651cf335f1e8423fe825eded56937d8R271) has the DOB input field with the required prop. Is this correct? and if so, what does it do? Does it handle the dob formatting requirement?

hmarcks:
@wittejm my understanding is that the required attribute helps communicate to some assistive technology. [This article](https://developer.paciellogroup.com/blog/2019/02/required-attribute-requirements/) explains how to achieve accessible validation by using `required`, `aria-invalid`, and `aria-describedby`, I'm no expert though.

And the prototype shouldn't have `required` on the DOB field, I probably assumed it was required a while ago.

wittejm:
Cool, thanks for the info and link. I think that Forrest and I might have decided ad hoc to make BDay optional at some point, when we noticed it's optional in OECI.

Issue #819, opened 2020-02-10
Keep the time eligibility date even if it's in the past 


wittejm:
This is to make the underlying time eligibility logic more interpretable by the user.

### This task is done when:
 - New passing unit tests.

Issue #803, opened 2020-02-07
Add or augment 5 test cases for the friendly rule 


wittejm:
The test cases should cover the combinations described in the comments on the following:
https://github.com/codeforpdx/recordexpungPDX/pull/773

wittejm:
will be completed with with #773

wittejm:
will actually be completed with #825

PR #825, opened 2020-02-11
Ensure that date will be eligible is always set

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/819

Previously if the date will be eligible was before today, we would just set it to None. While this is ok, it is more elegant to just set it to the appropriate past date as then we can analyze what would have happened if we tried to expunge it in the past. A bunch of tests that had set the date_will_be_eligible to None were then fixed. While fixing those tests, I noticed that we didn't use the arrest date but instead the disposition date as the starting point for calculating eligibility dates for the dismissals, and so that was taken care of. A previously skipped test related to the friendly rule was fixed and moved to the new test_friendly_rule.py. Since we don't have to worry about date_will_be_eligible being None, we can now use >= comparisons in the Expunger.

### DIFF
--- a/src/backend/tests/test_friendly_rule.py
+++ b/src/backend/tests/test_friendly_rule.py
@@ -1,14 +1,38 @@
+from dateutil.relativedelta import relativedelta
+
 from expungeservice.expunger import Expunger
 from expungeservice.models.expungement_result import EligibilityStatus
 from expungeservice.models.record import Record
 from tests.factories.case_factory import CaseFactory
 from tests.factories.charge_factory import ChargeFactory
-from tests.factories.expunger_factory import ExpungerFactory
 from tests.time import Time
+from datetime import date
+
+
+def test_eligible_mrc_with_single_arrest():
+    three_yr_mrc = ChargeFactory.create(disposition=["Convicted", Time.THREE_YEARS_AGO])
+
+    arrest = ChargeFactory.create(disposition=["Dismissed", Time.THREE_YEARS_AGO])
+
+    case = CaseFactory.create()
+    case.charges = [three_yr_mrc, arrest]
+    record = Record([case])
+    expunger = Expunger(record)
+
+    expunger.run()
+    assert arrest.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE
+    assert (
+        arrest.expungement_result.time_eligibility.reason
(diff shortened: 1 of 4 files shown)


KentShikama:
@wittejm Would you say this resolves https://github.com/codeforpdx/recordexpungPDX/issues/803 too?

src/backend/tests/test_friendly_rule.py
```diff
@@ -20,10 +44,40 @@ def test_arrest_is_unaffected_if_conviction_is_older():
     expunger.run()
 
     assert arrest.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE
-    assert arrest.expungement_result.time_eligibility.date_will_be_eligible == None
+    assert arrest.expungement_result.time_eligibility.date_will_be_eligible == arrest.disposition.date
```
wittejm: I suggest passing a separate charge date for this charge, so that 
`     assert arrest.expungement_result.time_eligibility.date_will_be_eligible == arrest.disposition.date`
fails as it should, but 
`     assert arrest.expungement_result.time_eligibility.date_will_be_eligible == arrest.date`
passes

src/backend/tests/test_friendly_rule.py
```diff
@@ -1,14 +1,38 @@
...
+        == "The friendly rule: time eligibility of the arrest matches time eligibility of the conviction."
+    )
+    assert arrest.expungement_result.time_eligibility.date_will_be_eligible == date.today()
+
+    assert three_yr_mrc.expungement_result.time_eligibility.status is EligibilityStatus.ELIGIBLE
+    assert three_yr_mrc.expungement_result.time_eligibility.reason == ""
+    assert three_yr_mrc.expungement_result.time_eligibility.date_will_be_eligible == date.today()
 
 
 def test_arrest_is_unaffected_if_conviction_is_older():
```
wittejm: This test name is really confusing. what's the actual thing being tested here? Would the arrest be affected if the conviction was newer?

src/backend/tests/test_friendly_rule.py
KentShikama: Unfortunately this requires changing `ChargeFactory.create` and as you know that has cascading effects.

PR #826, opened 2020-02-11
Add form validation to EditUser component.

grbarker:
Validation checks that name, email, and password fields exist, checks
the email is valid, checks that password is valid, and checks that
the passwords match. Password and Confirm Password fields were added
so the admin can change any or all of the user's info.

### DIFF
--- a/src/frontend/src/components/EditUser/index.tsx
+++ b/src/frontend/src/components/EditUser/index.tsx
@@ -7,6 +7,7 @@ import validateEmail from '../../service/email-validation';
 
 interface Props {
   users: UserState;
+  isAdmin?: boolean;
 }
 
 interface State {
@@ -94,17 +95,12 @@ class EditUser extends React.Component<Props, State> {
                 name="name"
                 type="text"
                 className="w-100 pa3 br2 b--black-20"
-                required={true}
                 aria-describedby={
                   this.state.missingName
                     ? 'name_input_message'
                     : undefined
                 }
-                aria-invalid={
-                  this.state.missingName
-                    ? true
-                    : false
-                }
+                aria-invalid={this.state.missingName}
                 onChange={this.handleChange}
               />
             </div>
@@ -117,17 +113,12 @@ class EditUser extends React.Component<Props, State> {
                 name="email"
                 type="email"
                 className="w-100 pa3 br2 b--black-20"
-                required={true}
                 aria-describedby={
                   this.state.invalidEmail
                     ? 'email_message'
                     : undefined
                 }
-                aria-invalid={
-                  this.state.invalidEmail
-                    ? true
-                    : false
-                }
+                aria-invalid={this.state.invalidEmail}
                 onChange={this.handleChange}
               />
             </div>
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/EditUser/index.tsx
```diff
@@ -48,37 +54,131 @@ class EditUser extends React.Component<Props, State> {
...
+                required={true}
+                aria-describedby={
+                  this.state.missingPassword
+                    ? 'password_input_message'
+                    : this.state.invalidPassword
+                    ? 'password_message'
+                    : undefined
+                }
+                aria-invalid={
+                  this.state.missingPassword || this.state.invalidPassword
```
wittejm: I'm pretty sure this ternery operator can be thrown away and you could just use `aria-invalid={ this.state.missingPassword || this.state.invalidPassword}`

src/frontend/src/components/EditUser/index.tsx
```diff
@@ -144,13 +241,45 @@ class EditUser extends React.Component<Props, State> {
...
+                </p>
+              ) : null}
+              {this.state.invalidResponse === true ? (
+                <p id="no_match_message" className="bg-washed-red mv4 pa3 br3 fw6">
+                  Technical difficulties try again later.
+                </p>
+              ) : null}
+            </div>
+
             <div className="bt b--black-20 pt5 mb5">
```
wittejm:
```suggestion
            <div className="visually-hidden bt b--black-20 pt5 mb5">
```

wittejm: Until we implement password-reset via email.

src/frontend/src/components/EditUser/index.tsx
```diff
@@ -90,21 +190,18 @@ class EditUser extends React.Component<Props, State> {
...
-                  this.state.invalidCredentials || this.state.missingPassword
+                  this.state.missingPassword
                     ? true
                     : false
                 }
                 onChange={this.handleChange}
               />
             </div>
-
             <div className="mb4">
```
wittejm: The change-role section shouldn't render if a non-admin user is updating their own account details.

You could use a similar thing as with the "Admin" button in the header, which is here:

https://github.com/codeforpdx/recordexpungPDX/blob/958feacf2d8b0e3abe1aedf99619f5aa35ab9313/src/frontend/src/components/Header/index.tsx#L34

It checks only a local variable which a user could hack to expose the "role" button, but the backend is protected to disallow a user to make themself admin, so it's fine.

src/frontend/src/components/EditUser/index.tsx
grbarker: Thanks for catching that! 
I went through and removed all the unnecessary ternery operators in those attributes.

src/frontend/src/components/EditUser/index.tsx
grbarker: Got it.

src/frontend/src/components/EditUser/index.tsx
grbarker: That would seem like a good idea. I went through and added that to the component. Thanks for the heads up.

PR #837, opened 2020-02-13
Class C felony rules string

aemrich:
Add expungement rules string for Class C felonies


wittejm:
pulled this into PR #864

PR #839, opened 2020-02-13
Prepare move off Heroku

KentShikama:
See https://github.com/codeforpdx/recordexpungPDX/blob/8d9456a734c9525298007c6f9eeb95039d3b6025/doc/recordsponge.com_devops_overview.md


KentShikama:
See https://github.com/codeforpdx/recordexpungPDX/pull/879 for deploy commands.

PR #840, opened 2020-02-13
Windows setup

loganballard:
Added a setup for docker and expung dev on Windows.  

Question: There are many things within the makefile which won't run on windows because they are specific to *nix-style commands.  example:

```
clean:
	rm -rf src/backend/*.egg-info &
	find . -type f -name \*~ | xargs rm &
	find . -type f -name \*pyc | xargs rm
```

Windows doesn't know about `xargs`, and Powershell aliases `rm` is to `Remove-Item`, which has different parameters/options to Unix's `rm`.  So many of these will have to be either rewritten to be cross-platform or conditionally checked for OS using something like: 

```
OSFLAG :=
ifeq ($(OS),Windows_NT)
	OSFLAG = WIN32
endif
```

Is this something worth focusing on?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 .env.test.local
 .env.production.local
 scrap.txt
+.editorconfig
 
 npm-debug.log*
 yarn-debug.log*
--- a/Makefile
+++ b/Makefile
@@ -38,62 +38,35 @@ dev_logs:
 	docker-compose -f docker-compose.dev.yml logs -f
 
 dev_psql:
-ifeq ($(OSFLAG), WIN32)
 	docker exec -ti $(shell docker ps -qf name=$(DB_CONTAINER_NAME)) psql -U postgres -d $(PGDATABASE)
-else
-	docker exec -ti $$(docker ps -qf name=$(DB_CONTAINER_NAME)) psql -U postgres -d $(PGDATABASE)
-endif
 
 bash_backend:
-ifeq ($(OSFLAG), WIN32)
 	docker exec -it $(shell docker ps -qf name=$(BACKEND_CONTAINER_NAME)) /bin/bash
-else
-	docker exec -it $$(docker ps -qf name=$(BACKEND_CONTAINER_NAME)) /bin/bash
-endif
 
 bash_frontend:
-ifeq ($(OSFLAG), WIN32)
 	docker exec -it $(shell docker ps -qf name=$(FRONTEND_CONTAINER_NAME)) /bin/bash
-else
-	docker exec -it $$(docker ps -qf name=$(FRONTEND_CONTAINER_NAME)) /bin/bash
-endif
+
+database_image:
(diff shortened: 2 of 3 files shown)


wittejm:
> Is this something worth focusing on?

Not really sure. I don't use `clean` very much in my own workflow and if the make targets for running the stack, checking the docker logs, shelling into the containers, and running the unit tests all work, then that's probably enough to not get slowed down at all.

loganballard:
unfortunately its more than just `clean`. For example:

```
dblogs:
	docker logs --details -ft $$(docker ps -qf name=$(DB_CONTAINER_NAME))
```

Windows doesn't understand the `$$(docker ps -qf name=$(DB_CONTAINER_NAME))`, which is used liberally throughout the makefile. A more pressing concern would be:

```
dev_test:
	docker exec -t $$(docker ps -qf name=$(BACKEND_CONTAINER_NAME)) mypy
	docker exec -t $$(docker ps -qf name=$(BACKEND_CONTAINER_NAME)) pytest
```

The windows rewrite would be something like:

```
dev_test:
ifeq ($(OSFLAG), WIN32)
	docker exec -t %BACKEND_CONTAINER_NAME% mypy
	docker exec -t %BACKEND_CONTAINER_NAME% pytest
else
	docker exec -t $$(docker ps -qf name=$(BACKEND_CONTAINER_NAME)) mypy
	docker exec -t $$(docker ps -qf name=$(BACKEND_CONTAINER_NAME)) pytest
endif
```

Which is fairly ugly (and also doesn't work at the moment) but would make it so that windows devs wouldn't have to manually run the makefile commands (which is probably worthwhile in the long run).

loganballard:
whoops accidentally closed the PR

wittejm:
gotcha! looks like the additional support would be quite useful then and also that you have a pretty solid handle on what needs to be done. 

Maybe branching each make target on the OSFLAG would clutter up the Makefile too much. Is there a tool that's better suited to handle the windows side of things?

loganballard:
Adding makefile which has been "ported" to windows.  Added conditional checks the makefile commands that assume *Nix-style command lines.  

It would be awesome if someone who does dev on mac or linux could take a look at this and make sure I didn't break the existing workflows on Mac/Linux.  Adding @aemrich and @KentShikama per Jordan's suggestion

loganballard:
removing the make recipes that aren't used anymore per Jordan's comment, leaving the PR open for another day to get feedback, but I'll merge tomorrow if no one objects

wittejm:
Oh dang. My bad, those targets are still the ones used in the travis build https://travis-ci.com/codeforpdx/recordexpungPDX so it fails without them. so we still need em. sorry... 

Though you don't have to worry about them working in Windows, since they're not part of a local dev workflow .

wittejm:
@loganballard Forgot to ask: are you using Windows pro? I think there was another windows developer who got docker working but she was the only person using Pro

PR #844, opened 2020-02-14
Show eligibility when known without disposition info, et al

### LINKED ISSUE #835: Class A Felony documentation
"
A conviction of a Class A Felony is not eligible, unless that Class A Felony was related to the Manufacturing/Delivery/Possession of Marijuana. A dismissal of a Class A Felony, like dismissals of all charges, is eligible. 
"

### This task is done when:
 -
 -

wittejm:
Edit: This PR incorporates and builds on @nickbeaird 's #850 refactor for unit tests.  This coincides with updating several of those tests for the new eligibility behavior here.

Several types are always eligible or always ineligible regardless of the ruling, so this change returns this eligibility for any valid charge types; otherwise charge type eligibility defaults to  "Needs more Analysis" if the disposition if unrecognized or missing. This default method is still defined on the Charge shared class. 

closes #835,


wittejm:
@nickbeaird heads up I'm probably breaking your work on #850 #627 some more with this because I'm changing some charge type behavior and the corresponding tests.

src/backend/tests/models/charge_types/test_traffic_offenses.py
```diff
@@ -86,6 +86,19 @@ def test_dismissed_infraction_is_not_type_eligible(self):
         assert charge.expungement_result.type_eligibility.reason == "Ineligible by omission from statute"
         assert not charge.blocks_other_charges()
 
+    def test_no_dispo_violation_is_not_type_eligible(self):
+        charge = ChargeFactory.create(
+            statute="801.000", level="Class C Traffic Violation", disposition=None, date=date(1901, 1, 1)
+        )
+        charge.disposition = None
```
KentShikama: Redundant line

wittejm:
@nickbeaird in fact I'll just pull your work now and work on top of that, so don't worry about it

wittejm:
Completed the proposed feature change and incorporated PR #850, so re-upping review requests.

There are some unused imports I haven't cleared out yet, but will do.

src/backend/expungeservice/models/charge.py
```diff
@@ -34,6 +34,13 @@ def __post_init__(self):
         self.expungement_result = ExpungementResult(type_eligibility=type_eligibility, time_eligibility=None)
 
     def _build_type_eligibility(self):
+        type_eligibility = self._type_eligibility()
```
nickbeaird: I am wondering if this may cause some issues here. I am seeing below that this will raise a NotImplementedError. That is A-okay if the goal here is to use this as a base model that should not be implemented. Although, if the goal is to have your end users overwrite this method, then you might want to look at at using an Abstract Base Class @abstractmethod.

src/backend/expungeservice/models/charge.py
```diff
@@ -34,6 +34,13 @@ def __post_init__(self):
...
+        if type_eligibility:
+            return type_eligibility
+        else:
+            return self._default_type_eligibility()
```
nickbeaird: Depending no what the team's style is here, you can drop the else.

src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py
```diff
@@ -8,8 +8,9 @@
 class Schedule1PCS(Charge):
     type_name: str = "Schedule 1 PCS"
 
-    def _default_type_eligibility(self):
+    def _type_eligibility(self):
         if self.dismissed():
-            return TypeEligibility(EligibilityStatus.ELIGIBLE, reason="Eligible under 137.225(1)(b)")
-        else:
-            return TypeEligibility(EligibilityStatus.ELIGIBLE, reason="Eligible under 137.225(5)(C)")
+            # TODO: Are violation dismissals for a PCS actually eligible? This does not appear to be the case.
```
nickbeaird: Just checking, but I assume that this is tracked somewhere?

src/backend/expungeservice/models/charge_types/subsection_12.py
```diff
@@ -14,7 +14,8 @@ class Subsection12(Charge):
 Applicable subsections: 137.225(12) for convictions; 137.225(1)(b) for dismissals."""
     )
 
-    def _default_type_eligibility(self):
+    # TODO: the elif convicted(): block needs to be added here pending the open issue for a Subsection 12 overhaul.
```
nickbeaird: Same goes for this TODO.

src/backend/expungeservice/models/charge_types/traffic_non_violation.py
```diff
@@ -8,8 +8,8 @@
 class TrafficNonViolation(Charge):
```
nickbeaird: This may be a dumb question, but what is a TrafficNonViolation Charge?

src/backend/tests/models/test_charge.py
```diff
@@ -1,12 +1,27 @@
...
 
-from datetime import date
+from datetime import date, datetime, timedelta
 from dateutil.relativedelta import relativedelta
 from expungeservice.models.disposition import Disposition
 from tests.factories.charge_factory import ChargeFactory
 from tests.factories.case_factory import CaseFactory
 
 
+class ChargeTypeTestsParent(unittest.TestCase):
```
nickbeaird: Cool, I have not thought to do this yet. This seems smart. I have not thought this through yet, but do we also need to run a teardown per class?

src/backend/tests/models/charge_types/test_unclassified.py
```diff
@@ -0,0 +1,44 @@
...
+from expungeservice.models.expungement_result import EligibilityStatus
+from expungeservice.models.disposition import Disposition
+
+from tests.factories.charge_factory import ChargeFactory
+from tests.models.test_charge import ChargeTypeTestsParent
+
+
+class TestSingleChargeUnclassified(ChargeTypeTestsParent):
+    def setUp(self):
+        ChargeTypeTestsParent.setUp(self)
```
KentShikama:
```suggestion
        super().setUp()
```

See https://stackoverflow.com/q/576169/2750819

src/backend/tests/models/charge_types/test_traffic_offenses.py
KentShikama: Looks like it is still here

src/backend/tests/models/charge_types/test_schedule_1_p_c_s.py
```diff
@@ -0,0 +1,78 @@
...
+from expungeservice.models.disposition import Disposition
+from tests.models.test_charge import ChargeTypeTestsParent
+
+
+class TestSingleChargeConvictionsSchedule1PCS(ChargeTypeTestsParent):
+
+    """def setUp(self):
+        ChargeTypeTestsParent.setUp(self)
+        # Shouldn't need this.
+    """
```
KentShikama: Left over comment?

src/backend/tests/models/charge_types/test_unclassified.py
```diff
@@ -0,0 +1,44 @@
...
+            unclassified_dismissed.expungement_result.type_eligibility.reason
+            == "Unrecognized Charge : Further Analysis Needed"
+        )
+
+    def test_charge_that_falls_through(self):
+        self.charge_dict["name"] = "Aggravated theft in the first degree"
+        self.charge_dict["statute"] = "164.057"
+        self.charge_dict["level"] = "Felony Class F"
+        charge = self.create_recent_charge()
+        self.charges.append(charge)
```
KentShikama: What does `self.charges` do here?

src/backend/tests/models/test_charge.py
```diff
@@ -1,12 +1,27 @@
...
+    def setUp(self):
+        self.charge_dict = ChargeFactory.default_dict()
+        self.charges = []
+        last_week = datetime.today() - timedelta(days=7)
+        self.convicted = Disposition(ruling="Convicted", date=last_week)
+        self.dismissed = Disposition(ruling="Dismissed", date=last_week)
+        self.diverted = Disposition(ruling="Diverted", date=last_week)
+        self.unrecognized_disposition = Disposition(ruling="Something unrecognized", date=last_week)
+        self.no_complaint = Disposition(ruling="No Complaint", date=last_week)
```
KentShikama: We should just extract these as constants like

`CONVICTED = Disposition(ruling="Convicted", date=LAST_WEEK)`

and then just call them with `SomeClass.CONVICTED`

src/backend/expungeservice/models/charge_types/traffic_non_violation.py
KentShikama: It is a traffic crime that is more severe than a violation.

src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py
wittejm: Added a new issue, thanks

PR #847, opened 2020-02-15
Remove acquittal language

loganballard:
changing internal language to dismissed from acquitted.  Left a few instances of "acquit" in there, such as in `src/backend/tests/models/charge_types/test_type_analyzer_dismissed_charges.py`, where it is specifically checking for "acquitted" language being returned from search.


loganballard:
integrating kent's changes, stand by

PR #850, opened 2020-02-15
627 refactor unit tests

nickbeaird:
Refactor the unit tests to match the model filename structure. For example the felony_class_a.py file now has a test_felony_class_a.py counterpart. 

Files were made based on the class type that was being tested. A small change was to assert that each test was returning a specific instance type. 

Tests that included multiple class types were left in their own test file.


src/backend/tests/models/charge_types/test_felony_class_b.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class TestSingleChargeConvictionsFelonyClassB(unittest.TestCase):
+    def setUp(self):
+        last_week = datetime.today() - timedelta(days=7)
+        self.single_charge = ChargeFactory.build(disposition=Disposition(ruling="Convicted", date=last_week))
+        self.charges = []
+
+    def create_recent_charge(self):
+        charge = ChargeFactory.save(self.single_charge)
```
loganballard: could this just be:

```
def create_recent_charge(self):
        return ChargeFactory.save(self.single_charge)
```

src/backend/tests/models/charge_types/test_felony_class_b.py
```diff
@@ -0,0 +1,41 @@
...
+        assert charge.expungement_result.type_eligibility.reason == "Further Analysis Needed"
+
+    def test_class_felony_is_added_to_b_felony_attribute(self):
+        self.single_charge["name"] = "Aggravated theft in the first degree"
+        self.single_charge["statute"] = "164.057"
+        self.single_charge["level"] = "Felony Class B"
+        charge = self.create_recent_charge()
+
+        assert isinstance(charge, FelonyClassB)
+        assert charge.__class__.__name__ == "FelonyClassB"
```
loganballard: This is probably my own ignorance, but is there a meaningful difference between the check on this line and the one above?

src/backend/tests/models/charge_types/test_felony_class_a.py
```diff
@@ -2,13 +2,37 @@
...
+from expungeservice.models.charge_types.felony_class_a import FelonyClassA
+from expungeservice.models.disposition import Disposition
 from expungeservice.models.expungement_result import EligibilityStatus
 
 from tests.factories.charge_factory import ChargeFactory
-from expungeservice.models.disposition import Disposition
 
 
-class TestSingleChargeAcquittals(unittest.TestCase):
+class TestSingleChargeConvictionsFelonyClassA(unittest.TestCase):
```
loganballard: Seems like for Class A/B/C, all of these testing classes share the setup/create_recent_charge methods, should they be refactored into a generic `TestSingleChargeConvictionsFelony(unittest.TestCase)` base class, which then ``TestSingleChargeConvictionsFelonyClass{A/B/C}(`TestSingleChargeConvictionsFelony)` could inherent from?

src/backend/tests/models/charge_types/test_marijuana_ineligible.py
```diff
@@ -0,0 +1,92 @@
...
+from datetime import datetime, timedelta
+
+from expungeservice.models.expungement_result import EligibilityStatus
+from expungeservice.models.charge_types.marijuana_ineligible import MarijuanaIneligible
+
+from tests.factories.charge_factory import ChargeFactory
+from expungeservice.models.disposition import Disposition
+
+
+class TestSingleChargeConvictionsMarijuanaIneligible(unittest.TestCase):
```
loganballard: This one could inherent that "base testing case" class too

src/backend/tests/models/charge_types/test_marijuana_ineligible.py
```diff
@@ -0,0 +1,92 @@
...
+        return charge
+
+    def test_marijuana_ineligible_statute_475b3493c(self):
+        self.single_charge["name"] = "Unlawful Manufacture of Marijuana Item"
+        self.single_charge["statute"] = "475B.349(3)(C)"
+        self.single_charge["level"] = "Felony Class C"
+        marijuana_felony_class_c = self.create_recent_charge()
+        self.charges.append(marijuana_felony_class_c)
+
+        assert isinstance(marijuana_felony_class_c, MarijuanaIneligible)
```
loganballard: Could all of these common asserts be a single helper method?

```
def check_marijuana_inelligibility():
        assert isinstance(marijuana_felony_class_a, MarijuanaIneligible)
        assert marijuana_felony_class_a.type_name == "Marijuana Ineligible"
        assert marijuana_felony_class_a.expungement_result.type_eligibility.status is EligibilityStatus.INELIGIBLE
        assert marijuana_felony_class_a.expungement_result.type_eligibility.reason == "Ineligible under 137.226"
```

src/backend/tests/models/charge_types/test_felony_class_b.py
KentShikama: Just the latter is harder to read and navigate references with :P

wittejm:
@nickbeaird I pulled your changes here and am working on them with my active work #844 , so I'm closing this PR.

PR #852, opened 2020-02-16
Fix filtered results when aliases returns redundant cases.

wittejm:
Deals with the fact that groupby doesn't group properly if items are not sorted by key.


wittejm:
this is fine.

Issue #747, opened 2020-01-21
Connect "EditUser" component to the backend


ForrestLonganecker:
This task is done when:

* Edit user form is populated with the selected user.
* Admin user is able to edit an exissting user by submitting an edit user form to the back end.
* Form validates that the edit user form contains updated content.
* Non-Admins are unable to edit user information.
* Error messages match required fields

hmarcks:
![account-flow](https://user-images.githubusercontent.com/6887982/72775076-a4aa0c00-3bd2-11ea-84a9-93aa94a5407c.png)

Accessible validation technique: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19

PR #858, opened 2020-02-18
Fix radio button issue.

grbarker:
Radio button was not controlled and was buggy with selection switching.
Control was added and buttons update state properly.

This addresses part of issues #746 and #747.


src/frontend/src/components/AddUser/index.tsx
```diff
@@ -93,17 +99,12 @@ class AddUser extends React.Component<Props, State> {
                 name="name"
                 type="text"
                 className="w-100 pa3 br2 b--black-20"
-                required={true}
```
wittejm: I still don't know this: what does the `required` property do and what does removing it do? It only has to do with the aria, right?

hmarcks: It's for screen readers to announce that it's required. Please see https://developer.paciellogroup.com/blog/2019/02/required-attribute-requirements

I think `required` should be static and paired with `aria-invalid ` and `aria-describedby`, but I don't have experience with validation.

grbarker: I see. I removed the required property and ran the app and did not see any difference in how the validation functioned. I'll put them all back. Thanks for pointing that out.

grbarker: I just checked the User component and it does already have a button to take you to the edit-user page just like the new user button in the Userlist component. If it is not working I'll have to take another look into it.

wittejm: Ah! I see clicking the name works, but I was only trying the pencil-icon. I think that button is what we should have for editing that user's info, and not make the person's name a link.

@hmarcks I checked your mockup and saw you didn't put the edit icon next to users, only groups. @ForrestLonganecker and I decided some time ago that we liked the idea of using the pencil icon for users also. What do you think?

src/frontend/src/components/EditUser/index.tsx
```diff
@@ -186,7 +192,9 @@ class EditUser extends React.Component<Props, State> {
                         name="role"
                         value="search"
                         className="v-top"
-                        checked
+                        checked={this.state.role === "search"}
+                        onChange={this.handleRadioChange}
+                      />
```
wittejm:
```suggestion

```

Shouldn't be here

wittejm: hanging_bracket

wittejm:
Cool, this looks good to me. I'd say merging this now is fine and the change to the edit button could be handled separately. Up to you though.

src/frontend/src/components/EditUser/index.tsx
grbarker: I added an onClick to the pencil icon to redirect to the edit-user page. Since it was just one line I decided to go ahead and include it here.

wittejm: Cool.

src/frontend/src/components/AddUser/index.tsx
hmarcks: I thought the link was a bit more appropriate since you may just be looking at their profile for more info and not necessarily editing, especially if we expose the user list to non-admins in the future, but I'm fine with either.

wittejm: @hmarcks Ah that make sense. Do you have something in mind for what a non-editor's view of a person's profile would look like? I don't think that's in the prototype. 

Basically the only information that isn't included in the table on the admin page is their email address -- maybe it just belongs there? This would require that the user list stay hidden from non-admins.

hmarcks: I don't have that view and we may never want it. I think it could be fine to show the email in the admin list and use the edit button. It could get to be a lot of info if there are a lot of users viewed on a small screen, but we can always change the layout

PR #859, opened 2020-02-18
Move charge type tests to pytest format

KentShikama:
Follow up from #844 . I would recommend reviewing commit by commit.


KentShikama:
> Is using Pytest as the testing framework and not just the runner a goal of this project?

Yeah we are gradually trying to migrate over.

I am trying to make it such that every test is locally reasonable. We may want to directly set the keyword arguments in `ChargeFactory.create` and have `ChargeFactory.create` just take the same defaults as `default_dict` right now in the future. We can always reinsert a setup method through pytest fixtures if we need it.

PR #869, opened 2020-02-19
Add person crimes accuracy

aemrich:
Addresses issue #760. 

Related to issue #868.

We may need more tests for person crimes -- right now the person crimes test suite only checks to see if felony b person crimes are accurately classified.

PR #870, opened 2020-02-19
Person crime rewrite

### LINKED ISSUE #760: Classify Person Crimes more accurately
Our current definition in the data model for a PersonCrime is only approximately correct. The actual definition is in section 14 of this page : https://secure.sos.state.or.us/oard/displayDivisionRules.action?selectedDivision=712
 
Also, the definition for Person Crime only applies for expungement in a rather specific rule:  Class B felonies that are also Person Crimes are ineligible (137.225(5)(a). 

That means that Class C Felonies and lower can be expunged even if they are person crimes. Michael said some of these statutes can be prosecuted as Class C or Class B crimes. So, to get the right classification for type eligibility we need to tag Class C felonies first, then Person Crimes, then Class B felonies. The classification order in https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/expungeservice/models/helpers/charge_classifier.py needs to be updated.
 

### This task is done when:
 - A charge is classified as a Class C felony before it's classified as a Person Crime.
 - The complete set of statutes in the above definition is used to define a Person Crime. 
 - Moar unit tests!

### LINKED ISSUE #801: Docstring for Person Crime
This is the human-readable rules breakdown for the Person Crime charge type. It will be a string field `PersonCrime.expungement_rules`.

It can follow the general style of #800 

related to #760  and #797

### LINKED ISSUE #868: Check Superseded Person Crimes
Check that the categorizations for the statutes listed below are correct.

As a background -- for issue #760, we added person crime statutes from [this source](https://secure.sos.state.or.us/oard/displayDivisionRules.action?selectedDivision=712) to our charge classifier. We found that the statutes listed below were superseded by previous classifications - we need to check that these classifications make sense.

I've commented out these statutes in the charge classifier and test person crimes files. Un-comment to test.

Subsection 6:
- 162165 
- 162185 
- 163145

Subsection 12:

- 163165 
- 163175 
- 163205 
- 163225 
- 163275 
- 163525 
- 163535 
- 164395 
- 164405 
- 166165 
- 166220   

Marijuana Ineligible:
- 475B359 
- 475B367 
- 475B371

Traffic Non Violation:
- 811705

Duii:
- 8130105

wittejm:
A few additions to @aemrich 's overhaul for Person Crime (replacing PR #869 ) 
- renamed to Person Felony, which is the language used in the expungement rule
- moved the statutes lists to the Person Felony class instead of the charge_classifier. This is part of the intended migration of most of the type-specific code out of the charge classifier.
- More comments and the expungement_rules docstring for the charge type
- Added logic to handle the possible OECI error of missing subsection info (with new unit tests)
- Unit tests to verify that person felonies that are classifiable under different types do not mistakenly get tagged as Person Felony
- Unit tests to verify a person felony that isn't a B felony doesn't get misclassified

closes #760 
closes #868 
closes #801


src/backend/expungeservice/models/charge_types/person_felony.py
```diff
@@ -0,0 +1,177 @@
...
+                return TypeEligibility(
+                    EligibilityStatus.NEEDS_MORE_ANALYSIS,
+                    reason="OECI may be missing a statute subsection which would make this charge a person crime, and thus ineligible under 137.225(5)(a)",
+                )
+            else:
+                return TypeEligibility(EligibilityStatus.INELIGIBLE, reason="Ineligible under 137.225(5)(a)")
+
+
+"""The following sections are defined in (ORS)[https://www.oregonlaws.org/] as class B felonies. Some of these have only subsections which are a class B felony.
+This list is not used directly to determine expungement eligibility because we instead use the charge level provided in the OECI record.
```
KentShikama: What is the purpose of this comment?

src/backend/tests/models/charge_types/test_person_felony.py
```diff
@@ -0,0 +1,193 @@
...
+    "1643772C",  # [contains subsection] # Computer Crime—Theft of an Intimate Image; class C misdemeanor
+    "1648863",  # [contains subsection] # Tree Spiking (Injury); class B felony
+    "1663853",  # [contains subsection] # Felony Possession of a Hoax Destructive Device; class C felony
+    "1673204",  # [contains subsection] # Felony Animal Abuse I; class C felony
+    "4757526A",  # [contains subsection] # Manufacturing or Delivering a Schedule IV Controlled Substance Thereby Causing Death to a Person; adjacent to a class B felony
+    "6099903B",  # [contains subsection] # Maintaining Dangerous Dog; class C felony
+    "8304752",  # [contains subsection] # Hit and Run Boat; no level
+    "8373652B",  # [contains subsection] # Unlawful Operation of Weaponized Unmanned Aircraft System; class C felony
+    "8373652C",  # [contains subsection] # Unlawful Operation of Weaponized Unmanned Aircraft System; class B felony
+]
```
KentShikama: This list seems to be duplicated: https://github.com/codeforpdx/recordexpungPDX/pull/870/files#diff-69da7724337f6dd0eaca7583209e1ba2R21

src/backend/tests/models/charge_types/test_person_felony.py
```diff
@@ -0,0 +1,193 @@
+from expungeservice.models.expungement_result import EligibilityStatus
+from expungeservice.models.charge_types.person_felony import PersonFelonyClassB
+from expungeservice.models.charge_types.felony_class_c import FelonyClassC
+from tests.factories.charge_factory import ChargeFactory
+from tests.models.test_charge import Dispositions
+import pytest
+
+person_felony_statutes = [
```
KentShikama: Similarly this list seems to be duplicated

src/backend/expungeservice/models/charge_types/person_felony.py
wittejm: The implementation is in some sense overkill because the complete list of person felonies includes things that can never be B felonies. But when looking at the level of a charge we consider what is in OECI and not what is the charge level named in the statute.

Arguably, you could shorten the list to only the person felonies that could also be B felonies, which would instead by this list.. I put it here as a comment is basically as extra background info. It's here for lack of a better place to put it, since we're basically trying to keep all of the legal verbiage within the source code and then extract for user-facing docs (or in this case not, because it's kind of excessive information. )

src/backend/tests/models/charge_types/test_person_felony.py
wittejm: That's intentional, so that if we update the charge type logic without updating the tests we will know it because of the failing test.

src/backend/tests/models/charge_types/test_person_felony.py
wittejm: yep, see above

src/backend/tests/models/charge_types/test_person_felony.py
KentShikama: Was this list independently written though? I don't see the point if all future changes will just involve modifying the original list and then copy pasting those changes to replace this duplicated list.

src/backend/expungeservice/models/charge_types/person_felony.py
KentShikama: What is the purpose of the long list of crimes that follow then? Is that not duplicated above in the code mostly?

src/backend/tests/models/charge_types/test_person_felony.py
wittejm: belt and suspenders?

wittejm: I see what you mean though

src/backend/expungeservice/models/charge_types/person_felony.py
wittejm: It's just the actual set of statutes typically considered class B person crimes. It seemed useful to me to keep the actual list of 30 statutes that typically apply , out of the 90 defined person felonies. The code above doesn't explicitly identify the class B felonies.

KentShikama: Ok

wittejm:
@KentShikama I removed the redundant copies of statute lists as you suggested. I feel like this change makes the unit test a little more to verify the code does what it's programmed to do (tautologically) vs verifying doing what we designed it to do. But since in practice we're just going to be copy-pasting over our statute lists if there are any changes, this doesn't add much reliability, so your suggestion makes sense. Should be ready for another look-over and merge I think .

src/backend/expungeservice/models/helpers/charge_classifier.py
```diff
@@ -172,11 +165,27 @@ def _felony_class_c(level):
...
+            else:
+                return FelonyClassB
 
     @staticmethod
     def _felony_class_a(level):
         if level == "Felony Class A":
             return FelonyClassA
+
+    @staticmethod
+    def _crime_against_person(statute):
```
KentShikama: It looks like this method returns `Optional[bool]` and thus is relying on the fact that None is falsey. We should just return `False`.

src/backend/tests/models/charge_types/test_person_felony.py
```diff
@@ -0,0 +1,87 @@
...
+    Only test the first 5 statutes just to not spam another ~75 tests.
+    """
+    charge_dict = ChargeFactory.default_dict(disposition=Dispositions.CONVICTED)
+    charge_dict["name"] = "Generic"
+    charge_dict["statute"] = person_felony_statute_not_b_felony
+    charge_dict["level"] = "Felony Class C"
+    charge_dict["disposition"] = Dispositions.CONVICTED
+    person_felony_not_class_b_convicted = ChargeFactory.create(**charge_dict)
+    assert not isinstance(person_felony_not_class_b_convicted, PersonFelonyClassB)
+    print(type(person_felony_not_class_b_convicted))
```
KentShikama: Leftover print

Issue #829, opened 2020-02-12
Criminal Mistreatment and Endangering Welfare of a Minor (Child Abuse and Elder Abuse issues)


michaelzhang43:
Convictions for Endangering Welfare of a Minor and Criminal Mistreatment, both in the First and Second Degree, if a C Felony or below, should be labeled as "Further Analysis Needed". 
Criminal Mistreatment in the First Degree has statute number 163.205
Criminal Mistreatment in the Second Degree has statute number 163.200
Endangering Welfare of a Minor has statute number 163.575

Documentation for Criminal Mistreatment in the First Degree can be 
"Criminal Mistreatment in the First Degree is not eligible if it constitutes child or elder abuse, meaning if the victim was either above 65 or under 18"

Documentation for Criminal Mistreatment in the Second Degree can be:
"Criminal Mistreatment in the First Degree is not eligible if it constitutes elder abuse, meaning if the victim was either above 65"

Documentation for Endangering Welfare of a Minor can be:
"Endangering Welfare of a Minor is not eligible if it constitutes child abuse, as described in 419B.005"

If Criminal Mistreatment is a B Felony or above, it is not eligible. 
Documentation for Criminal Mistreatment can simply follow other rules:
- If a B Felony, then it is also a person felony and therefore ineligible as a B level Person Felony
- If an A Felony, then it is ineligible as an A Felony

### This task is done when:
 -
 -

PR #871, opened 2020-02-19
Make the rules for Subsection 12 more accurate

### LINKED ISSUE #830: Overhaul Subsection 12 classification and behavior
The subsection lists the "attempted" charge for a collection of statute, which are distinct from being charged with the actual statute of the crime in question.

- The subsubsections in this subsection with "attempted" can be ignored entirely in the classification logic. These attempted crimes are prosecuted under different statutes from those named here, and the classifier handles them based on their level (misdemeanor, class c felony, etc)
- Criminal mistreatment is named here but its rule in Subsection 6 takes precedence, so its appearance here has no effect on the eligibility. 
- For the other crimes listed here, it is still useful information to the user that the statute appears in Subsection 12, because the subsection describes a more lenient criterion for eligibility, even though it doesn't strictly alter their eligibility as Misdemeanors and Class C felonies. 
- Assault 3 is listed as eligible under this subsection although it is listed as ineligible under subsection 6. For this particular contradiction, Subsection 12 takes precedence and assault 3 is eligible. 

Type eligibility reasons can be specific to the statutes because of the complexities here.
the extended rules string in `Subsection12.expungement_rules` should be exhaustive in explaining what this statute actually does, addressing the direct contradictions in the law and how they are handled.

wittejm:
* Subsection 6 overrides eligibility for Criminal mistreatment I
 * The lists of relevant statutes are migrated from charge_classifer to Subsection12 class fields, as part of the pending refactor for charge_classifier
 * The paragraphs that mention "attempt" are removed from the list of eligible statutes.

The subsection mentions some Attempt charges that are eligible. But because those charges are tried under different statutes from those listed, the statutes listed here have no effect on expungement eligibility. We could still tag these statutes as Needs More Analysis, and the code to do so is included here but commented out.

closes #830

### DIFF
--- a/src/backend/expungeservice/models/charge_types/subsection_12.py
+++ b/src/backend/expungeservice/models/charge_types/subsection_12.py
@@ -36,6 +36,8 @@ Dismissals are eligible under 137.225(1)(b)."""
     The following statutes are eligible if tried as an Attempt charge.
     But if this is the case, the actual charge will be filed under a different statute number, meaning this set of statutes never provides additional information for eligibility.
 
+    TODO: We should figure out what the statutes for attempt are so that we can identify those charges and apply this subsection.
+
     eligible_attempt_sections = [
         "163175",
         "162165",
@@ -53,13 +55,6 @@ Dismissals are eligible under 137.225(1)(b)."""
                     EligibilityStatus.NEEDS_MORE_ANALYSIS,
                     reason="Incest is possibly eligible under 137.225(12), if the victim was at least 18 years of age.",
                 )
-                """
-            elif self._section in self.eligible_attempt_sections:
-                return TypeEligibility(
-                    EligibilityStatus.NEEDS_MORE_ANALYSIS,
-                    reason="Criminal Attempt under this statute may be eligible under 137.225(12).",
-                )
-                """
             elif self._section in self.eligible_sections:
                 return TypeEligibility(
                     EligibilityStatus.ELIGIBLE,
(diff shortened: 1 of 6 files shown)


wittejm:
A test case for the time analyzer is failing because it relied on a charge that was both Subsection 12 and a class B felony.

To re-create this test, we need a different class B felony that also qualifies under some other subsection. Something else in subsection 12 or in PCS schedule 1, probably.

It's not obvious (and Idk off the top of my head) what a working example is, but we can find one.

wittejm:
Fixed the test with a different not-a-b-felony.

This also mostly completes #829

src/backend/expungeservice/models/charge_types/subsection_12.py
```diff
@@ -8,24 +8,60 @@
...
                     reason="Incest is possibly eligible under 137.225(12), if the victim was at least 18 years of age.",
                 )
-            else:
+                """
+            elif self._section in self.eligible_attempt_sections:
+                return TypeEligibility(
+                    EligibilityStatus.NEEDS_MORE_ANALYSIS,
+                    reason="Criminal Attempt under this statute may be eligible under 137.225(12).",
+                )
+                """
```
KentShikama: What is this comment for?

src/backend/tests/models/charge_types/test_subsection_12.py
```diff
@@ -115,28 +119,44 @@ def test_subsection_12_166165():
...
     charge_dict["disposition"] = Dispositions.CONVICTED
     subsection_12_charge = ChargeFactory.create(**charge_dict)
 
     assert isinstance(subsection_12_charge, Subsection12)
+    assert subsection_12_charge.expungement_result.type_eligibility.status is EligibilityStatus.NEEDS_MORE_ANALYSIS
+    assert (
+        subsection_12_charge.expungement_result.type_eligibility.reason
+        == "Criminal Attempt under this statute may be eligible under 137.225(12)."
+    )
+"""
```
KentShikama: Commented tests?

PR #872, opened 2020-02-20
Time eligibility reasons are updated for clarity

### LINKED ISSUE #619: Time eligibility reasons are inconsistent

wittejm:
Rewrote the time eligibility reasons to read more consistently and be more explicit in their meaning.

Here's an example of what two of these look like in results:

new_time_eligibility_reasons

This gets rid of the advice "recommend sequential expungement" because it's pretty hard to suss out exactly when this circumstance applies. The previous behavior was inaccurate, before and after our refactor the time analyzer.

- A dismissal is being time-blocked by a more recent dismissal.
- The more recent dismissal is eligible, making it possible to do the actual two-step expungement. 

We could do this, but we could also expect a user of the app to understand this expungement technique.

Similarly, "ten years from most recent other conviction" doesn't specify whether the other conviction is before or after the charge in question. I don't think that's a significant loss in interpretability / readability because the information is all still there in the results. 
  
closes #619 

I tweaked the language again since our convo on that issue, because I still wanted to have statute references in there.

### DIFF
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -46,11 +46,14 @@ class Expunger:
 
             if charge.convicted():
                 eligibility_dates.append(
-                    (charge.disposition.date + relativedelta(years=3), "Time-ineligible under 137.225(1)(a)")
+                    (
+                        charge.disposition.date + relativedelta(years=3),
+                        "Three years from date of conviction (137.225(1)(a))",
+                    )
                 )
 
             elif charge.dismissed():
-                eligibility_dates.append((charge.date, "Time eligible under 137.225(1)(b)"))
+                eligibility_dates.append((charge.date, "Eligible immediately (137.225(1)(b))"))
             else:
                 raise ValueError("Charge should always convicted or dismissed at this point.")
 
@@ -59,20 +62,27 @@ class Expunger:
 
             if charge.disposition.status == DispositionStatus.NO_COMPLAINT:
                 eligibility_dates.append(
-                    (charge.disposition.date + relativedelta(years=1), "Time-ineligible under 137.225(1)(b)")
+                    (
+                        charge.disposition.date + relativedelta(years=1),
+                        "One year from date of no-complaint arrest (137.225(1)(b))",
+                    )
                 )
 
             if most_recent_blocking_conviction:
+                other = "other " if charge.convicted() else ""
                 eligibility_dates.append(
                     (
                         most_recent_blocking_conviction.disposition.date + relativedelta(years=10),
-                        "Time-ineligible under 137.225(7)(b)",
+                        f"Ten years from most recent {other}conviction (137.225(7)(b))",
                     )
                 )
 
             if charge.dismissed() and most_recent_blocking_dismissal:
                 eligibility_dates.append(
-                    (most_recent_blocking_dismissal.date + relativedelta(years=3), "Recommend sequential expungement")
+                    (
+                        most_recent_blocking_dismissal.date + relativedelta(years=3),
+                        "Three years from most recent other arrest (137.225(8)(a))",
+                    )
                 )
 
             if charge.convicted() and isinstance(charge, FelonyClassB):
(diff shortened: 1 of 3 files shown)


wittejm:
forgot `make dev_test` again. Travis is a dear friend of mine.

wittejm:
While fixing test case assert statements for comparint `reason` strings, I noticed that there was a test case, `test_time_analyzer.test_ineligible_mrc_with_arrest_on_single_case()`

While adding the reason `assert` statement here I noticed the analyzer was applying the friendly rule. That's redundant with our other tests for the friendly rule, so this is now a test case in which the friendly rule doesn't get applied. Instead, the dismissal is blocked by 10 years from the date of a separate conviction. There are no other test cases for this, as far as I can tell.

Conflated two things on the same PR, a tad sloppy I know...

src/backend/expungeservice/expunger.py
```diff
@@ -59,20 +62,27 @@ def run(self) -> bool:
...
                 eligibility_dates.append(
-                    (charge.disposition.date + relativedelta(years=1), "Time-ineligible under 137.225(1)(b)")
+                    (
+                        charge.disposition.date + relativedelta(years=1),
+                        "One year from date of no-complaint arrest (137.225(1)(b))",
+                    )
                 )
 
             if most_recent_blocking_conviction:
+                other = "other " if charge.convicted() else ""
```
KentShikama:
```suggestion
                conviction_string = "other conviction" if charge.convicted() else "conviction"
```

Fairly nitpicky but I did spend 15 seconds feeling aesthetically annoyed that I'm guessing some future person might too.

KentShikama:
A future suggestion is that we could include a reference to the blocking conviction dismissal and it would only be a one/two line change.

src/backend/expungeservice/expunger.py
wittejm: ha fair. I did imagine mine my rustle someone's jimmies

wittejm:
Yea, I've thought about referencing the blocking charge. I wasn't under the impression it would only be one or two lines, but I can make an issue for it.

KentShikama:
I guess the issue is that we don't really have an ID for charges. The closest identification is something like "2nd charge of case XYZ". Maybe something like

```
eligibility_dates.append(
                    (
                        most_recent_blocking_conviction.disposition.date + relativedelta(years=10),
                        "Time-ineligible under 137.225(7)(b)",
                        f"Ten years from charge in case {most_recent_blocking_conviction.case()().case_number} (most recent {conviction}) (137.225(7)(b))",
                    )
                )
```

Issue #818, opened 2020-02-09
Add a pre-commit hook or make target to generate and sync documentation


wittejm:
For making good user-facing documentation for charge type rules,  #797 , we want to maintain the user-readable doc by syncing with the in-code rules strings on charge types. The doc file generator is in PR #814.  Running this automatically either in the dev_test make target or a pre-commit hook would be nice.

 Pinking @KentShikama for opinions

## This issue is done when:
 - Charge type documentation is tracks source code changes automatically

wittejm:
Re-generating the doc file is not hard so this issues adds only a minor convenience. Tagging as low priority

wittejm:
We weren't able to make this tool reliably cross-platform in a way that makes it useful enough, so closing the issue.

PR #873, opened 2020-02-20
adding a pre-commit hook for the autogeneration of documentation (#818)

loganballard:
addressing issue #818 
adding a pre-commit hook that uses python to auto-generate the documentation in charge-docs.  Not sure if this is the best way to do it, so welcoming feedback. Tossing this script in your `.git/hooks` folder will allow the docs to be generated when the commit is made, the disadvantage being after your commit is made, you'll want to make another commit to have the updated documentation...?

after creating this and reviewing it, I'm wondering if this would be better as a make target....

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,9 @@ dev_drop_database:
 dev_mock_oeci_up:
 	docker-compose -f docker-compose.dev.yml -f src/frontend/developerUtils/docker-compose.mock-oeci.yml up -d
 
+dev_gen_docs:
+	cd ./src/backend/ && python ./util/generate_charge_types_doc.py
+
 .PHONY: $(REQUIREMENTS_TXT)
 $(REQUIREMENTS_TXT):
 	pipenv lock -r > $@
--- a/src/backend/util/generate_charge_types_doc.py
+++ b/src/backend/util/generate_charge_types_doc.py
@@ -2,13 +2,14 @@
 This util script generates the user-facing documentation for charge types as an md file.
 """
 
-import pkgutil
+import sys
+from os import getcwd
+sys.path.append(getcwd())
 from expungeservice.models.charge import Charge
 from expungeservice.models.helpers.generator import get_charge_classes
 
 
 def generate_charge_type_doc():
-
     manual_doc_string = """# Charge Types
 The expungement system defines this custom set of charge types, to compute their correct type eligibility."
 """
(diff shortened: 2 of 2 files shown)


wittejm:
I just tried this, but putting the script into .git/hooks didn't actually do anything for me, when I then added changes to a charge type file and committed them. I haven't messed around with git hooks before.

the 'black` python formatter that @KentShikama set up for us #693 does a similar thing of making changes that require you to re-commit after your first commit. we should really be mentioning that install step in the readme install section...  And running that install adds a `pre-commit` script that will conflict with this one. 

Possible naive question, but is it safe to always count on an environment having a working ` #! /usr/bin/python`? Is this just ok because the hook is just an optional thing someone could put in their git config?

But ultimately I'm inclined to agree that having it in a make target sounds good. The hope is that it's a thing the developer never needs to think about except being aware of the doc change when it gets re-generated (which is the case because they need to add the changed doc file to the commit as well). With that in mind, putting it into the `dev_test` makes sense to me.

KentShikama:
I would prefer `dev_build`. `dev_test` needs to be as lean as possible as we run it every time with CI.

wittejm:
ah yes that makes more sense. My issue is that I rarely run `dev_build`, but that's really just a me problem.

KentShikama:
Also in theory you can run both pre-commit scripts: see https://pre-commit.com/#repository-local-hooks. I'm fine with either approach.

loganballard:
redone as a makefile recipe

wittejm:
I'm getting a syntax error on the f-string because my native `python` version is 2.7 . Changing the makefile to use `python3` also fails for being version 3.5, while f-strings are new in 3.6. Reliance on the default python version's not great, though I've been relying on `pipenv` to run the script with the Pipfile we currently have in /src/backend, which specifies version 3.7. That's allso not great since our install docs don't actually mention using pipenv outside of docker so we don't "officially support" it. 

Could fix this by switching to use str.format.

wittejm:
Having a standalone recipe sounds fine to me

loganballard:
@wittejm ah great point.  I am working under the assumption that we are using virtualenvs.  But that's not always going to be the case!  Since pipenv is the way this is working, I'll rewrite this using `pipenv shell`.  However, this is going to lead to other problems, as I've noticed that we have `uwsgi` as a dependency, which doesnt support windows... I'll figure out how to workaround this.

wittejm:
Cool, however you want to solve it sounds good. Though if it takes much more trouble to find a reliable solution, it may be more worthwhile to move to another issue since this issue is a relatively minor convenience

PR #875, opened 2020-02-21
Migrated to urlparse for more robust  URL parsing.

nk9:
I ran across this code and noticed that it wasn't using the built-in method to parse URLs. This isn't terribly robust, and you could run into trouble if for example the port is too large. It's also not a good idea to print credentials to the console, since this means that anyone who gets ahold of your log files would be able to take over your database.

I have not run your site, so I can't vouch for this code being drop-in compatible. But I encourage you to give it a try and accept the PR if it works.

I'd encourage you to rotate your credentials once this is accepted so that your log files no longer need to be treated as secret.

### DIFF
--- a/src/backend/expungeservice/database/database.py
+++ b/src/backend/expungeservice/database/database.py
@@ -6,6 +6,8 @@ import psycopg2
 import psycopg2.extras
 import os
 
+from urllib.parse import urlparse
+
 
 class Database(object):
     """Database connection class.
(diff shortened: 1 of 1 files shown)


wittejm:
Thanks for spotting this! Doesn't quite work as-is because `host = url.netloc.split(':')` assigns a `List[str]` and not a string. Easy fix, but,

We actually no longer need the code to handle the env variable `DATABASE_URL` because we're off Heroku and won't go back, so we're using just the set of separated env variables, same as dev. So I'm doing to clear this old heroku piece out.

PR #876, opened 2020-02-21
Deleted old heroku db support (also sqlalchemy, which we don't use)

wittejm:
Since we moved off of Heroku, we no longer on a db config that uses the env variable `DATABASE_URL` 

replaced #875


KentShikama:
So we're confident we don't even want keep Heroku around for testing purposes?

wittejm:
@KentShikama 
I suppose that was assumed with #839. 

Having a staging setup would be great, though I don't think Heroku is the place to do since it uses a docker setup with its own quirks (e.g. inserting a PORT var live into one of the config files), our new prod doesn't, so I wouldn't trust a heroku deploy to give us any increased confidence about changes in prod working. Do you think we should keep the DB code on hand? 

But, having a staging setup before March 11th would be great and we could potentially run it alongside the prod stack, since traffic volume isn't really a concern for the time being, no?

PR #877, opened 2020-02-21
Move arrested date element so it always renders

### LINKED ISSUE #822: Render the arrest date even if the disposition is missing.
The arrest date is always available as a field on the charge object

This change will be in the [Charge](https://github.com/codeforpdx/recordexpungPDX/blob/master/src/frontend/src/components/Charge/index.tsx) file. 

An example of the undesirable result is here:

arrest date missing

wittejm:
closes #822

### DIFF
--- a/src/frontend/src/components/Charge/index.tsx
+++ b/src/frontend/src/components/Charge/index.tsx
@@ -30,14 +30,10 @@ export default class Charge extends React.Component<Props> {
       if (disposition.amended) {
         dispositionEvent += " (Amended)"
       }
-      return <>
+      return (
         <li className="mb2">
           <span className="fw7">Disposition:</span> {dispositionEvent}
-        </li>
-        <li className="mb2">
-          <span className="fw7">Charged: </span> {date}
-        </li>
-      </>
+        </li> )
     };
 
     const dispositionRender = (disposition: any, date: any) => {
@@ -84,6 +80,9 @@ export default class Charge extends React.Component<Props> {
                 {`${statute}-${name}`}
               </li>
               {dispositionRender(disposition, date)}
+              <li className="mb2">
+                <span className="fw7">Charged: </span> {date}
+              </li>
             </ul>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/Charge/index.tsx
```diff
@@ -30,14 +30,10 @@ export default class Charge extends React.Component<Props> {
...
           <span className="fw7">Disposition:</span> {dispositionEvent}
-        </li>
-        <li className="mb2">
-          <span className="fw7">Charged: </span> {date}
-        </li>
-      </>
+        </li> )
     };
 
     const dispositionRender = (disposition: any, date: any) => {
```
loganballard: not related to this development, but any reason why known/null disposition is split into separate functions?  seems like they're logically the same thing, just with the null case split into its own function.  I would suggest splitting it into a function which makes the disposition string from the disposition object, then have the dispositionRender function create the actual html?

src/frontend/src/components/Charge/index.tsx
```diff
@@ -20,36 +20,31 @@ export default class Charge extends React.Component<Props> {
...
-      }
-      if (disposition.amended) {
-        dispositionEvent += " (Amended)"
+    const dispositionEvent = (disposition: any, date: any) => {
+      let dispositionEvent;
+      if (disposition === null) {
+        dispositionEvent = "Missing";
+      }  else {
+        dispositionEvent = disposition.status;
+        if (disposition.status === "Convicted") {
```
KentShikama: What about when status is acquitted?

wittejm: The "Dismissed" ruling is being assigned at the line just above this, `dispositionEvent = disposition.status;` and that should be the entire `dispositionEvent`. We exclude the dismissal date if the charge is dismissed because it has no purpose for eligibility of or applying for expungement.

src/frontend/src/components/Charge/index.tsx
wittejm: yep good point, done.

PR #878, opened 2020-02-21
Auto generate docs

loganballard:


loganballard:
python2 not compatible

PR #879, opened 2020-02-22
Add shell commands to update backend/frontend on recordsponge.com

KentShikama:
After some thought, I thought the Python script was unnecessary when we could just add a few lines to the Makefile like I did.

```
1. ssh root@recordsponge.com
2. sudo -iu kent
3. cd ~/recordexpungPDX/
4. make deploy
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,27 @@ dev_drop_database:
 dev_mock_oeci_up:
 	docker-compose -f docker-compose.dev.yml -f src/frontend/developerUtils/docker-compose.mock-oeci.yml up -d
 
+deploy: deploy_backend deploy_frontend
+
+.ONESHELL:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -71,6 +70,27 @@ dev_drop_database:
 dev_mock_oeci_up:
 	docker-compose -f docker-compose.dev.yml -f src/frontend/developerUtils/docker-compose.mock-oeci.yml up -d
 
+deploy: deploy_backend deploy_frontend
+
+.ONESHELL:
```
wittejm: what is ONESHELL?

KentShikama: Each command in a make task will otherwise be run in a separate shell so `cd` becomes effectively useless.

Issue #860, opened 2020-02-18
Dismissal of a Violation-level PCS is not eligible


wittejm:
Like other violations, these are not covered by the statute and the PCS eligibility refers only to convictions. 

Update the logic at: 
https://github.com/codeforpdx/recordexpungPDX/blob/8f29567c10653d4bd391dc927fcd04f34346cc01/src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py#L13

Reason string will match that of Violation dismissal.

### This task is done when:
 - If the PCS is dismissed, then check the charge's level string: if it (in lowercase) contains "violation", it's a violation. If it's not a violation, keep the current returned result.
 - see non_traffic_violation for the appropriate eligibility response for a dismissed violation.
 - create a new unit test for a dismissed PCS violation

wittejm:
completed with #885

PR #885, opened 2020-02-26
adding a check for dismissed PCS violations as well as a test

loganballard:
Fixes issue #860

### DIFF
--- a/src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py
+++ b/src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py
@@ -10,7 +10,11 @@ class Schedule1PCS(Charge):
 
     def _type_eligibility(self):
         if self.dismissed():
-            # TODO: Are violation dismissals for a PCS actually eligible? This does not appear to be the case.
+            if "violation" in self.level.lower():
+                return TypeEligibility(
+                    EligibilityStatus.INELIGIBLE,
+                    reason="Dismissed violations are ineligible by omission from statute"
+                )
             return TypeEligibility(EligibilityStatus.ELIGIBLE, reason="Dismissals are eligible under 137.225(1)(b)")
         elif self.convicted():
             return TypeEligibility(EligibilityStatus.ELIGIBLE, reason="Eligible under 137.225(5)(c)")
--- a/src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py
+++ b/src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py
@@ -15,6 +15,7 @@ class Schedule1PCS(Charge):
                     EligibilityStatus.INELIGIBLE,
                     reason="Dismissed violations are ineligible by omission from statute"
                 )
-            return TypeEligibility(EligibilityStatus.ELIGIBLE, reason="Dismissals are eligible under 137.225(1)(b)")
+            else:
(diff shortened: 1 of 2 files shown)


src/backend/tests/models/charge_types/test_schedule_1_p_c_s.py
```diff
@@ -68,3 +68,21 @@ def test_pcs_475992():
...
+    charge_dict["name"] = "Poss Controlled Sub 2"
+    charge_dict["statute"] = "4759924B"
+    charge_dict["disposition"] = Dispositions.DISMISSED
+    for level in ("Felony Class viOlation C", "Felony Violation", "violation Felony", "some other Violation"):
+        charge_dict["level"] = level
+        pcs_charge = ChargeFactory.create(**charge_dict)
+        assert isinstance(pcs_charge, Schedule1PCS)
+        assert pcs_charge.expungement_result.type_eligibility.status is EligibilityStatus.INELIGIBLE
+        assert pcs_charge.expungement_result.type_eligibility.reason == "Dismissed violations are ineligible by omission from statute"
+    charge_dict["level"] = "Felony Class C"  # also test non-violation
```
wittejm: This can be in a separate unit test  `test_pcs_dismissed_nonviolation`

src/backend/expungeservice/models/charge_types/schedule_1_p_c_s.py
```diff
@@ -10,7 +10,11 @@ class Schedule1PCS(Charge):
 
     def _type_eligibility(self):
         if self.dismissed():
-            # TODO: Are violation dismissals for a PCS actually eligible? This does not appear to be the case.
+            if "violation" in self.level.lower():
+                return TypeEligibility(
+                    EligibilityStatus.INELIGIBLE,
+                    reason="Dismissed violations are ineligible by omission from statute"
+                )
             return TypeEligibility(EligibilityStatus.ELIGIBLE, reason="Dismissals are eligible under 137.225(1)(b)")
```
wittejm: I suggest wrapping this line in an `else` block since that's the style we use generally.

loganballard:
@wittejm comments addressed

src/backend/tests/models/charge_types/test_schedule_1_p_c_s.py
```diff
@@ -68,3 +68,28 @@ def test_pcs_475992():
...
     assert pcs_charge.expungement_result.type_eligibility.status is EligibilityStatus.ELIGIBLE
     assert pcs_charge.expungement_result.type_eligibility.reason == "Eligible under 137.225(5)(c)"
+
+
+def test_pcs_dismissed_violation():
+    charge_dict = ChargeFactory.default_dict()
+    charge_dict["name"] = "Poss Controlled Sub 2"
+    charge_dict["statute"] = "4759924B"
+    charge_dict["disposition"] = Dispositions.DISMISSED
+    for level in ("Felony Class viOlation C", "Felony Violation", "violation Felony", "some other Violation"):
```
wittejm: Ah shoot. Just saw this too. There's no such thing as a Felony Violation, so these test cases seem weird. If we did handle them, we'd probably report an OECI data error, but I don't think we need to bother. So I would probably just take them out.

wittejm: But "Class C Violation" seems like a plausible string to check for

loganballard: updated with sensible test cases

wittejm:
Cool. feel free to merge!

PR #889, opened 2020-02-27
Conn adduser

grbarker:
This PR address issue #746 and connects the AddUser component to the backend. 

A check to make sure that only admins can add users is still needed.


wittejm:
Found a behavior bug: If the form validation yields errors, this doesn't stop the actual request to the backend. I was able to successfully create a new user if the email address format is invalid, or if the second password field is a mismatch.

The endpoint code itself checks that all required fields are present and that the password is at least 8 characters long. That's why the form validation might look like it's blocking the request to the endpoint.

You'll need to only call the endpoint after the validation passes with a success.

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -56,33 +66,61 @@ class AddUser extends React.Component<Props, State> {
...
+      this.state.password,
+      this.state.group,
+      admin
+    ).catch(error => {
+      if (error.response.status === 403) {
+        // error if user is not admin
+        this.setState({ errorType: 'unauthorized' });
+      } else if (error.response.status === 422) {
+        this.setState({
+          errorType: 'duplicate email',
```
wittejm:
```suggestion
          errorType: 'endpoint',
```

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -255,6 +293,11 @@ class AddUser extends React.Component<Props, State> {
                   Technical difficulties try again later.
                 </p>
               ) : null}
+              {this.state.errorType === 'duplicate email' ? (
+                <p id="duplicate_email_message" className="bg-washed-red mv4 pa3 br3 fw6">
```
wittejm:
```suggestion
                <p id="endpoint_error_message" className="bg-washed-red mv4 pa3 br3 fw6">
```

src/frontend/src/components/AddUser/index.tsx
```diff
@@ -280,6 +293,11 @@ class AddUser extends React.Component<Props, State> {
                   Technical difficulties try again later.
                 </p>
               ) : null}
+              {this.state.errorType === 'duplicate email' ? (
```
wittejm:
```suggestion
              {this.state.errorType === 'endpoint' ? (
```

Issue #809, opened 2020-02-07
ManufactureDelivery is a new charge type


wittejm:
Manufacture and delivery of a controlled substance is either a Class A or Class B felony, which determines whether it is eligible. But we want a unique charge type because Class B felony is  "Needs more Analysis", and we can be definitive in the case of Manu/Del.

tied to #808 

### This task is done when:
 - Classify a ManufactureDelivery charge with the statute 475.9921b -- verify this or possible other identification method with Michael.
 - Dismissals are always type eligible under 137.225(1)(b).
 - Type Eligibility of a conviction is determined by the Level: B Felony is eligible and A Felony is ineligible. "Felony Unclassified" is a potential OECI error and "Needs more analysis". 
 - The class B felony case still follows the B felony time eligibility rule.
 - If the charge is "felony unclassified" and needs more analysis, it's still subject to the B-felony rule. 
 - Unit tests for each of these different pieces of the enhancement.

michaelzhang43:
DRUG CRIMES
Manufacturing/Delivery/Possession
of a Controlled Substance in Schedules I or II

I. Manufacturing, Delivery, Manufacturing/Delivery and Manu/Del should be treated the same
For these, there are these possibilities:
1. The underlying substance is for Heroin, Cocaine, or Crack, in which case it usually is classified as a B Felony and follows B Felony analysis (20 years, no arrests). It is sometimes classified as a "Felony Unclassified" - in these cases, we should 
2. The underlying substance is for Marijuana, in which case it is usually classified as an A Felony BUT is type-eligible and follows time eligibility of a type eligible misdemeanor. It is sometimes classified as a "Felony Unclassified"
3. The statute numbers for this is 475.9921b

**Takeaway**
1. If the class level is B Felony, then it follows B Felony rules
2. If the class level is unclassified or A Felony, then it needs more analysis (asking the user about whether the substance was marijuana, and I'll include this in the manual)

PR #892, opened 2020-02-28
Manufacture Delivery might be eligible.

### LINKED ISSUE #804: Delivery of marijuana to a minor is eligible
The bug report has the following as a type-eligible charge:

4758604A-Delivery of Marijuana to Minor

The eligibility is given under a different statute from the usual expungement statute:

https://www.oregonlaws.org/ors/137.226

It's probably best to create a new charge type for this, and it's possible that the charge type could include other marijuana charges, depending on the details in 137.226

wittejm:
Manu/del is identifiable with "Manu/del" or "delivery" keywords in the charge name.

This is only to pick of the possibility of having a marijuana charge that is not directly identifiable by name or statute.
 - Inspect manually to see if the charge meets eligibility as a marijuana charge
 - Allow fall-through to charge level charge types if it's schedule 2. 

Addresses #809 
closes #804 

Still needs fixed unit tests.


wittejm:
Unit tests failing because leap day is coming up. That is seriously excellent.

wittejm:
Adding also the MarijuanaEligible type, which could have a better name since it's similar to MarijuanaIneligible.

This adds some more exacting checks in the `charge_classifier` for drug/marijuana charges.

@michaelzhang43 said we can remove 475992 from PCS schedule 1 statutes , so that is breaking a few of our tests that I have not updated. We need to verify that we can handle eligibility correctly without relying on the statute number (it turns out to be very unindicative for eligibility).

Still a WIP because we should add more tests for the various possible records, particularly some real records that we can pull from the Bug Reports page or obtain from Michael.

KentShikama:
It looks like some tests in https://github.com/codeforpdx/recordexpungPDX/blob/master/src/backend/tests/models/charge_types/test_schedule_1_p_c_s.py need to be adjusted.

wittejm:
@KentShikama I think we need to have this change in for drug charges and it gets us 90% of the way, and the last 10% can be handled by quantum analysis. But please lmk if you have suggestions on this portion.

src/backend/tests/models/charge_types/test_manufacture_delivery.py
```diff
@@ -0,0 +1,103 @@
...
+    manufacture_delivery_charge = ChargeFactory.create(**charge_dict)
+
+    assert isinstance(manufacture_delivery_charge, ManufactureDelivery)
+    assert manufacture_delivery_charge.expungement_result.type_eligibility.status is EligibilityStatus.NEEDS_MORE_ANALYSIS
+    assert (
+        manufacture_delivery_charge.expungement_result.type_eligibility.reason == "This may be eligible if it is a charge for marijuana, under 137.226. See additional legal details."
+    )
+
+def test_manufacture_delivery_manudel_felony_unclassified():
+    charge_dict = ChargeFactory.default_dict()
```
KentShikama: This function no longer exists

src/backend/expungeservice/models/charge_types/marijuana_eligible.py
```diff
@@ -0,0 +1,23 @@
+from dataclasses import dataclass
+
+from expungeservice.models.charge import Charge
+from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+from expungeservice.models.disposition import DispositionStatus
+
+
+@dataclass(eq=False)
```
KentShikama:
```suggestion
@dataclass
```

src/backend/expungeservice/models/charge_types/manufacture_delivery.py
```diff
@@ -0,0 +1,28 @@
+from dataclasses import dataclass
+
+from expungeservice.models.charge import Charge
+from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+
+
+@dataclass(eq=False)
```
KentShikama:
```suggestion
@dataclass
```

PR #893, opened 2020-02-28
Update message added to OECI component

wittejm:
Wasn't totally sure where this belongs, but because of the eCourt portion, the oeci login makes the most sense.

For our twice-weekly updates I'm tempted to elaborate that the website will actually probably not go offline unless something is wrong. 

I didn't style the text (and used `<br>`, shame) but I'd appreciate your input on the appearance and placement @hmarcks 

updatesched

### DIFF
--- a/src/frontend/src/components/OeciLogin/index.tsx
+++ b/src/frontend/src/components/OeciLogin/index.tsx
@@ -67,7 +67,7 @@ class OeciLogin extends React.Component<Props, State> {
   public render() {
     return (
       <main className="mw8 center ph2">
-        <section className="mw6 center cf mt4 mb3 pa4 pa5-ns pt4-ns bg-white shadow br3">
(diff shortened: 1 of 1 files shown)


hmarcks:
hey @wittejm I have actually been meaning to create an issue to update the OECI login style because it looks so similar to the main login. Half the time I'm, "I already logged in! Oh right." If you agree with this layout then I can just make these updates:

OECI login v2

wittejm:
@hmarcks Yes, awesome. I've noticed this issue too when showing the app to people and that is a great solution. Looks great!

hmarcks:
👍 I'll update this branch tomorrow

hmarcks:
@wittejm updated. It looks like the failing test is unrelated.

wittejm:
@hmarcks Great, thanks. Yep, the failing test case is happening because tomorrow is leap day.

Issue #887, opened 2020-02-27
Fix subsection 12


wittejm:
We're updating our interpretation of subsection 12 (It's worth reading the subsection directly here: https://www.oregonlaws.org/ors/137.225)

 - Shorten the list of statutes to the actual subset that are already eligible under some other criterion from the statute (this is already mostly the case, but double-check for correctness.). Add notes on the list to identify why each statute is eligible (e.g. some are Class C Felonies)
 - Update the `expungement_rules` to remove the verbiage "in practice, this is never consiered..." etc.  
 - Update the unit tests if necessary

PR #897, opened 2020-02-28
Re-add two statutes to person crimes list

### LINKED ISSUE #886: Assault II is a Person crime, and ineligible if it is a Class B felony
The statute (163175) was commented out in our list of Person Felonies, I think because it is named in Subsection 12.

But the eligible charge named in Subsection 12 is actually  Attempted Assault II, which is charged under a different statute. 

So 163175 is a Person Crime and should be uncommented. A single-line fix (no change required in the unit test because it relies on the same list of statutes), but verify this with @KentShikama and @michaelzhang43  

I found this in a still-open Bug Report from Michael.

I want verification because I think we talked at length about Assault II already, so I'm surprised it still seems to be wrong here, unless I'm misremembering something.

### LINKED ISSUE #888: Robbery 2 is a person crime
We initially commented out robbery 2 because it fell under subsection 12. This is incorrect ( 12 covers attempted robbery ), so it needs to be in the list of person crimes.

### This task is done when:
 -
 -

wittejm:
assault II and robbery II re-added to Person Crimes list, due to re-interpretation of Subsection 12 -- see issue #887 

closes #888  
closes #886

### DIFF
--- a/src/backend/expungeservice/models/charge_types/person_felony.py
+++ b/src/backend/expungeservice/models/charge_types/person_felony.py
@@ -50,7 +50,7 @@ A person felony that is not a class B felony is not considered under this subsec
         "163149",  # Aggravated Vehicular Homicide;
         # "1631603", # [contains subsection] # Felony Assault;
         # "163165",  # [Subsection 6] Assault III;
-        # "163175",  # [Subsection 12] Assault II;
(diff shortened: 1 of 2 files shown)


wittejm:
clarified in #898

PR #898, opened 2020-02-28
Removed Subsection 12 from charge types

### LINKED ISSUE #887: Fix subsection 12

wittejm:
Subsection 12 only provides a list of statutes that are "more lenient to expunge" but only if they're already eligible under other criteria. It names "Attempt" charges which are difficult to identify in the expunger. Charges for the actual listed statutes (for Robbery 2, etc) are not affected by subsection 12.

closes #887 

Overall, it's a pretty un-useful statute so we're removing it from analysis.

 * All statutes remaining under Subsection 12 are re-classified to Felony Class C and have the same eligibility.

 * test_friendly_rule got updated because it now relies on a different charge type with NEEDS_MORE_ANALYSIS.


src/backend/tests/test_friendly_rule.py
```diff
@@ -94,11 +94,12 @@ def test_eligible_mrc_with_violation():
...
 def test_needs_more_analysis_mrc_with_single_arrest():
     charge_dict = ChargeFactory.default_dict()
-    charge_dict["name"] = "Incest"
-    charge_dict["statute"] = "163.525"
+    charge_dict["name"] = "Assault in the third degree"
+    charge_dict["statute"] = "163.165"
     charge_dict["level"] = "Felony Class C"
     charge_dict["disposition"] = Disposition(ruling="Convicted", date=Time.THREE_YEARS_AGO)
     three_yr_mrc = ChargeFactory.create(**charge_dict)
+    print(three_yr_mrc)
```
KentShikama: Debugging print

PR #901, opened 2020-02-29
Group charges by eligible dates in summary

### LINKED ISSUE #827: "Eligible [date]" charge sections in the summary panel
There are a very small number of different eligibility dates that can appear on a single record, because the most recent conviction has its own date but all other convictions will have a shared date. Similar deal possible with dismissals. And Class B felonies can have their own date. That means there are usually 2-3 (and *maybe* 4) dates, which can pretty comfortably fit into the summary panel  

These should be blue-labeled like the the will-be-eligible labels in the charge details.
 
This task is done when:
 - The backend Summarizer code is updated to collect these new lists of charges.
 - The frontend is updated to show the new summary info

wittejm:
- Constructs the new data structure in the record summarizer in the backend
 - The eligible charges are given as a list of 2-tuples, in order to imitate a sorted dictionary (which doesn't exist in JSON)
 * The charge name is split on `"("` to exclude any subsequent text which looks like clutter, and can still be found in the charge details. 

The summary panel now looks like this, for example:

new_summary_panel
 
I removed the parenthetical note on charge names because the example otherwise looks like this:

I'm not sure how often this sort of parenthetical happens, or if it's always ok to toss out like this.  @michaelzhang43 what do you think? 

by_dates_summary_with_parens

closes #827


src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -53,15 +54,27 @@ def summarize(record) -> RecordSummary:
...
             c.name
             for c in record.charges
             if c.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.ELIGIBLE_NOW
         ]
+        eligible_charges_by_date = [["now",eligible_charges_now]]
+        will_be_eligible_charges : Dict[date, List[str]] = {}
+        for charge in record.charges:
+            if charge.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.WILL_BE_ELIGIBLE:
+                date_eligible = charge.expungement_result.time_eligibility.date_will_be_eligible
+                charge_name = charge.name.split("(")[0]
```
KentShikama: Maybe call it `charge_short_name` to make it clear you're just trying to shorten the name for aesthetic reasons.

src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -53,15 +54,27 @@ def summarize(record) -> RecordSummary:
...
+                if date_eligible not in will_be_eligible_charges.keys():
+                    will_be_eligible_charges[date_eligible] = [charge_name]
+                else:
+                    will_be_eligible_charges[date_eligible].append(charge_name)
```
KentShikama:
```python
if will_be_eligible_charges.get(date_eligible):
    will_be_eligible_charges[date_eligible].append(charge_name)
else:
    will_be_eligible_charges[date_eligible] = [charge_name]
```

src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -53,15 +54,27 @@ def summarize(record) -> RecordSummary:
...
+        for dateval in sorted(will_be_eligible_charges):
```
KentShikama: date_value

src/frontend/src/components/RecordSummary/ChargesList.tsx
```diff
@@ -1,27 +1,48 @@
...
-         {(listItems.length > 0 ? listItems : "None")}
-        </ul>
+        {summarizedCharges};
       </div>
-      )
+    )
+  }
+
+  buildListItems(chargesNames: string[]) {
+    let listItems = chargesNames.map(((chargeName: string, index:number) => {
```
KentShikama: Prefer `const` over `let`

src/backend/expungeservice/models/record_summary.py
```diff
@@ -17,7 +17,7 @@ class RecordSummary:
     record: Record
     total_charges: int
     cases_sorted: Dict[str, List[str]]
-    eligible_charges: List[str]
+    eligible_charges_by_date: List[List[Any]] # Each list element is a list of length 2: [str, List[str]]
```
KentShikama: Do you know about `Tuple[str, List[str]]`?

KentShikama: Do you know about `Tuple[str, List[str]]`?

src/frontend/src/components/RecordSummary/ChargesList.tsx
KentShikama: Prefer `const` over `let`

src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -53,15 +54,27 @@ def summarize(record) -> RecordSummary:
         for county, balance in county_balances.items():
             county_balances_list.append(CountyBalance(county, round(balance, 2)))
         total_charges = len(record.charges)
-        eligible_charges = [
+        eligible_charges_now = [
             c.name
             for c in record.charges
             if c.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.ELIGIBLE_NOW
         ]
+        eligible_charges_by_date = [["now",eligible_charges_now]]
```
KentShikama:
```suggestion
        eligible_charges_by_date = [("now",eligible_charges_now)]
```

`List`s of fixed lengths should be `Tuple`s

src/backend/expungeservice/models/helpers/record_summarizer.py
```diff
@@ -53,15 +54,27 @@ def summarize(record) -> RecordSummary:
...
+        for charge in record.charges:
+            if charge.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.WILL_BE_ELIGIBLE:
+                date_eligible = charge.expungement_result.time_eligibility.date_will_be_eligible
+                charge_name = charge.name.split("(")[0]
+                if date_eligible not in will_be_eligible_charges.keys():
+                    will_be_eligible_charges[date_eligible] = [charge_name]
+                else:
+                    will_be_eligible_charges[date_eligible].append(charge_name)
+        for dateval in sorted(will_be_eligible_charges):
+            eligible_charges_by_date.append([dateval.strftime('%b %-d, %Y'), will_be_eligible_charges[dateval]])
```
KentShikama:
```suggestion
            eligible_charges_by_date.append((date_value.strftime('%b %-d, %Y'), will_be_eligible_charges[date_value]))
```

src/backend/expungeservice/models/record_summary.py
wittejm: Ah cool. I did not

KentShikama:
Looks like `mypy` doesn't actually fail the build: https://travis-ci.com/codeforpdx/recordexpungPDX/jobs/294181144#L331

wittejm:
Dang, gimme a proper compile error.  fixed

src/frontend/src/components/RecordSummary/ChargesList.tsx
```diff
@@ -1,27 +1,48 @@
...
 export default class ChargesList extends React.Component<Props> {
   render() {
-
-    const listItems = this.props.eligibleCharges.map(((chargeName:string, index:number) => {
-      return <li className="bt b--light-gray pt1 mb1">{chargeName}</li>
+    const summarizedCharges = this.props.eligibleChargesByDate.map(((chargeGroup:any, index:number) => {
+      let eligibilityDate = chargeGroup[0];
+      let chargesNames = chargeGroup[1];
+      let listItems = this.buildListItems(chargesNames);
+      let labelColor = (eligibilityDate==="now" ? "green" : "dark-blue");
```
KentShikama: Prefer `const` over `let`

KentShikama:
@wittejm This will require a bit more thought as to how handle charges with ambiguous time eligibility dates. Currently everything that is possibly eligible is set to be Eligible Dec 31, 9999, which is a default that can certainly be improved upon.

KentShikama:
I'm thinking that if there is any ambiguity, we should display "Charge summary results cannot be displayed until all questions are answered". What do you think @wittejm ?

KentShikama:
Screenshot from latest commit:

Screen Shot 2020-04-05 at 9 58 51 AM

wittejm:
Let's check with @michaelzhang43. I think I like showing the summary without requiring all the disambiguations first, and given that, I think showing the best date as the "possibly eligible" date is fine for the same reason that that's what we're showing in the charge details.

KentShikama:
What if there are two or three possibly eligible dates?

KentShikama:
In either case, I think it is fine to go ahead and merge this PR. Showing a message saying "Charge summary results cannot be displayed until all questions are answered" is a frontend adjustment that just hides information.

michaelzhang43:
Is the question whether to show charge results for Needs-more-analysis-charges in the summary panel before the user provides the additional information?

I would be in favor of either 1) not including the results in the summary panel until they are disambiguated or 2) keeping them separate in a 'Needs more analysis' section

wittejm:
> or 2) keeping them separate in a 'Needs more analysis' section

Yea, that's what I was assuming it would need to look like. To address Kent's question under that assumption:

> What if there are two or three possibly eligible dates?

If it says needs more analysis, it would be not incorrect to say, "Possibly eligible [sooner]" even if the alternate possibility is later.

PR #910, opened 2020-03-01
Add logic for 137.225(1)(c) to expunger (Probation Revoked)

KentShikama:
Instead of doing a full text fuzzy search, we now do a fuzzy search per event for "probation revoked" and extract the date. We assume probations apply to every conviction in the case as in practice every case in which there is a probation all convictions are usually merged into the “main” conviction.


KentShikama:
I forgot to switch the base branch back to master before merging https://github.com/codeforpdx/recordexpungPDX/pull/775

KentShikama:
Merging based on approval from https://github.com/codeforpdx/recordexpungPDX/pull/775

Issue #861, opened 2020-02-18
Show eligibility when possible for charge types without known ruling


wittejm:
The test cases for charge types generally cover dismissals and convictions, but now that we have more specific eligibility for some types with an unknown ruling, we can augment the unit tests to match for all charge types.

Some charge types are also still missing this additional eligibility status, when it could be known despite having no or an unrecognized disposition.

How to handle missing and unrecognized dispositions: If the charge type is either always eligible or always ineligible regardless of the disposition on that charge, we can provide that eligibility information. See Parking Tickets as an example of this.

This is series of standalone fixes ( See PR #890 for an example of what each fix will look like), one for each charge type, so if you're working on this issue please open a new PR for any of your incremental work rather than leave an open in-progress branch, so we can jointly make progress and there is no redundant work done on the issue.

aemrich:
Felony Class C charges should always be eligible, even if the disposition is unrecognized. Currently, Felony Class C charges only handle convicted and dismissed dispositions

aemrich:
Part of this issue is double checking the charge types to see if they correctly handle unrecognized dispositions

aemrich:
test_parking_ticket.py is the model for unit tests for unrecognized dispositions

wittejm:
Expanded the issue description above including the reference to the template PR #890 to follow

aemrich:
Charge types to review:

- [ ] civil offense

- [ ] duii

- [ ] felony class a

- [ ] felony class b

- [ ] felony class c

- [ ] juvenile charges

- [ ] marijuana ineligible

- [ ] misdemeanor

- [ ] parking ticket

- [ ] person felony

- [ ] schedule 1 p c s

- [ ] subsection 6

- [ ] traffic non violation

- [ ] traffic violation

- [ ] unclassified charge

- [ ] violation

wittejm:
Completed by a variety of PRs from several contributors!

PR #913, opened 2020-03-01
Adds juvenile charge tests for more dispositions

aemrich:
-adds tests for convicted, unrecognized, and no disposition


aemrich:
Related to #861

PR #909, opened 2020-03-01
Added hyperlinks to case numbers in error banners

KorynLA:
- Changes backend wording so case number links are added to the end of the error message.
- Added hyperlinked case numbers to end of messages.
- Frontend has been updated to iterate through each case number found in the error text using regex and appending it to the end of the text as hyperlinks.

PR #916, opened 2020-03-02
Issue 695

KorynLA:
New updates with the proposed changes from PR #909 :
- Used '[]' as wrappers for backend error case numbers.
- Updated frontend to use match and append hyperlinked case numbers
- Updated backend test_crawler_expunger to no longer fail with newly wrapped error case numbers.

### DIFF
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -213,7 +213,7 @@ class Expunger:
         cases_with_unrecognized_disposition: Set[Tuple[str, str]] = set()
         for charge in charges:
             if charge.blocks_other_charges():
-                case_number = f"[{charge.case()().case_number}]";
+                case_number = f"[{charge.case()().case_number}]"
                 if not charge.disposition and charge.case()().closed():
                     cases_with_missing_disposition.add(case_number)
                 elif charge.disposition and charge.disposition.status == DispositionStatus.UNRECOGNIZED:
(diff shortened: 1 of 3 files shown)


src/backend/expungeservice/expunger.py
```diff
@@ -213,7 +213,7 @@ def _filter_cases_with_errors(charges: List[Charge]):
         cases_with_unrecognized_disposition: Set[Tuple[str, str]] = set()
         for charge in charges:
             if charge.blocks_other_charges():
-                case_number = charge.case()().case_number
+                case_number = f"[{charge.case()().case_number}]";
```
KentShikama: semicolon

src/frontend/src/components/SearchResults/index.tsx
```diff
@@ -12,8 +12,19 @@ export default class SearchResults extends React.Component<Props> {
     const errors = ( this.props.record.errors ?
       this.props.record.errors.map(((errorMessage: string, errorIndex: number) => {
         const id= "record_error_" + errorIndex;
+        
+        const errorMessageArray = errorMessage.split(/(\[.*?\])/g); 
+        console.log("Message" + errorMessageArray);
```
KentShikama: log message

src/backend/expungeservice/expunger.py
```diff
@@ -24,7 +24,7 @@ def run(self) -> bool:
         """
         open_cases = [case for case in self.record.cases if not case.closed()]
         if len(open_cases) > 0:
-            case_numbers = ",".join([case.case_number for case in open_cases])
+            case_numbers = ",".join([f"[{case.case_number}]" for case in open_cases])
```
wittejm: While we're here you can fix this space:
```suggestion
            case_numbers = ", ".join([f"[{case.case_number}]" for case in open_cases])
```

wittejm:
merged this via an updated branch #920

PR #917, opened 2020-03-04
New charge type: Contempt of Court

### LINKED ISSUE #891: Contempt of Court
Is a civil offense

Is most reliable to identify by name.lower() contains "contempt of court".

### This task is done when:
 - new charge type file
 - added to charge_classifier
 - unit tests for convicted/dismissed/missing/unrec dispositions

jjames1011:
-This pull request added a new charge type. (contempt of court) 

-Implemented a solution by adding 2 new files: 

contempt_of_court.py
test_contempt_of_court.py

and contributing to 1 existing:

charge_classifier.py

The charge classifier file now can classify charges with "contempt of court" in the name as type ContemptOfCourt which is not eligible for expungement (Consulted with Michael Zhang on this).

(I forgot to add new branch in fork before making changes and committing I will do so for future PR's) 

Closes #891


wittejm:
I dug through some additional record examples and found cases where the contempt charge is a violation or a misdemeanor, along with the cases that have no defined charge level. The charge could be eligible under these conditions (a convicted violation, or both a dismissed or convicted misdemeanor). I believe these eligibility criteria would take precedence.

We should check with @michaelzhang43 on this. If we can take this approach, it might be best to classify a charge first by its level, if it has one, before looking for "contempt of court", which would just require moving around the order of operations in the classifier to capture charge by level first. 

Similarly, we already have the Civil Offense classify method with priority also, below classification by level. If any civil offense (which we classify by having statute chapter 1-99) meets some other criterion like charge level, we should use that first since "civil offenses" themselves are omitted from the statute, and we want to match to any positive criteria whenever possible.

src/backend/expungeservice/models/charge_types/contempt_of_court.py
```diff
@@ -0,0 +1,21 @@
...
+from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+
+
+
+@dataclass(eq=False)
+class ContemptOfCourt(Charge):
+    """ This is a civil offense """
+
+    type_name: str = "Contempt of Court"
+    expungement_rules: str = (
```
wittejm: This should be a longer-form string, i.e. including that fact that it may or may not be classifiable by charge level. The details here can depend on how we decide to handle violation/misdemeanor contempt charges 

Also it can note that statutes for a contempt charge include 33015, 33055, and 33065.

src/backend/tests/models/charge_types/test_contempt_of_court.py
```diff
@@ -0,0 +1,70 @@
...
+
+from tests.factories.charge_factory import ChargeFactory
+from tests.factories.case_factory import CaseFactory
+from tests.models.test_charge import Dispositions
+
+def test_contempt_of_court_dismissed():
+    charge_dict = {
+        "case": CaseFactory.create(type_status=["Contempt of Court", "Closed"]),
+        "name": "contempt of court",
+        "statute": "33",
```
wittejm: The first two digits are just the chapter, and it looks like the typical statutes for contempt are 33065, 33055, and 33015, so we can use those in our test cases. 

Chapter 33 https://www.oregonlaws.org/ors/chapter/33 covers a bunch of non-charge related stuff, but it seems possible there would be other charges in there that aren't contempt.

michaelzhang43:
Hi all,
Sorry for the late reply on this. 

Contempt of Court is simple. Technically. It's considered a civil matter between the court and the individual. It is not considered a crime.  So they are not considered for any reason in expungement analysis. They are as traffic violations EVEN IF THEY ARE LABELED MISDEMEANORS OR FELONIES. They do not block charges, and they are not eligible for expungement. 

For the curious, Cntrl+F "contempt" in the annotations of 137.225 
https://www.oregonlegislature.gov/bills_laws/ors/ano137.html

wittejm:
Cool, thanks for the extra detail on this @michaelzhang43

wittejm:
replaced by #928

PR #918, opened 2020-03-04
Add loggedIn key to System interfaces

ARodriguezHacks:
This Pull Request attempts to fix a browser refresh problem in which the isAdmin key becomes undefined.

Issue should be resolved aside from a minor lag in the user interface, noticeably in the Header component where the Admin button doesn't appear right away upon refresh.


wittejm:
replaced by #1041

PR #926, opened 2020-03-05
Update with unrecognized taffic violation test

KorynLA:
Associated with #861


src/backend/tests/models/charge_types/test_traffic_offenses.py
```diff
@@ -86,6 +86,15 @@ def test_no_dispo_violation_is_not_type_eligible():
     )
     assert not charge.blocks_other_charges()
 
+def test_unreognized_violation_is_not_type_eligible():
```
KentShikama:
```suggestion
def test_unrecognized_violation_is_not_type_eligible():
```
typo

PR #928, opened 2020-03-05
Update contempt test statute

jjames1011:
Updated statute for test_contempt_of_court.py file 

In regard to issue #891


src/backend/expungeservice/models/charge_types/contempt_of_court.py
```diff
@@ -0,0 +1,21 @@
+from dataclasses import dataclass
+
+from expungeservice.models.charge import Charge
+from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+
+
+
+@dataclass(eq=False)
+class ContemptOfCourt(Charge):
+    """ This is a civil offense that is always ineligible regardless of level / conviction status"""
```
wittejm: We don't use class docstrings anywhere else, so I would remove this but use it as the `expungement_rules`,

jjames1011: Done with latest commit

wittejm:
See my comment on your other PR. Since this branch is out of sync with master and now has some conflicts, you have a few options.

One is doing some `git squash` and `rebase` wizwaz. This blog post describes approximately the way I would do it: 
https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/

But honestly, I would just do it a simpler way instead because your PR here is new files and otherwise just a few new lines. So, similarly to what I described in your other PR:

- Set aside all the work you want to keep 
- check out `master`, and blow it away `git reset --hard upstream/master`
- destroy and re-create the feature branch
- re-apply your work manually
- commit and force push the branch to your fork

PR #931, opened 2020-03-05
Add loggedIn key to System interfaces

### LINKED ISSUE #862: Persist the isAdmin state when navigating around the app
We save an `isAdmin` flag in redux when logging in. When a user navigates to an app page by changing the url, the page component loads as expected, but the `isAdmin` flag is reset to `false`, hiding admin-only functions from the logged-in admin. 

I'm not sure if fixing that behavior is something redux is fit for, by persisting the `isAdmin` flag. But we need a way to consistently determine whether the logged in user is admin, whenever the page is loading. 

Maybe the flag could be in the login cookie?

wittejm:
copies changes from @ARodriguezHacks 's PR #918 

Add the `redux-persist` dependency. 

Extend `refreshLocalAuth` to verify isAdmin value. 

This copy is just to replace the git commit history with a rebase on master. 
closes #862

### DIFF
--- a/src/frontend/src/redux/store.tsx
+++ b/src/frontend/src/redux/store.tsx
@@ -3,6 +3,8 @@
 // https://redux.js.org/recipes/usage-with-typescript
 import { configureStore, getDefaultMiddleware } from '@reduxjs/toolkit';
 import { combineReducers } from 'redux';
+import { persistStore, persistReducer } from 'redux-persist';
+import storage from 'redux-persist/lib/storage';
 
 // Reducers:
 import { usersReducer } from './users/reducer';
@@ -10,17 +12,36 @@ import { recordsReducer } from './records/reducer';
 import { systemReducer } from './system/reducer';
 import { groupsReducer } from './groups/reducer';
 
+const persistConfig = {
+  key: 'root',
+  storage,
+  version: 1,
+  whitelist: ['system']
+};
+
+const systemConfig = {
+  key: 'system',
+  storage,
+  whitelist: ['isAdmin']
+};
+
 const rootReducer = combineReducers({
-  system: systemReducer,
+  system: persistReducer(systemConfig, systemReducer),
   records: recordsReducer,
   users: usersReducer,
   groups: groupsReducer
 });
 
+const pReducer = persistReducer(persistConfig, rootReducer);
+
 const store = configureStore({
   reducer: rootReducer,
-  middleware: [...getDefaultMiddleware()]
+  middleware: getDefaultMiddleware({
+    serializableCheck: false
+  })
 });
+
+export const persistor = persistStore(store);
 
 export type AppState = ReturnType<typeof rootReducer>;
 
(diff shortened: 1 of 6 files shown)


src/frontend/src/redux/system/types.ts
```diff
@@ -8,11 +8,13 @@ export const LOG_OUT = 'LOG_OUT';
 
 interface LogInAction {
   type: typeof LOG_IN;
+  loggedIn: boolean;
```
KentShikama: What is this `loggedIn` boolean used for?

src/frontend/src/redux/store.tsx
```diff
@@ -3,25 +3,46 @@
...
+
 const rootReducer = combineReducers({
-  system: systemReducer,
+  system: persistReducer(systemConfig, systemReducer),
   records: recordsReducer,
   users: usersReducer,
   groups: groupsReducer
 });
 
+const pReducer = persistReducer(persistConfig, rootReducer);
```
KentShikama: I would just spell it out `const persistReducer`

src/frontend/src/redux/system/actions.ts
```diff
@@ -37,9 +38,19 @@ export function logOut(): any {
...
+    return apiService(dispatch, {
+      url: '/api/auth_token',
+      method: 'post'
+    }).then((response: any) => {
+      dispatch({
+        type: LOG_IN,
+        loggedIn: response.data.loggedIn
+      });
+      history.push('/oeci');
+    });
```
KentShikama: Why do we need to hit the backend again here?

src/frontend/src/redux/system/actions.ts
```diff
@@ -37,9 +38,19 @@ export function logOut(): any {
   };
 }
 
-export function refreshLocalAuth() {
-  return {
-    type: LOG_IN
+
+export function refreshLocalAuth(): any {
+  return (dispatch: Dispatch) => {
+    return apiService(dispatch, {
```
wittejm: This post request is missing a body which for this endpoint should normally be email/password. But we don't hold onto the password in the frontend so there's no way to quietly use call method. 

Agreeing with Kent: it's not clear why this backend call is required

KentShikama: I don't like the phrasing "agreeing with Kent" as it makes it sound like I was making a rhetorical question, which I try hard not to do in a PR review. If you were a student in a classroom and a fellow student asked the professor "what is a land-value tax?", it would be odd to say "I agree with that question". This is a fuzzy line but I want to keep the distinction between "Why do we need to hit the backend again here?" versus "I don't think we need to hit the backend again here".

wittejm: Fair. i did mean, "I have the same questions as Kent"

KentShikama: But now that you believe we don't need a backend call...

wittejm: I'm honestly don't know. I naively tried eliminating the the calling code and I didn't figure out a way to get the persist feature to work, so I just don't know what's going on or required.

KentShikama: Sounds good; I guess we'll wait for a reply :P

PR #937, opened 2020-03-06
Add currency formatting service.

### LINKED ISSUE #780: Fix monetary formatting for search page to match summary
On the search page, the case component displays the balance in plain numbers (0, for example). It needs to match what is on the static design as well as the format in the summary ($1.23).

keppel2:
Fix #780 .

### DIFF
--- a/src/frontend/src/service/currency-format.test.ts
+++ b/src/frontend/src/service/currency-format.test.ts
@@ -1,5 +1,5 @@
 import currencyFormat from './currency-format.ts';
-describe('API SERVICE TEST', () => {
+describe('CURRENCY FORMAT SERVICE TEST', () => {
   it('formats currency', () => {
     var str = currencyFormat(145);
     expect(str).toEqual('$145.00');
--- a/src/frontend/src/service/currency-format.test.ts
+++ b/src/frontend/src/service/currency-format.test.ts
@@ -1,0 +1,13 @@
+import currencyFormat from './currency-format.ts';
+describe('API SERVICE TEST', () => {
+  it('formats currency', () => {
+    var str = currencyFormat(145);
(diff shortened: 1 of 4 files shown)


src/frontend/src/service/currency-format.test.ts
```diff
@@ -0,0 +1,13 @@
+import currencyFormat from './currency-format.ts';
+describe('CURRENCY FORMAT SERVICE TEST', () => {
+  it('formats currency', () => {
+    var str = currencyFormat(145);
```
wittejm: Recommend using `const` instead of `var`
Also I'd suggest renaming the variable to `currencyString`

src/frontend/src/service/currency-format.ts
```diff
@@ -0,0 +1,8 @@
+export default function currencyFormat(input: number) {
+  var wholePart = Math.trunc(input);
+  var wholeStr = wholePart.toLocaleString();
+  var fractionPart = Math.abs(input) % 1;
+  var fractionStr = fractionPart.toFixed(2).substring(1);
+  var rt = '$' + wholeStr + fractionStr;
```
wittejm: `const`s here too , also I suggest renaming the rt variable to `currencyString`

keppel2: Done

src/frontend/src/service/currency-format.test.ts
keppel2: I reuse the str here.

keppel2:
All JEST tests are passing.

src/frontend/src/service/currency-format.test.ts
keppel2: I'll rename it though.

Issue #846, opened 2020-02-15
Create a unit test for the summarizer


wittejm:
PR #841 introduced a behavior change in the record summarizer that could have a fairly simple unit test:
Enumerate the charge types, and write for each one: `assert charge.hidden_from_record_summary()  == correct_value`

PR #938, opened 2020-03-06
This PR is in reference to #846

jjames1011:
PR #841 introduced a behavior change in the record summarizer which should have corresponding tests. 

**This Pull Request:** 

-Adds Unit Tests for the summarizer by testing against the charge types for expected behavior (defined within charge.py or the specific charge type file) 

**Implemented a solution by:**

Created test_record_sumarizer_by_charge_type.py:

Iterates over each charge type and check their hidden_in_record_summary() attribute for expected output.


wittejm:
Thanks for the PR! there's a bit of work to do here though:

You have some git snarl in here including your commits from your other feature, and it's out of sync with master. Since you're starting a few feature, I recommend you:
- set aside your Contempt of Court code somewhere (I'll expand on that on your Contempt PR)
- set aside the new test file you wrote for this pr, plus any other code you've written but not checked in 
- blow away and replace your master branch: while master is your active branch, `git reset --hard upstream/master`
- destroy and re-create the feature branch `test_record_summarizer` 
- copy the new unit test code into place there. Don't include the Contempt code, see below
- update your work to account for the recent changes in `master`.   
- commit and force-push to update the PR.   you'll need to force push onto your github fork with `git push -f origin test_record_summarizer` , because you're overwriting the commit history.

A recent change in `master` refactored our test code to no longer use `charge_dict`, and instead we just pass values in `ChargeFactory.create()`, so you'll need to adjust your code accordingly.

Since your Contempt feature is not yet merged into `master`, it won't appear in this PR, and that's fine. In general, your work in some branch A to address an issue shouldn't appear in a different branch B that you're working on, until after branch A has been merged into `upstream/master`, and and then pulled from there into branch B with `git pull -r upstream master` 

Also, a good name for a PR is just a description for what the feature change is, in this case e.g. "Add unit tests for the record summarizer"

Otherwise, the actual approach in the unit tests looks good to me.

PR #942, opened 2020-03-08
Added new test file test_record_summarizer_by_charge_type.py. Resolves #846

### LINKED ISSUE #846: Create a unit test for the summarizer

jjames1011:
This PR resolves issue #846 by:

Added a new test file: 

test_record_summarizer_by_charge_type.py

which tests against each charge type's hidden_in_record_summary() method for expected output which is declared in models/charge.py or the specific charge type file. 

Creating new PR to fix merge issues and update file to include requested updates...

"A recent change in master refactored our test code to no longer use charge_dict, and instead we just pass values in ChargeFactory.create(), so you'll need to adjust your code accordingly."

### DIFF
--- a/src/backend/tests/models/test_record_summarizer_by_charge_type.py
+++ b/src/backend/tests/models/test_record_summarizer_by_charge_type.py
@@ -1,0 +1,245 @@
+from datetime import date as date_class
+from expungeservice.models.charge_types.civil_offense import CivilOffense
+from expungeservice.models.charge_types.duii import Duii
+from expungeservice.models.charge_types.felony_class_a import FelonyClassA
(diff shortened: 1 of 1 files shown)


wittejm:
Updated these test cases to follow the recent changed in charge types.

    DUII charges are always Ambiguous (the charge classifier returns DUII and Diverted);
    similarly with Subsection6, which can be classified by Level also.
    Deleted Manudel , Added MarijuanaEligible
    Added ContemptOfCourt

PR #943, opened 2020-03-10
Add ability to display multiple dates for each charge

KentShikama:
`charge.type_eligibility` represents the type eligibility of a charge. `charge.expungement_result.type_eligibility` now represents the merged type eligibility of a charge post-merge as a single charge can be a superposition of multiple charge subclasses.

### DIFF
--- a/src/backend/expungeservice/models/helpers/record_merger.py
+++ b/src/backend/expungeservice/models/helpers/record_merger.py
@@ -1,7 +1,17 @@
-from typing import List, Dict, Set
+from dataclasses import replace
+from datetime import date
+from typing import List, Dict, Set, Optional
 
 from expungeservice.models.ambiguous import AmbiguousRecord
-from expungeservice.models.expungement_result import TimeEligibility
+from expungeservice.models.charge import Charge
+from expungeservice.models.expungement_result import (
+    TimeEligibility,
+    ExpungementResult,
+    TypeEligibility,
+    EligibilityStatus,
+    ChargeEligibility,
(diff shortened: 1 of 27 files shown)


src/backend/expungeservice/models/helpers/record_merger.py
```diff
@@ -16,14 +26,108 @@ def merge(
...
+            return EligibilityStatus.INELIGIBLE
+        else:
+            return EligibilityStatus.NEEDS_MORE_ANALYSIS
+
+    @staticmethod
+    def merge_time_eligibilities(time_eligibilities: Optional[Set[TimeEligibility]]) -> Optional[TimeEligibility]:
+        if time_eligibilities:
+            status = RecordMerger.compute_time_eligibility_status(time_eligibilities)
+            reason = ", ".join([time_eligibility.reason for time_eligibility in time_eligibilities])
+            date_will_be_eligible = date.max  # TODO: Fix
```
KentShikama: @wittejm I believe the frontend does not use this value.

wittejm: Yep, I recall putting it in there thinking "hmm it seems like the frontend should probably be using this at some point" but that doesn't seem like good reasoning in retrospect.

wittejm:
This is pretty hard to read:

diverted_2_times

KentShikama:
@wittejm Addressed with https://github.com/codeforpdx/recordexpungPDX/pull/949

PR #945, opened 2020-03-11
Use binary psycopg2

keppel2:
This allows me to run `pipenv install` in `src/backend`. I had to do `brew upgrade piping`.


keppel2:
s/piping/pipenv

keppel2:
That makes commit hooks harder. Also, the docker method requires a hack to share the repo.

keppel2:
Seems opposite of usual dev/prod divide.

keppel2:
I don't see the problem with this. It involves c code that most of us don't even know. The binary should be fine. The docker solution is a good long term option for certain things, but this should be ok.

KentShikama:
Could you expand a bit on what you mean by the "docker method"? I was just referring to the fact that people seem to be happy to just run `make dev_up` to launch the docker containers for any type of dev work. What is your context for wanting to run `pipenv install` natively?

I'm not sure if it is still relevant but apparently there are reliability concerns with the binary package:

> 
> Unfortunately, after the packages were released, it was reported of occasional segfaults of Python processes using Psycopg from the Wheel package. This was traced to the use of the Python ssl library concurrently with Psycopg opening connections, for instance in a multithread program opening concurrently https resources and database connections. The problem seems caused by a non-reentrant OpenSSL initialization function (unfortunately invoked by libpq at every connection) and the fact that the Python process ends up binding two different versions of the libssl library: the system one via the Python ssl library (e.g. imported by urllib or requests) and the Wheel one imported by the libpq, in turn imported by psycopg2.
> 
> While the problem doesn't affect many users, a library behaving unreliably in combination with part of the stdlib is a situation less than optimal. The workaround is to force installing Psycopg from source, but this must be specified explicitly in the project dependencies (e.g. using the --no-binary flag in the pip command line or in the requirements.txt file); the Python packaging system doesn't really have a way to declare something like "install a package preferably from source"... so we had to create one ourselves. - https://www.psycopg.org/articles/2018/02/08/psycopg-274-released/.

keppel2:
@aemrich had a different error with psycopg2 (on `master`). Involved build dependencies. Mine was in the c code.

keppel2:
In this case, it is to run `util/generate_charge_types_doc.py`.

keppel2:
@KentShikama @aemrich need review to merge.

keppel2:
@KentShikama Don't think we'll hit that problem. Review?

PR #946, opened 2020-03-11
Change handling of statutes with subsections in person felony

KentShikama:
Since the charge is classified as a Class B Felony, we assume even if the subsection is missing that it is a Person Felony.


wittejm:
I don't see why we want this change? If the subsection is mismatched from a person felony statute, it might not be a person felony, which is why they need more analysis.

Issue #936, opened 2020-03-06
Add Staging Environment and Deploy to Staging Steps


loganballard:
Currently our deployment process involves a lot of manual steps, and production is run out of a home directory.  There is also not a staging environment.

### This task is done when:
 - Production is run out of a shared directory
 - Staging environment exists and mimics production (including database)
 - Deployment steps for production and staging are easy to understand and can be run by a dev with access to the production server
 - Staging environment can be taken up and down on demand

kenichi:
see https://github.com/codeforpdx/recordexpungPDX/compare/master...kenichi:devops

goals of this branch:
* make initial developer setup and running as simple as possible (`make up initdb`)
* make staging, production, etc deploys as simple as possible (`cd src/ops; make staging`)

kenichi:
>  - Production is run out of a shared directory

self-contained single-image deploys allow for horizontal scaling, easy rollback, platform agnostic-ism

> - Staging environment exists and mimics production (including database)

✅ though they have separate databases, `record_sponge_db` (prod) and `record_sponge_db_staging` (staging)

> - Deployment steps for production and staging are easy to understand and can be run by a dev with access to the production server

make targets in src/ops/Makefile (`staging` or `prod`), docs in src/ops/README.md and doc/, and comments in Makefile.

> - Staging environment can be taken up and down on demand

at this point, the method of doing this would be to start or stop the `staging` docker container on the recordsponge.com host.

wittejm:
sweet. issue sounds completed to me.

PR #954, opened 2020-03-12
Deploy rework

loganballard:
addresses #936 

Note that the only thing I've changed here is the Makefile, whereas `staging` branch on recordexpungPDX is significantly behind `master`, so I merged those changes in.  I can branch off of staging instead if we don't want the latest changes from `master` as well. 

Open question that I have is that I'd like to try this out on just the staging branch first before running the production deployment (moving over from kent's home dir to the shared directory), so I haven't made any changes to the production deployment process yet.  I'd like to get a bit of feedback first.

The basic gist of this is that we now have a directory structure that looks like:

```
root@ubuntu-s-3vcpu-1gb-nyc3-01:/deployments# pwd
/deployments
root@ubuntu-s-3vcpu-1gb-nyc3-01:/deployments# ls -alh
total 16K
drwxr-xr-x  4 root root      4.0K Mar 12 02:48 .
drwxr-xr-x 24 root root      4.0K Mar 12 02:48 ..
drwxrwx---  3 root deployers 4.0K Mar 12 02:50 prod
drwxrwx---  3 root deployers 4.0K Mar 12 02:51 staging
```

Where the `prod` and `staging` directories contain a `recordexpungPDX` directory with source code. The group `deployers` contains:

```
root@ubuntu-s-3vcpu-1gb-nyc3-01:/deployments# cat /etc/group | grep deployers
deployers:x:1004:logan,jordan,kent,kenichi
```

Then we will run our deployments out of prod and staging directories.

-Logan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,19 +116,19 @@ $(REQUIREMENTS_TXT):
 
 .ONESHELL:
 deploy_staging_update_repo:
-	cd ~/recordexpungPDX/
+	cd /deployments/staging/recordexpungPDX
 	git reset --hard
 	git checkout staging
 	git pull origin staging:staging
 
 .ONESHELL:
 deploy_staging_backend: deploy_staging_update_repo
-	cd ~/recordexpungPDX/src/backend/
+	cd /deployments/staging/recordexpungPDX/src/backend/
 	kill $(ps -ef | grep -v grep | grep 127.0.0.1:3032 | awk '{print $2}')
-	$(shell tr '\n' ' ' < ~/recordexpungPDX/config/expungeservice/expungeservice.staging.env) nohup pipenv run uwsgi --socket 127.0.0.1:3032 --module expungeservice.wsgi &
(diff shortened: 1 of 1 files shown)


wittejm:
I just force-updated staging to match master, so that your PR could just be your own commits. But instead I think this PR should just go directly to master (close this PR and open a new one).

I tried messing around with the make target but I'm getting permission error (from my own user account) trying to write to  /deployments/staging. Doesn't look right since the groups thing looks set up fine.

That's all I have for now. The approach looks good overall. I can keep poking around with it

kenichi:
n.b. i had also created a group for deploys, sorry i left that out of the notes:

```
$ grep sponge /etc/group
recordsponge:x:1003:kenichi,kent,jordan
```

kenichi:
major updates to deploy process and local dev environment make this out of date. env files are now shared by the `recordsponge` group in /etc/recordsponge, and the new make targets in /src/ops/Makefile expect this for prod and staging. branching strategy is next!

PR #961, opened 2020-03-20
Extract Record Creator from search endpoint

KentShikama:
Please review commit by commit.

Before this PR, we weren't really testing the bulk of our search endpoint high-level logic. Now the `CrawlerFactory` runs through most of the same logic the search endpoint in production does: they both run `build_record` in `RecordCreator`.


wittejm:
Also this is fun:

line_changes

PR #962, opened 2020-03-20
Minor performance improvements in Expunger

KentShikama:


KentShikama:
- Short circuiting means the comparison is not executed.
- `ids` are apparently faster than comparing the entire dataclass.

PR #963, opened 2020-03-22
Add charge type for dismissed criminal charges

KentShikama:
We do not want to have to ask users questions for manu/del and subsection 6 if the charge is dismissed as they are type eligible. This will also boost performance as we will not have as many ambiguous records.

~~We assume that for unclassified charges that are dismissed we can just say they are eligible but this may be a dangerous assumption.~~

We remove ambiguity when Duii is a non-dismissal as they are known to be ineligible (or unknown).

### DIFF
--- a/src/backend/expungeservice/models/helpers/charge_classifier.py
+++ b/src/backend/expungeservice/models/helpers/charge_classifier.py
@@ -51,15 +51,19 @@ class ChargeClassifier:
         yield ChargeClassifier._traffic_crime(self.statute, self.level, self.disposition)
         yield ChargeClassifier._parking_ticket(self.violation_type)
         yield ChargeClassifier._violation(self.level)
-        yield ChargeClassifier._dismissed_charge(self.disposition)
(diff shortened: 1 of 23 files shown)


src/backend/expungeservice/models/helpers/charge_classifier.py
```diff
@@ -123,21 +136,23 @@ def _subsection_6(section, level, statute):
             return [Subsection6] + ChargeClassifier._classification_by_level(level, statute)
```
wittejm: Could you check for a dismissal here instead? And similar with Manudel?

wittejm:
In fact, an unclassified charge should probably be considered a record error just as much as an unrecognized disposition is currently. They'll be resolved the same way by the user followup question

KentShikama:
After this PR is merged, we may want to review and "redistribute" some of the `expungement_rules` strings to the appropriate charge type. Specifically some strings may contain assumptions that the class is still handling dismissed charges.

PR #964, opened 2020-03-22
Fix case parser when disposition table is longer than 5 elements

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/952

The HTML fixture was generated as follows:

```
download case HTML from OECI
strip out between <head></head>
strip <tbody>
strip </tbody>
strip onload="Javascript:checkRefresh()"
replace href=\"[^"]*"> with href="#">
replace href=\'[^']*'> with href="#">
replace <FIRST NAME> with Joe
replace <MIDDLE NAME> with
replace <LAST NAME> with Doe
replace <DATE OF BIRTH> with 1900
replace <GENDER> with Gender
replace <RACE> with Race
replace <STREET ADDRESS> with 123 EXAMPLE STREET
replace <CASE NUMBER> with 123456789
replace <OFFICER NAME> with Example Officer
strip xmlns:[^\"]*".*?"
```


KentShikama:
@NickSchimek By any chance did you use a script to sanitize/anonymize the HTML case fixtures?

src/backend/tests/test_crawler_expunger.py
```diff
@@ -198,3 +198,16 @@ def test_probation_revoked_affects_time_eligibility(record_with_revoked_probatio
...
+
+
+@pytest.fixture
+def record_with_odd_event_table_contents():
+    return CrawlerFactory.create(
+        record=JohnDoe.SINGLE_CASE_RECORD, cases={"CASEJD1": CaseDetails.CASE_WITH_ODD_EVENT_TABLE_CONTENTS,},
+    )
+
+
+@pytest.mark.skip()
```
wittejm: Why is this new test skipped?

KentShikama: Because it fails right now. This doesn't fix the issue. We're doing TDD. Write a failing test (this PR). And then fix it.

NickSchimek:
> @NickSchimek By any chance did you use a script to sanitize/anonymize the HTML case fixtures?

No, I did not. I used the find all and replace functionality on the text editor. I figured this isn't something that should need to be done that often, and the inconsistent display of information from OECI would still require a human verifier.

src/backend/tests/test_crawler_expunger.py
NickSchimek: I'd say it's generally good practice not to skip it if you are doing TDD. Not skipping it has a few benefits. The obvious benefits are that you see it fail first, and then pass when fixed. 

When the failed test is pushed up to a repo as it has been here then the integration test will show that this test is actually failing which is a good thing, because it adds confidence for the reviewers that this test is actually testing the issue that will be fixed. The next commit would then show a passing test further adding confidence for the reviewers that this is a valid fix.

KentShikama: Yeah I should have marked this as `pytest.mark.xfail` instead of `skip`.

PR #970, opened 2020-03-27
Roll back reach-scripts version

wittejm:
The react-scripts version bump from 3.2 to 3.4 apparently breaks our dev frontend unless the latest version of node is running.

In dev, we're using the latest node image available on dockerhub, which is 13.11. But the latest actual version of node is 13.12, and is not supported yet on dockerhub

For now, fix this error by rolling back the react-scripts version. We could also make our own custom Dockerfile for node to use the latest version.

Also we're starting to specify our node version in the Dockerfile, because this should be explicit.

### DIFF
--- a/src/frontend/Dockerfile.dev
+++ b/src/frontend/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM node:latest as build
+FROM node:13.11 as build
 RUN mkdir /usr/src/app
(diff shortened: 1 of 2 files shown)


KentShikama:
Superseded by https://github.com/codeforpdx/recordexpungPDX/pull/973

PR #973, opened 2020-03-29
Bump down react-scripts to 3.3.0

KentShikama:
For whatever reason react-scripts 3.4.0 fails to launch in our docker container. We do not want to downgrade back to 3.2.0 as it contains vulnerabilities. A middle ground is to resolve to an older version of react-dev-utils and use react-scripts 3.3.0.


wittejm:
This works when I build locally. But I notice also that dockerhub's node image 13.12 is usable here because it got posted on their site just in the last 2 days. 

react-scripts 3.4.0 was working fine with node 13.12 running native, and I just tried it here in our docker container and it works fine too.

But, then, just to check that computers are fundamentally logical, I tried changing the node version to 13.11 and react-scripts to 3.4.0, which is the combo that was busted for me the other day. But it works fine. 

I also tried node 13.11 , react-scripts 3.3.0, and without `react-dev-utils`. That works today too, although it wasn't working for me the other day. 

Either computers are fundamentally not logical, or docker is much less magical than I had been led to believe. 

What did you add `react-dev-utils` for here? Was that necessary for getting react-scripts 3.3.0 to work?

KentShikama:
@wittejm Did you remove all images before testing the failing versions?

KentShikama:
> What did you add react-dev-utils for here? Was that necessary for getting react-scripts 3.3.0 to work?

Indeed. See https://github.com/facebook/create-react-app/issues/8499

Issue #972, opened 2020-03-29
Feature: User follow-up questions


wittejm:
After user runs the record search, any charges requiring disambiguation show a radio-button form to provide further information. 

(more detailed description for feature can go here)

wittejm:
Mock-up from @hmarcks : 

![user-submitted-data](https://user-images.githubusercontent.com/2104990/77858011-c1a42200-71b5-11ea-8796-86df021b834d.png)

wittejm:
In answer to your in-image comment, the entire analysis does need to run again because these ambiguities might impact time eligibility of other charges. With that in mind, I think the user should be expected to answer all the follow-up questions before they can re-run the analysis. It will take less time because there's no fetching from OECI this time. 

But, should all the answers be required? What if the user doesn't know yet what one of the answers was, so they want to leave that ambiguity untouched, but re-run with the answers they have?

And between these two options, should the submit button still go in-line or be at the bottom?

wittejm:
In fact, the running time for analysis would typically be pretty short unless the the record is really long. It also varies depending on the number of ambiguities in the record.

 I'd guess the running time can be long enough that it shouldn't be forced to re-run after each answer, but short enough that it wouldn't be terrible if the user could re-run it after submitting only one or some answers, if they wanted to.

hmarcks:
Thanks for the feedback @wittejm 

**1) If the user doesn’t know the answer**

Do you think we'd want to encourage an answer? Since the analysis won’t be accurate without an answer, at least they could see a potential scenario. And it's not locked in with the ability to edit.

So that’s an argument for keeping a single Submit button, but not actually requiring all. Just an idea, not saying it’s better than a button for each.

**2) Rerun after each submit?**

To ensure the user doesn’t have to wait for analysis between answering each question, I see a couple options:

a) Could we re-run after each submit, but keep the UI visible so they can locate any other outstanding questions?

b) Remove individual Submit buttons and provide a single button in the banner or in a sticky header/footer.

wittejm:
> Do you think we'd want to encourage an answer?

Good point, making a good reason to not have an "I don't know" option.

> (although would multiple questions be very common?)

[...]

PR #976, opened 2020-03-30
Allow frontend to start displaying questions for ambiguous charges

KentShikama:
This is the backend work for https://github.com/codeforpdx/recordexpungPDX/issues/972.

### DIFF
--- a/src/backend/expungeservice/crawler/crawler.py
+++ b/src/backend/expungeservice/crawler/crawler.py
@@ -98,4 +98,4 @@ class Crawler:
             )  # TODO: Log error if format is not correct
             ruling = disposition_data.get("ruling")
             charge["disposition"] = Disposition(date, ruling, "amended" in disposition_data["event"].lower())
-        return ChargeCreator.create(charge_id, **charge)
+        return ChargeCreator.create(charge_id, **charge)[0]  # TODO: Fix
--- a/src/backend/expungeservice/crawler/crawler.py
+++ b/src/backend/expungeservice/crawler/crawler.py
@@ -2,12 +2,11 @@ from concurrent.futures.thread import ThreadPoolExecutor
 from copy import copy
 from dataclasses import replace
 from itertools import product
-from typing import List
+from typing import List, Tuple, Optional
 
 import requests
 from datetime import datetime
 
-from expungeservice.models.case import Case
 from expungeservice.models.helpers.charge_creator import ChargeCreator
 from expungeservice.models.disposition import Disposition
 from expungeservice.crawler.parsers.param_parser import ParamParser
@@ -16,6 +15,7 @@ from expungeservice.crawler.parsers.record_parser import RecordParser
 from expungeservice.crawler.parsers.case_parser import CaseParser
 from expungeservice.crawler.request import *
 from expungeservice.models.ambiguous import AmbiguousCharge, AmbiguousCase
+from expungeservice.models.record import Question
 
 
 class Crawler:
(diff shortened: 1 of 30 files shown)


src/backend/expungeservice/models/helpers/charge_classifier.py
```diff
@@ -137,7 +155,10 @@ def _subsection_6(section, level, statute):
             "163165",  # ( ineligible if under subection(1)(h) ; Assault in the third degree of a minor 10 years or younger)
         ]
         if section in conditionally_ineligible_statutes:
-            return [Subsection6] + ChargeClassifier._classification_by_level(level, statute)
+            question_string = "TODO: FIXME: Is this charge eligible?"
```
wittejm: "Does this charge meet the eligibility requirements listed in 137.225(6)? "
options = ["Yes", "No"]

src/backend/expungeservice/models/helpers/charge_classifier.py
```diff
@@ -85,47 +86,64 @@ def _drug_crime(statute, section, name, level):
...
+            return AmbiguousChargeTypeWithQuestion([MarijuanaEligible])
 
     @staticmethod
     def _manufacture_delivery(name, level, statute):
         if any([keyword in name for keyword in ["delivery", "manu/del", "manufactur"]]):
             if "2" in name:
                 if level == "Felony Unclassified":
-                    return [FelonyClassA, FelonyClassB]
+                    question_string = "Was the charge for an A Felony or B Felony?"
+                    options = ["A Felony", "B Felony"]
```
wittejm: could be a C felony?

KentShikama: For Schedule 2, I don't think so.

src/backend/expungeservice/models/helpers/charge_classifier.py
KentShikama: We eventually want to split it out but I want to get the "framework" down first.

src/backend/expungeservice/crawler/crawler.py
```diff
@@ -35,7 +35,9 @@ def login(self, username, password, close_session=False) -> bool:
 
         return Crawler.__login_validation(self.response)
 
-    def search(self, first_name, last_name, middle_name="", birth_date="") -> List[AmbiguousCase]:
+    def search(
+        self, first_name, last_name, middle_name="", birth_date=""
+    ) -> Tuple[List[AmbiguousCase], List[Question]]:
```
wittejm, line 40: To support interactive updates as the user submits one or more answers, the search endpoint response needs to also provide the Answers that they have submitted so far. In the first response from the search endpoint, there won't be any answers attached but the response structure is the same. Up to you if you think that's next-PR material.

KentShikama: The frontend will continue to hit the disambiguate endpoint which returns the questions (with answers) in the record summary object.

KentShikama:
@wittejm I think this is ready but it would be nice to get another pair of eyes on https://github.com/codeforpdx/recordexpungPDX/pull/976/files#diff-1643e6866424073bbb3377c0d2036383 for any leaks.

KentShikama:
There is still a pending TODO for the R&J exception: I'm waiting on @michaelzhang43 for the question. I don't think it blocks the PR merge.

src/backend/expungeservice/crawler/crawler.py
wittejm: oh right!

PR #994, opened 2020-04-03
Frontend various refactoring

wittejm:
Just some cleanup, details in the commit messages.


wittejm:
Similar nesting of components can be useful for the Admin feature code too.

Issue #1004, opened 2020-04-06
Romeo and Juliet exception question to user


michaelzhang43:
For possible Romeo and Juliet exceptions, prompt user with the following question:

"
Select 'Yes' if all the following are true:
1. The charges were for statutory rape, i.e. the only reason consent was not present is due to the age of the victim
2. You were under 23 years old when the act occurred
2. The victim was less than five years younger than you when the act occurred
"

### This task is done when:
 -
 -

PR #1005, opened 2020-04-07
Coerce label for R&J exception to be possibly eligible

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/1004

Before question is asked:
Screen Shot 2020-04-07 at 11 19 40 AM

After "Yes" is selected:
Screen Shot 2020-04-07 at 11 21 50 AM

After "No" is selected:
Screen Shot 2020-04-07 at 11 22 36 AM

### DIFF
--- a/src/backend/expungeservice/charge_classifier.py
+++ b/src/backend/expungeservice/charge_classifier.py
@@ -21,7 +21,11 @@ from expungeservice.models.charge_types.person_felony import PersonFelonyClassB
 from expungeservice.models.charge_types.contempt_of_court import ContemptOfCourt
 from expungeservice.models.charge_types.civil_offense import CivilOffense
 from expungeservice.models.charge_types.unclassified_charge import UnclassifiedCharge
-from expungeservice.models.charge_types.sex_crimes import SexCrime, RomeoAndJulietIneligibleSexCrime
+from expungeservice.models.charge_types.sex_crimes import (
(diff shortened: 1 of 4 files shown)


michaelzhang43:
Just for "Possibly Eligible," Can we remove all the text after (6)(f) so that it just reads 
"Possibly meets requirements under 137.225(6)(f)"

michaelzhang43:
And for "Yes" we can remove the word "Contact"

KentShikama:
@michaelzhang43 Two screenshots have "Possibly Eligible", which one are you referring to?

KentShikama:
Note that the "type reason" string for the "before" screenshot is (auto)generated as a merge of the "type reason" strings of the disambiguated versions.

KentShikama:
Are you afraid of getting unnecessary emails because people decide to take action before answering the question? I think in general we need to change the "before" text (including for manu/del) as the diamond was a good temporary solution but it is still a bit confusing.

michaelzhang43:
I'm not worried about additional emails, but that the interface would be confusing

KentShikama:
I have fixed the typo:

Screen Shot 2020-04-07 at 11 35 04 AM

Let us discuss a better UI for the "before" states in general in a different PR/issue as I think we can come up with a better generic fix.

PR #1013, opened 2020-04-08
Frontend questions feature

wittejm:
This behavior is simpler than the current mockups in #972 , because it  updates the record immediately on clicking a radio button option for each question. We should talk about this in our meeting. 

The current design only supports the possibility of one question per charge. This is true of the backend code also.

Comment on the code for building the banner:
- It doesn't duplicate case IDs.
- if there are multiple charges with questions on a single case, the list will show that case as "unanswered" until all the attached questions are answered, not just some of them. This is currently untested because I didn't have an example record on hand (extensive mock records for good offline testing is a forthcoming feature)

### DIFF
--- a/src/frontend/src/components/RecordSearch/Record/Question.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/Question.tsx
@@ -9,39 +9,45 @@ interface Props {
   question?: QuestionData;
 }
 
-interface State {
-  selected_answer: string
-}
+class Question extends React.Component<Props> {
 
-class Question extends React.Component<Props, State> {
-
-  state = {
-    selected_answer: (
-    this.props.question ?
-    this.props.question.selected_answer :
-    "")
+  handleRadioChange = (e: React.BaseSyntheticEvent) => {
+    store.dispatch(selectAnswer(e.target.name, e.target.value))
   };
 
   render() {
-    return ( this.props.question ?
-      (
-        <div>
-          <div>Question</div>
-          <button
-            onClick={() => {
-              console.log("clicked");
-              store.dispatch(selectAnswer("15PK267144-1", "clicked for jordan"));
-              store.dispatch(selectAnswer("900633651-1", "clicked answer"));
-            }}
-                type="button">
-                    Alias
-          </button>
-          {"state: " + this.state.selected_answer}
-          {"props: " + this.props.question.selected_answer}
-
+    if (this.props.question) {
+    const options : {[option: string] : string} = this.props.question.options;
+    return (
+        <div className="flex-l items-start justify-between w-100 bt bw3 b--light-purple pt3">
+          <div>
+            <fieldset className="mb4">
+              <legend className="fw7">{this.props.question.question}</legend>
+              <div className="radio">
+                {Object.keys(options).map(
+                  (option_str :string)=>{
+                    const id : string = this.props.ambiguous_charge_id + options[option_str];
+                  return (
+                    <div key={id}>
+                    <input
+                      type="radio"
+                      name={this.props.ambiguous_charge_id}
+                      id={id}
+                      value={options[option_str]}
+                      onChange={this.handleRadioChange}
+                      />
+                    <label htmlFor={id}>{option_str}</label>
+                    </div>
+                    )
+                })}
+              </div>
+            </fieldset>
+          </div>
         </div>
-      ): null
-      );
+      )
+    } else {
+      return (<div></div>);
+    }
   }
 }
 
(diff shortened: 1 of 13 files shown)


wittejm:
The amount of manual type checking that I'm doing here is a bit over the top and I think some of the other frontend code tends to just use `any` a lot more.

wittejm:
oh right and will fix the vertically stacked answers thing...

hmarcks:
I have to do some updates to the radio css to get them to be inline. Feel free to pass that off to me if you'd like.

wittejm:
gotcha. That sounds good @hmarcks , I had thought it was a problem I introduced with an extra `div`, but that seems to be not the case.

src/frontend/src/components/RecordSearch/Record/Question.tsx
```diff
@@ -19,24 +19,26 @@ class Question extends React.Component<Props> {
...
+                  <div className="dib" key={id}>
+                  <input type="radio" name={this.props.ambiguous_charge_id} id={id} value={options[option_str]} onChange={this.handleRadioChange}/>
+                  <label htmlFor={id}>{option_str}</label>
+                  </div>
+                )
+              })}
+            </div>
+            <div className="radio-spinner absolute" role="status">
+              <span className="spinner spinner--sm mr1"></span>
+              <span className="f6 fw5">Updating&#8230;</span>
```
hmarcks, line 39: @wittejm can you try only showing this when loading?

wittejm:
@KentShikama it appears in the new dev setup the record isn't getting saved in-session. Are you seeing this behavior?
` ERROR:root:code 428 Must hit the search endpoint with question generating records first.`

src/frontend/src/components/RecordSearch/Record/Question.tsx
wittejm: can do

KentShikama:
Unfortunately I am not. The endpoint has changed to `http://localhost:5000` if that makes a difference.

wittejm:
The loading spinner now only shows while awaiting a disambiguate response and this PR should be ready to go. @KentShikama @hmarcks, want to take another look?

hmarcks:
Nice, looks good to me!

src/frontend/src/redux/search/actions.ts
```diff
@@ -1,15 +1,38 @@
...
+      total_balance_due: receivedRecord.total_balance_due,
+      cases: receivedRecord.cases,
+      errors: receivedRecord.errors,
+      summary: receivedRecord.summary,
+    };
+    dispatch({
+      type: SEARCH_RECORD,
+      record: record,
+      questions: receivedRecord.questions
+
```
KentShikama, line 28: space

src/frontend/src/redux/search/actions.ts
```diff
@@ -1,15 +1,38 @@
...
+  if (validateSearchResponseData(data)) {
+    const receivedRecord = data.record;
+    const record: RecordData = {
+      total_balance_due: receivedRecord.total_balance_due,
+      cases: receivedRecord.cases,
+      errors: receivedRecord.errors,
+      summary: receivedRecord.summary,
+    };
+    dispatch({
+      type: SEARCH_RECORD,
```
KentShikama, line 25: Probably change the name to something like `DISPLAY_RECORD`

src/frontend/src/redux/search/reducer.ts
```diff
@@ -20,14 +22,26 @@ export function searchReducer(
...
+        questions: action.questions,
+        loading: false
+      };
     case SEARCH_RECORD_LOADING:
       // When an API call is made for a record, loading state is toggled to true.
       // while loading state is true, a spinner is rendered on the screen. If loading
       // state is false, and no results were fetched, no "search results found" will be displayed.
-      return { ...state, record: {}, loading: true };
+      // return { ...state, record: {}, questions: {}, loading: true };
+      return {...state, loading: true};
```
KentShikama, line 36: And maybe just `RECORD_LOADING` here.

src/frontend/src/components/RecordSearch/Record/Question.tsx
```diff
@@ -0,0 +1,79 @@
...
+                <span className="f6 fw5">Updating&#8230;</span>
+              </div> :
+                null
+              )
+            }
+          </fieldset>
+        </div>
+      )
+    } else {
+      return (<div></div>);
```
KentShikama, line 50: What happens if you return nothing?

wittejm: I didn't love this, but `render()` needs to return a react element. The alternative I thought of was to not render a Question component at all if the charge has none, but that requires fetching from the redux store inside Charge, which I didn't love more.

wittejm: oh, cute. `null` is not ok, but `<></>` is, so I'll do that.

PR #1019, opened 2020-04-09
Remove Travis CI and other unused files

KentShikama:
@kenichi I had to remove the .travis.yml file so we don't get failing status checks but then I realized you had already started working on this. Hence, I've stolen your commit here.


kenichi:
👍

PR #1034, opened 2020-04-13
Fix unrecognized charges when length of statute is 4

KentShikama:


wittejm:
this needs some justification, and also unit tests probably

PR #979, opened 2020-03-30
added static method to record creator to sort the record cases by date

jjames1011:
Implements a solution for issue #933. Lists the cases in reverse chronological order by case date so that most recent cases are displayed first.

PR #1038, opened 2020-04-13
add sorting by date to RecordCreator

wittejm:
replaces #979 to account for new disambiguate logic, placing the `sort` call in `analyze_ambiguous_record` where it gets called by both the `search` and `disambiguate` endpoints.   

co-author @jjames1011

### DIFF
--- a/src/backend/expungeservice/record_creator.py
+++ b/src/backend/expungeservice/record_creator.py
@@ -67,5 +67,5 @@ class RecordCreator:
 
     @staticmethod
     def sort_record_by_case_date(record):
-        sortedCases = sorted(record.cases, key=lambda case: case.date, reverse=True)
-        return replace(record, cases=tuple(sortedCases))
+        sorted_cases = sorted(record.cases, key=lambda case: case.date, reverse=True)
+        return replace(record, cases=tuple(sorted_cases))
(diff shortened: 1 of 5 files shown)


wittejm:
github CI build appears to be failing on a server timeout error. https://www.githubstatus.com/ is currently calling the service "Degraded"

KentShikama:
You can click re-run jobs next time (but be careful on mindlessly clicking as it could be a rare generative test error)

Screen Shot 2020-04-13 at 7 08 29 PM

src/backend/expungeservice/record_creator.py
```diff
@@ -62,4 +62,10 @@ def analyze_ambiguous_record(ambiguous_record: AmbiguousRecord):
             charge_id_to_time_eligibilities.append(charge_id_to_time_eligibility)
             ambiguous_record_with_errors.append(record_with_errors)
         record = RecordMerger.merge(ambiguous_record_with_errors, charge_id_to_time_eligibilities)
-        return record
+        sorted_record = RecordCreator.sort_record_by_case_date(record)
+        return sorted_record
+
+    @staticmethod
+    def sort_record_by_case_date(record):
+        sortedCases = sorted(record.cases, key=lambda case: case.date, reverse=True)
```
KentShikama, line 70:
```suggestion
        sorted_cases = sorted(record.cases, key=lambda case: case.date, reverse=True)
```

wittejm: oh boo. yep

PR #1039, opened 2020-04-14
Merge contempt into civil offense

wittejm:
This keeps the test cases for different dispositions on civil offenses, so that it can assert the correct eligibility rules for each.


src/backend/expungeservice/charge_classifier.py
```diff
@@ -236,6 +229,8 @@ def _civil_offense(statute, name):
             return AmbiguousChargeTypeWithQuestion([CivilOffense])
         elif "fugitive" in name:
             return AmbiguousChargeTypeWithQuestion([CivilOffense])
+        elif "contempt" in name:
```
KentShikama, line 232: Have you double checked this is a safe abbreviation compared to "contempt of court"?

wittejm: Yea, but also every contempt of court case I found says "contempt of court", so this may as well be more specific

Issue #1042, opened 2020-04-16
Add Manual


hmarcks:
We have a manual to document how to use RecordSponge and how to file for expungement in a [Google Doc](https://docs.google.com/document/d/1Su078uVl_tg0TuvyAUGr8n2IBQolhWl0H3rN1LUFRNI/edit). We want to take the information in the Google Doc and put it on our website & app. Unauthenticated users will also have access to it.

### This task is done when:
 - A page is created on recordsponge.com with the Manual content that can be accessed when you are logged in or not

![Screen_Shot_2020-04-15_at_8_14_00_PM](https://user-images.githubusercontent.com/6887982/79403871-c1e24280-7f55-11ea-81d4-c3889eea4b4a.png)
----
![Screen_Shot_2020-04-15_at_8_11_09_PM](https://user-images.githubusercontent.com/6887982/79403875-c4449c80-7f55-11ea-8701-99c35f79f9f1.png)

hmarcks:
I'm currently making some edits to the content and creating the HTML. See progress here - feedback welcome: https://codepen.io/hmarcks/full/oNjgmZV

I'll need help with actually creating the view in React.

wittejm:
completed with #1047

PR #1044, opened 2020-04-16
[WIP] Add manual page

wittejm:
Empty for now. Start of #1042 

The page appears at `/manual` and is linked in the nav bar for logged-in users.

There's also a link added to the Landing page, but it's not styled, ~~and it doesn't work!~~ Edit: fixed.

### DIFF
--- a/src/frontend/src/components/AppRouter/index.tsx
+++ b/src/frontend/src/components/AppRouter/index.tsx
@@ -14,6 +14,7 @@ import AuthenticatedRoute from '../AuthenticatedRoute';
 import PublicRoute from '../PublicRoute';
 import InterimPage from '../InterimPage';
 import Landing from '../Landing';
+import Manual from '../Manual';
 
 class AppRouter extends React.Component {
   public redirect = () => <Redirect to="/" />;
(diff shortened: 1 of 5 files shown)


wittejm:
@hmarcks you can push to this branch.

The header just conditionally renders everywhere if the user is logged in. What do you think it should look like if the user isn't logged in? Should the user be expected to hit the back button to navigate elsewhere?

hmarcks:
@wittejm Can you just show the same <nav> when on /manual and not logged in?

Logged in version looks good, thank you

wittejm:
yes, but I think it should exclude the Search and Logout links but include the Login button as well as Manual. While we're at it, should we have a Home link for the landing page?

hmarcks:
Yeah, it should only have logo, Manual, and Log In. I think I'll just add a "back" link in the header of the Manual and make the logo a link to home.

wittejm:
Ok, cool. And, should the login page look the way it does now, or should we add the manual/ other nav content there too?

hmarcks:
hmm, good question, I think it's fine how it is

wittejm:
@hmarcks I decided to make a feature branch for this on the main repo, so that we don't accidentally disrupt each other's work by pushing directly to this branch. closing this PR and preplacing it with the new [WIP] PR

PR #1050, opened 2020-04-18
Remove num charges from search results table 

### LINKED ISSUE #990: Remove number of charges column from DB

wittejm:
Data table for a search event no longer contains any results info.
- Updates SQL create scripts 
- Updates test cases to not check for result info.
- We no longer have any notion of "stats" in the project.
- Changes to db operations are backwards-compatible with the existing database, so we don't need to migrate the DB in sync with this deployment. We could even delete the relevant columns manually.

closes #990


src/backend/expungeservice/database/db_util.py
```diff
@@ -12,3 +16,24 @@ def wrapper(database, *args, **kwargs):
...
+    search_param_string = user_id + json.dumps(request_data["aliases"], sort_keys=True)
+    hashed_search_params = hashlib.sha256(search_param_string.encode()).hexdigest()
+    _db_insert_search_event(g.database, user_id, hashed_search_params)
+
+
+@rollback_errors
+def _db_insert_search_event(database, user_id, hashed_search_params):
+    database.cursor.execute(
+        """
+        INSERT INTO  SEARCH_RESULTS(search_result_id, user_id, hashed_search_params)
```
KentShikama, line 32: Do you not need to run migrations to delete the columns first? The old db will expect 5 entries but only receive 3. How will it know that the final 2 should be just null?

wittejm: Nah it just works, and I tested on both versions of the table. Those two columns are nullable, so because we specify to insert to only the 3 columns, it only expects those values.

Running migrations in tandem would be good practice, but not strictly necessary.

PR #1057, opened 2020-04-21
Ensure inputs to RecordCreator.build_record are hashable

KentShikama:
This will be key in making our API stateless.


src/backend/expungeservice/record_creator.py
```diff
@@ -11,10 +11,18 @@
 from expungeservice.request import error
 
 
+@dataclass(frozen=True)
+class Alias:
```
wittejm, line 15: this should go in `models/`

PR #1060, opened 2020-04-22
Freeze node docker container to pull node version 13.13

KentShikama:
Our frontend tests started failing today: see https://github.com/codeforpdx/recordexpungPDX/pull/1057/checks?check_run_id=608399130. Since we don't specify the node version in docker, the node version was automatically bumped to node v14, which is not compatible with the latest node-sass version. This change freezes our node version at 13.13.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ jobs:
   frontend:
     runs-on: ubuntu-latest
     container:
-      image: node:alpine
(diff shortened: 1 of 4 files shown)


KentShikama:
I'm going to go ahead and merge this as otherwise all our other PRs will show red Xs. Feel free to follow up with any better ways to handle this.

kenichi:
perfect! these all look like the right changes and i'm glad we've pinned the node version now.

PR #1065, opened 2020-04-23
Reduced to Violation charge type

### LINKED ISSUE #1054: Expungement Logic: Fare Violation
I previously stated that offenses with statute less than six digits should be considered Civil Offenses and not eligible. There is an exception for Fare Violations. Fare Violations have the following attributes

Statute No: 2915 (Trimet Code, not ORS, inconsistent) 
Name: Non-Payment of Fare; Fare Violation (consistent)

Fare Violations follow Violation logic. So: 
Mark Fare Violation convictions as eligible under Violation. 
Mark Fare Violation dismissals as ineligible 

Suggested way to identify these cases is to see if the statute contains (the non-case-sensitive phrases) "fare violation" or "non-payment of fare"
--

--

--

### This task is done when:
 -
 -

### LINKED ISSUE #1058: Reduced Misdemeanors otherwise eligible should be marked eligible
This is kind of a fun one, as it involves an idiosyncratic disposition created by an idiosyncratic practice by the Multnomah DA's office. 

In community court (which adjudicates petty crimes), a person can plead guilty to enter a diversion program, at which point the case is immediately reduced from a misdemeanor to a violation. If a person successfully completes the program, their case is dismissed. 

Now, dismissals of Violations are typically ineligible for expungement. However, DAs have typically not objected to dismissed violations which were originally misdemeanors. The Charge will then display the phrase "(Reduced - DA Elected)". As such, these should be marked as type eligible. Therefore:

**If 1) a Violation 2) has the word "Reduced" (non-case-sensitive) in its name AND/OR has the phrase "treated as" (non-case-sensitive), and 3) its Disposition is Dismissed, then it is Type Eligible**

Multnomah County Community Court processes at least four or five cases with this type of decision per day, so this will make a potentially big difference. 

Screen Shot 2020-04-21 at 11 06 35 AM

### This task is done when:
 -
 -

wittejm:
closes #1058 
closes #1054

### DIFF
--- a/src/backend/expungeservice/charge_classifier.py
+++ b/src/backend/expungeservice/charge_classifier.py
@@ -17,6 +17,7 @@ from expungeservice.models.charge_types.marijuana_ineligible import MarijuanaIne
 from expungeservice.models.charge_types.marijuana_eligible import MarijuanaEligible, MarijuanaUnder21
 from expungeservice.models.charge_types.misdemeanor import Misdemeanor
 from expungeservice.models.charge_types.violation import Violation
+from expungeservice.models.charge_types.reduced_to_violation import ReducedToViolation
 from expungeservice.models.charge_types.parking_ticket import ParkingTicket
 from expungeservice.models.charge_types.person_felony import PersonFelonyClassB
 from expungeservice.models.charge_types.civil_offense import CivilOffense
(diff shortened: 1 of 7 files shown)


wittejm:
Addressed both charge types in one PR because of some shared changes to the classifier and adjacent tests.

wittejm:
Build checks are being "cancelled" and failing on their own, apparently?

Issue #1069, opened 2020-04-25
Improve error message when OECI is down for "maintenance"


KentShikama:
Currently the error message makes it seem like the user got their password wrong.

![Screen Shot 2020-04-25 at 5 43 00 AM](https://user-images.githubusercontent.com/6329898/80280240-bad6d500-86b7-11ea-83c6-5aefd0e4fdbc.png)

PR #1070, opened 2020-04-26
Change error message if OECI is in maintenance

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/1069

~~The message "The eCourt site is offline during the 4th weekend of each month between 6 PM PST on Friday until noon on Sunday. During this time, record search will not function." will only show during the time period in which it applies.~~

Screen Shot 2020-04-26 at 8 03 47 AM

### DIFF
--- a/src/backend/expungeservice/crawler/crawler.py
+++ b/src/backend/expungeservice/crawler/crawler.py
@@ -1,69 +1,73 @@
 from concurrent.futures.thread import ThreadPoolExecutor
 from dataclasses import replace
+from functools import partial
 from itertools import product
 from typing import List, Tuple, Optional
 
-import requests
 from datetime import datetime
 
 from expungeservice.charge_creator import ChargeCreator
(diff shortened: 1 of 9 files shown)


hmarcks:
Do you think we should persist the message about the downtime window (like we have now) since it's such a big block of time? Then repeat the error message like you have it if they get the response.

KentShikama:
Are you suggesting we implement something on the frontend like

```
if time.now() is 4th weekend of the month:
    display message on OECI login
```

hmarcks:
I'm just suggesting we leave the message there all the time, and show the error message you created if the form is submitted and we get the service down response:
Screen_Shot_2020-04-26_at_12_30_19_PM

wittejm:
I agree the message about downtime should show always, so the user isn't surprised about this issue.

KentShikama:
@hmarcks @wittejm Got it. I reverted that portion.

PR #1072, opened 2020-04-26
Remove account link from header

KentShikama:
Follow up from https://github.com/codeforpdx/recordexpungPDX/pull/1047

It looks like there was intent to remove the account link: https://github.com/codeforpdx/recordexpungPDX/pull/1047/files#diff-55e17045cee2e59c129fd7d44781c75bL29.


hmarcks:
Since it's not yet functioning, it looks like we need a new issue to implement it. This might include updating your password or deleting your account.

PR #1074, opened 2020-04-28
Edit record results backend support

wittejm:
Search endpoint accepts two new fields: `edits`, and `additions`

This commit implements editing and deleting cases, and editing charges, but not adding new cases, or adding/deleting new charges. Putting `additions` in the request does nothing for now.

The code has a little bit of slop like the pointless "service" fixture used in the test case that I didn't really look into.

This change is sufficient to support editing Dispositions on a record.

### DIFF
--- a/src/backend/tests/test_edit_results.py
+++ b/src/backend/tests/test_edit_results.py
@@ -1,0 +1,162 @@
+import copy
+import json
+from typing import List, Any, Callable, Tuple
+from datetime import datetime
+
+import pytest
+
+from expungeservice.crawler.crawler import Crawler
+from expungeservice.endpoints import search
+from expungeservice.models.ambiguous import AmbiguousCase
+from expungeservice.models.case import Case, OeciCase, CaseSummary
+from expungeservice.models.charge import OeciCharge
+from expungeservice.models.record import Question, Record
+from expungeservice.models.disposition import Disposition, DispositionStatus
+from tests.endpoints.endpoint_util import EndpointShared
+from tests.factories.crawler_factory import CrawlerFactory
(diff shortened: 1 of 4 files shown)


src/backend/tests/test_edit_results.py
```diff
@@ -0,0 +1,162 @@
...
+        []
+        )
+    assert len(record.cases) == 2
+    assert record.cases[0].summary.location == "earth"
+    assert record.cases[1].summary.location == "ocean"
+    assert record.cases[1].summary.balance_due_in_cents == 10000
+    assert record.cases[1].summary.date == datetime.date(datetime.strptime("1/1/1001", "%m/%d/%Y"))
+
+    location = "earth",
+    current_status = "CLOSED",
```
wittejm, line 135: TODO: remove these 2 lines

wittejm:
So. `additions` doesn't need to be its own parameter in the endpoint. After all, adding charges to existing cases will necessarily be under "edits". My thinking for separating these out is that the using request data format of `edits` as a dictionary

`edits: {case_number1: {action: "edit", ...data}, case_number2: {action: "edit", ...data}, }`

makes it easiest to apply all the edits by walking the cases in the search result and looking up each `case_number` to see if each has an entry in `edits`.  

So, if we included "add" actions in this `edits` data works too, it would involve: after using `edits` via lookup to edit/delete charges, next walk all of its entries like 

`for case_number, data in edits.items(): if data['action'] == 'add'` 

to add stuff.

That's actually fine. It makes the endpoint signature look the most intuitive.

wittejm:
request final review/approval @KentShikama

PR #1076, opened 2020-04-29
Add list of charges with disposition originally unknown to API

KentShikama:
This allows the frontend to display a question for unknown dispositions even after the user submits an answer in case they want to change their answer. Note the unknown dispositions are fetched out before the user edits are applied to the search results.


KentShikama:
@wittejm Yeah I'll fix that in a later PR.

Issue #687, opened 2020-01-04
Feature: User can interactively assign dispositions that were missing or unrecognized


wittejm:
This is a large-ish scale feature. 

When there are record errors due to missing/unrecognized dispositions, the time analysis for the rest of the record is potentially inaccurate and the user needs to be warned about this.

These dispositions might be known to the user even if they are missing in OECI, or they might just be a typo in OECI that is still human-readable. A fairly sophisticated solution is to let the user set the disposition of each charge in the results page and rerun the analysis.

@hmarcks has created a mockup for the feature:
 
![override](https://user-images.githubusercontent.com/6887982/71539335-89641d80-2900-11ea-99a6-44ae640e57fa.png)

### This task is done when:
 - The search results page includes the new interactive elements on charges with missing/unknown dispos.
 - A new endpoint accepts the record object that has been updated with custom dispositions, and returns the record with the updated expungement result.

hmarcks:
Updated mockup:

![missing-disposition-input](https://user-images.githubusercontent.com/6887982/80050144-76421280-84da-11ea-8e5c-6e337220dbf8.png)

hmarcks:
@wittejm here's some HTML:

[dispositionform-working-file.html.zip](https://github.com/codeforpdx/recordexpungPDX/files/4536862/dispositionform-working-file.html.zip)

wittejm:
all done!

PR #1088, opened 2020-05-02
Add support for probation revoked in disposition editing

KentShikama:
Related to #687


wittejm:
Need to translate "revoked" into a Convicted ruling:

unrecognized_provoked_with_date

PR #1089, opened 2020-05-04
Answering the unknown Disposition question is more Reactive

wittejm:
* The "Open" selection and "Missing" Disposition in charge info are both renamed to "Unknown"
 * Clicking Dismissed or Unknown triggers immediate update because no more info is needed. The Submit button doesn't appear when these are selected. 
 * Error tags disappear when the user clicks away from Probation Revoked to Convicted. The behavior here *could* get more fine-grained if we wanted to, especially when clicking on and off the Convicted option.
 * The submit button disappears after the user does a correct submit, and reappears if the user changes a field
 * The "updating" spinner is used here while loading any changes.

### DIFF
--- a/src/frontend/src/components/RecordSearch/Record/DispositionQuestion.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/DispositionQuestion.tsx
@@ -153,11 +153,11 @@ class DispositionQuestion extends React.Component<Props, State> {
                     <label htmlFor={this.props.ambiguous_charge_id + "-unknown"}>Unknown</label>
                 </div>
             </div>
-            <div className={this.state && this.state.submitPending ? "" : "visually-hidden"}>
+            <div className={this.state && (this.state.status === "Convicted" || this.state.status === "revoked") ? "" : "visually-hidden"}>
               <label className="db fw6 mt3 mb1" htmlFor="n">Date Convicted <span className="f6 fw4">mm/dd/yyyy</span></label>
               <input value={this.state.conviction_date} onChange={this.handleDateFieldChange} className="w5 br2 b--black-20 pa3" id="n" type="text" name="conviction_date"/>
             </div>
-            <div className={this.state && this.state.submitPending && this.state.status === "revoked" ? "" : "visually-hidden"}>
+            <div className={this.state && this.state.status === "revoked" ? "" : "visually-hidden"}>
               <label className="db fw6 mt3 mb1" htmlFor="n">Date Probation Revoked <span className="f6 fw4">mm/dd/yyyy</span></label>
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/RecordSearch/Record/DispositionQuestion.tsx
```diff
@@ -16,28 +16,70 @@ interface State {
...
+  handleRevokedClick = () => {
+    this.setState<any>({
+      ["status"]: "revoked",
+      ["submitPending"] : true
+    });
+  };
+
+  handleUnknownClick = () => {
+    this.setState<any>({
+      ["status"]: "Unknown",
```
KentShikama, line 71: I'm used to

```
this.setState({"key": "value", ...})
```

but what is the difference compared to

```
this.setState({["key"]: "value", ...})
```

src/frontend/src/components/RecordSearch/Record/DispositionQuestion.tsx
```diff
@@ -46,19 +88,26 @@ class DispositionQuestion extends React.Component<Props, State> {
...
-            this.props.ambiguous_charge_id,
-            this.state.status,
-            this.state.conviction_date || "1/1/2020", // The date for a dismissal doesn't matter, but the backend expects something so here we are.
-            this.state.probation_revoked_date
-          )
-        )
+        this.setState<any>({
+          ["submitPending"] : false
+        });
+        this.dispatchAnswer();
```
KentShikama, line 94: Should this not be pending?

hmarcks:
Awesome! I think the date fields could also be removed after they have been answered unless we want to make it very clear that the user has manually updated this data (although this doesn't scale to the full edit feature)

Screen_Shot_2020-05-04_at_9_32_46_PM

hmarcks:
Scratch that, until we have the full edit feature it would be too confusing to edit the date (select different option then select again)

src/frontend/src/components/RecordSearch/Record/DispositionQuestion.tsx
wittejm: submitPending is true while waiting for the user to click Submit, so in this case the button becomes hidden immediately after the click and before dispatching the update to the backend.

src/frontend/src/components/RecordSearch/Record/DispositionQuestion.tsx
wittejm: Ah, I was copying the format from line 83, `[e.target]: e.value` , in which a variable string key needs to be in brackets. But a string literal key doesn't, so I've removed those. Thanks.

src/frontend/src/components/RecordSearch/Record/DispositionQuestion.tsx
KentShikama: Maybe call it buttonClickPending then.

PR #1091, opened 2020-05-05
Clear probation revoked

wittejm:
Bugfix: currently, if you submit a Probation Revoked disposition answer, then change that answer to Convicted, the Probation Revoked date persists in Redux and the time eligibility is incorrect.

Deleting a key in Redux is annoying:
https://medium.com/better-programming/deleting-an-item-in-a-nested-redux-state-3de0cb3943da

But overwriting it with "" is easy.

This fix required making the backend handle a date field when it is an empty string (it gets skipped).

### DIFF
--- a/src/backend/expungeservice/record_creator.py
+++ b/src/backend/expungeservice/record_creator.py
@@ -180,7 +180,8 @@ class RecordCreator:
                         else:
                             charge_edits["disposition"] = None
                     elif key in ("date", "probation_revoked"):
-                        charge_edits[key] = datetime.date(datetime.strptime(value, "%m/%d/%Y"))
+                        if value:
+                            charge_edits[key] = datetime.date(datetime.strptime(value, "%m/%d/%Y"))
                     else:
                         charge_edits[key] = value
                 edited_charges.append(replace(charge, **charge_edits))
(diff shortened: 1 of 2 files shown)


src/backend/expungeservice/record_creator.py
```diff
@@ -180,7 +180,8 @@ def _edit_charges(charges: Tuple[OeciCharge, ...], edits):
                         else:
                             charge_edits["disposition"] = None
                     elif key in ("date", "probation_revoked"):
-                        charge_edits[key] = datetime.date(datetime.strptime(value, "%m/%d/%Y"))
+                        if value is not "":
```
KentShikama, line 183:
```suggestion
                        if value:
```

wittejm: "" is falsey, so that doesn't work

wittejm: Wait

wittejm: Oh, right, I was dealing with falsiness issues elsewhere, but I did this for ... Self-documenting code!

wittejm: Oh, right, I was dealing with falsiness issues elsewhere, but I did this for ... Self-documenting code!

KentShikama: We know it's a string from the types though. So if it's false it must be the empty string. I just don't like how it makes me pause upon review.

wittejm: Yep yep very well

PR #1104, opened 2020-05-08
Charge summary panel format and add NMA to summary serializer

KentShikama:
- Add NMA to record summarizer in backend 
- Change one line print of charge to include arrested date and disposition

An example charge summary panel post-PR:

Screen Shot 2020-05-07 at 10 59 31 PM


wittejm:
Good improvement I think but the summary panel is looking a little crowded. 

summary_panel_with_more_info

I think ideally these added details would fit on one line. We could re-proportion the columns, but also move one or more of these sections to be stacked vertically. 

What do you think @michaelzhang43 @hmarcks ? I can also just do the re-proportioned columns change this morning so we can see directly and compare.

michaelzhang43:
What do yall think about taking out the Eligible Cases entirely? I think having the arrest date provides better heuristic info than a case number

hmarcks:
If this new data is important for the summary I can look at a new layout, but it seems redundant to me because this information is in the details below.

KentShikama:
Large screen view:

Screen Shot 2020-05-08 at 2 02 25 PM

Mobile view:

Screen Shot 2020-05-08 at 2 02 38 PM

michaelzhang43:
This looks good. 

Should we also include a 'Needs more analysis' section like in the pdfs?

KentShikama:
@michaelzhang43 Sounds good; I've created an issue for it https://github.com/codeforpdx/recordexpungPDX/issues/1105

michaelzhang43:
Awesome

michaelzhang43:
I like the idea of having all charges listed in one of the categories (Eligible, Future, Ineligible, NMA) along with identifying information, so that the summary panel can provide a complete accounting of all charges.  Of our options, for a person with a lengthy history, arrest date is probably the best reminder for which charges the panel is referring to.

PR #1107, opened 2020-05-11
check manu/del crimes for keywords "iii" and " iv"

### LINKED ISSUE #1096: Excluding Schedule 3 from Marijuana consideration
Drug Delivery/Manufacturing/Possession offenses for marijuana are subject to different rules than other drug offenses. Non-marijuana offenses follow "normal" rules while marijuana offenses have special rules. We currently can tell that a Delivery/Manufacturing/Possession does not involve marijuana if it says that the name of the charge contains (Schedule) "II" or "2". 

**Similarly, we should parse out chargenames that contain "III" or "3" or "IV" or "4" **

### This task is done when:
 -
 -

wittejm:
Checks for " iv" (with a leading space) because the string "iv" is in the word "delivery", making a false positive. 

And the other schedule keywords don't have associated tests, so... :shrug: lazy option.

closes #1096

### DIFF
--- a/src/backend/expungeservice/charge_classifier.py
+++ b/src/backend/expungeservice/charge_classifier.py
@@ -155,12 +155,12 @@ class ChargeClassifier:
     @staticmethod
     def _manufacture_delivery(name, level, statute):
         if any([manu_del_keyword in name for manu_del_keyword in ["delivery", "manu/del", "manufactur"]]):
-            if any([schedule_2_keyword in name for schedule_2_keyword in ["2", "heroin", "cocaine", "meth"]]):
+            if any([schedule_2_keyword in name for schedule_2_keyword in ["2", "ii", "heroin", "cocaine", "meth"]]):
                 if level == "Felony Unclassified":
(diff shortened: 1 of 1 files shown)


src/backend/expungeservice/charge_classifier.py
```diff
@@ -151,14 +151,15 @@ def _marijuana_eligible(section, name, birth_year, disposition):
...
-                "2", "heroin", "cocaine", "meth"]]):
+                "2", "ii", "heroin", "cocaine", "meth"]]):
                 if level == "Felony Unclassified":
                     question_string = "Was the charge for an A Felony or B Felony?"
                     options = ["A Felony", "B Felony"]
                     return AmbiguousChargeTypeWithQuestion(
                         [FelonyClassA, FelonyClassB], question_string, options
                     )
-            elif "3" in name or "4" in name:
+            elif any ([schedule_2_keyword in name for schedule_2_keyword in [
```
KentShikama, line 161:
```suggestion
            elif any([schedule_3_keyword in name for schedule_3_keyword in [
```

PR #1125, opened 2020-05-14
Use node:13.13.0 image instead of alpine

KentShikama:


kenichi:
just curious about this one - I only used alpine images for size, did you run into issues with musl libc?

KentShikama:
@kenichi Ah no this was me just experimenting. The issue was that I was trying to do volume mapping in a docker container (aka a docker container in a docker container), which apparently doesn't work. In the end, I removed the container from the CI process.

PR #1155, opened 2020-05-18
Add links to scrapped case detail pages from OECI

KentShikama:
Unfortunately, navigating directly to the OECI link even while logged in does not work. Since we already fetch the case detail page while searching, we can just expose that under our own URL.

### DIFF
--- a/src/backend/expungeservice/endpoints/case_detail_page.py
+++ b/src/backend/expungeservice/endpoints/case_detail_page.py
@@ -1,0 +1,30 @@
+from os import path
+from pathlib import Path
+
+from flask.views import MethodView
+from flask_login import login_required
+
+from expungeservice.crawler.crawler import Crawler
+
+
+class CaseDetailPage(MethodView):
+    @login_required
+    def get(self, id):
+        url = f"https://publicaccess.courts.oregon.gov/PublicAccessLogin/CaseDetail.aspx?CaseID={id}"
+        html = Crawler.fetch_link(url)
+        if html:
+            return html.text
+        else:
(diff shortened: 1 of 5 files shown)


src/backend/expungeservice/endpoints/case_detail_page.py
```diff
@@ -0,0 +1,30 @@
...
+
+from flask.views import MethodView
+from flask_login import login_required
+
+from expungeservice.crawler.crawler import Crawler
+
+
+class CaseDetailPage(MethodView):
+    @login_required
+    def get(self, id):
```
wittejm, line 12: Codacy recommends renaming this which I think is the only suggestion that I support here. Did not know `id` was a built-in!

KentShikama: See https://flask.palletsprojects.com/en/1.1.x/views/#method-based-dispatching. 

The following `id` has nothing to do with the `id` above:
```
Help on built-in function id in module builtins:

id(obj, /)
    Return the identity of an object.
    
    This is guaranteed to be unique among simultaneously existing objects.
    (CPython uses the object's memory address.)
```

KentShikama: Frameworks like Django (and I'm assuming Flask) use [metaclasses](https://stackoverflow.com/questions/100003/what-are-metaclasses-in-python) (or at least class decorators) that transform classes themselves, creating syntax/patterns that can be surprising if you are not aware of the existence of metaclasses.

Issue #1079, opened 2020-04-30
Feature: Create & Edit Cases & Charges


hmarcks:
Placeholder for the feature to add and edit cases and charges.

### This task is done when:
 - Many things

hmarcks:
Here's a first draft of the UI to continue our conversation:

![add-case-and-charge](https://user-images.githubusercontent.com/6887982/80750469-c2e6a880-8aed-11ea-9c9e-6d264b2baf74.png)

wittejm:
Sweet!

hmarcks:
Some more thoughts on how to handle unknown vs unrecognized dispositions when we have this full edit functionality, and how to handle the other clarification questions:

![unknown-vs-unrecognized](https://user-images.githubusercontent.com/6887982/81139136-4d823a00-8f2a-11ea-8cc9-e66ca1da18bc.png)

hmarcks:
UI for undo edits and removing cases / charges. This should finish up the UI design for adding & editing cases & charges.

<img src="https://user-images.githubusercontent.com/6887982/81746062-77c48200-946b-11ea-9a65-d5f113b74dc3.png" height="400">

wittejm:
It just occurred to me that changing the case number is a weird mechanic. We identify the data objects by their case numbers, so when we apply edits we're reconciling a custom case number with a different OECI case number. We would need to retain the real case number and spoof it with the visible case number. Then the user could see the true case number only by undoing their edits to the case. I'm not sure there's actual value to this. But if someone disagrees and thinks some version of this is useful (@michaelzhang43 ?) it's doable. 

But being able to change case numbers of new cases is nice: Making the case number optional means one will get generated like `CASE001` or such. But if you want to rename your custom cases, then that's useful. 

Maybe, if it's an OECI case, display the case number but not in an editable field, or gray out the field?

michaelzhang43:
That sounds good. I agree that we should only allow changes to case number for custom cases. Are there any engineering reasons not to?

wittejm:
Only allowing changes to the case number for custom cases would be easy.

michaelzhang43:
Great, let's go with that then

hmarcks:
Sounds good, I updated the mockup. @wittejm I'll have some HTML for you soon.

KentShikama:
I'm thinking we should just auto-generate the case number for the user if it is a user-generated case.

hmarcks:
[...]

PR #1156, opened 2020-05-19
[WIP] Begin frontend functionality for Add/Edit Case

wittejm:
So far includes responsive "add case" and  "edit case" buttons to show the edit panel, and panel includes responsive buttons and form validation.

Redux/axios not touched yet.

### DIFF
--- a/src/backend/expungeservice/record_creator.py
+++ b/src/backend/expungeservice/record_creator.py
@@ -12,7 +12,7 @@ from expungeservice.crawler.crawler import Crawler, InvalidOECIUsernamePassword,
 from expungeservice.expunger import ErrorChecker, Expunger
 from expungeservice.generator import get_charge_classes
 from expungeservice.models.ambiguous import AmbiguousCharge, AmbiguousCase, AmbiguousRecord
-from expungeservice.models.case import Case, OeciCase, CaseSummary, CaseCreator
+from expungeservice.models.case import Case, OeciCase, CaseSummary, CaseCreator, EditStatus
 from expungeservice.models.charge import OeciCharge, Charge
 from expungeservice.models.charge_types.unclassified_charge import UnclassifiedCharge
 from expungeservice.record_merger import RecordMerger
@@ -149,9 +149,10 @@ class RecordCreator:
             if case_number in edits.keys():
                 if edits[case_number]["action"] == "edit":
                     edited_case, new_charges = RecordCreator._edit_case(case, edits[case_number])
-                    edited_cases.append(edited_case)
+                    edited_cases.append(replace(edited_case, summary=replace(edited_case.summary, edit_status=EditStatus.EDITED)))
                     new_charges_acc += new_charges
-                # else: if the action name for this case_number isn't "edit", assume it is "delete" and skip it
+                elif edits[case_number]["action"] == "deleted":
+                    edited_cases.append(replace(case, summary=replace(edited_case.summary, edit_status=EditStatus.DELETED)))
             else:
                 edited_cases.append(case)
         for (case_number, case_data) in edits.items():
@@ -167,7 +168,8 @@ class RecordCreator:
                         violation_type="",
                         current_status="",
                         case_detail_link="",
-                        balance_due_in_cents=0
+                        balance_due_in_cents=0,
+                        edit_status=EditStatus.ADDED
                     ),
                     charges=(),
                 )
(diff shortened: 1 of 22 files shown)


wittejm:
The current commit has a responsive panel and should have (nearly?) all the behavior prior to processing an actual edit.

A few differences from the mockup are:
- The edit-case and add-charge buttons are hidden while the case is being edited.
- The Case Number now only appears in the Case header, since we're not making it editable.
- The buttons on the panel are different for Add vs Edit case: Create/Cancel vs Update/Remove/Cancel
- Form validation is "live" -- it's impossible to type values into Balance that aren't of "###.##" format or Birth Year that aren't "####' format. We can eliminate a few validation error banners this way.

@hmarcks thoughts on these differences?

add_case
editing_panel

hmarcks:
Looking good!

> The edit-case and add-charge buttons are hidden while the case is being edited.
Can you explain this? I think we always want to show the buttons.

> The Case Number now only appears in the Case header, since we're not making it editable.
Sounds good

> The buttons on the panel are different for Add vs Edit case: Create/Cancel vs Update/Remove/Cancel
Oh? That's the same in the mockups

> Form validation is "live" -- it's impossible to type values into Balance that aren't of "###.##" format or Birth Year that aren't "####' format. We can eliminate a few validation error banners this way.
Sounds good, only concern with no validation message for Birth Year is that it may be confusing if you don't enter the correct year format (2 numbers) then aren't sure why it's not submitting.

wittejm:
> Can you explain this? I think we always want to show the buttons.

The example header pre-click and post-click:
pre-edit
clicked-edit

If the edit button stays there, what should it do? I think it should either become inactive or disappear. The expected behavior on click isn't obvious as to whether it causes an update or cancel.

If the +charge button stays, that lets the user open the add/edit-case panel and the add-charge panel at the same time. I could go either way on this but lean toward allowing only one at a time, so that the user must create an empty case and them populate it with charges. If the user can edit a case, and then add charges to it simultaneously, they might get confused about the expected workflow: in that "updating a case" might mean either 1) "update case info and all contained charges also, if there are pending charge edits" or that it would mean 2) "just update case info, and leave the charge changes pending."
 

> Oh? That's the same in the mockups

Ah yep. I was only looking at the final mockup : https://github.com/codeforpdx/recordexpungPDX/issues/1079#issuecomment-627601420

> Sounds good, only concern with no validation message for Birth Year is that it may be confusing if you don't enter the correct year format (2 numbers) then aren't sure why it's not submitting.

Good point. I can keep that validation message too.

hmarcks:
Oh got it, sorry I was thinking of the submit/remove/cancel buttons. I agree we should limit to one action at a time so hiding the buttons should work well. Thanks

src/backend/expungeservice/record_creator.py
```diff
@@ -2,7 +2,7 @@
 from functools import lru_cache
 from itertools import product, groupby
 from typing import List, Dict, Tuple, Any, Callable, Type
-from datetime import datetime
+from datetime import date, datetime
```
wittejm, line 5: will undo

wittejm:
Latest commit covers Add and Update case. Still todo are Undo changes and Remove.

src/frontend/src/redux/search/reducer.ts
```diff
@@ -40,14 +43,24 @@ export function searchReducer(
       }
       return {...state, questions: questions, loading: true};
     case ANSWER_DISPOSITION:
-      const edits = JSON.parse(JSON.stringify(state.edits));
+      edits = JSON.parse(JSON.stringify(state.edits));
```
KentShikama, line 46: Why remove the const here? `edits` can just be a different variable with the same name across the two cases.

src/backend/expungeservice/record_creator.py
```diff
@@ -149,11 +149,33 @@ def _edit_search_results(search_result_cases: List[OeciCase], edits) -> Tuple[Li
             if case_number in edits.keys():
                 if edits[case_number]["action"] == "edit":
                     edited_case, new_charges = RecordCreator._edit_case(case, edits[case_number])
-                    edited_cases.append(edited_case)
+                    edited_cases.append(replace(edited_case, summary=replace(edited_case.summary, edit_status=EditStatus.EDITED)))
```
KentShikama, line 152: This should be done in edit case:

`edited_summary = replace(case.summary, edit_status=EditStatus.EDITED, **case_summary_edits)`

src/frontend/src/redux/search/reducer.ts
wittejm: Ah, seems that case blocks don't have a new scope unless I make a new `{}` block. Adding that to allow two `const`s seems fine.

wittejm:
Pushed changes for Remove case and Undo.

There's a confirm message for hitting Remove on a manually added case. I wasn't sure whether to do the same for hitting U

wittejm:
n

wittejm:
Also todo: don't pre-populate balance and DOB fields when Adding new case.

KentShikama:
Should we close this?

PR #1161, opened 2020-05-21
Add support for unknown/missing disposition

KentShikama:
Relative to https://github.com/codeforpdx/recordexpungPDX/pull/1160

OECI charges will no longer have an optional Disposition. The previous value of None will become a DispositionStatus.UNKNOWN with a date of today as the disposition is unknown/missing/deferred as of today. This interpretation fits in nicely with how some existing charges have dispositions of "conditional discharge" or "deferred", which both mean that the disposition is missing as of today and will be input in the future. In any case, the disposition date for a DispositionStatus of UNKNOWN is never used.

### DIFF
--- a/src/backend/expungeservice/charge_classifier.py
+++ b/src/backend/expungeservice/charge_classifier.py
@@ -44,7 +44,7 @@ class ChargeClassifier:
     level: str
     section: str
     birth_year: Optional[int]
-    disposition: Optional[Disposition]
+    disposition: Disposition
 
     def classify(self) -> AmbiguousChargeTypeWithQuestion:
         def classification_found(c):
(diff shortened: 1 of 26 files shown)


src/backend/expungeservice/charge_classifier.py
```diff
@@ -156,7 +156,7 @@ def _marijuana_violation(name, level):
     @staticmethod
     def _marijuana_eligible(section, name, birth_year, disposition):
         if section == "475860" or "marij" in name or "mj" in name.split():
-            if birth_year and disposition:
+            if birth_year and disposition.status != DispositionStatus.UNKNOWN:
```
KentShikama, line 159: This is just a direct refactoring for the scope of this PR. But in a future PR, I do think that this should check for `disposition.convicted()` though.

Issue #1152, opened 2020-05-18
Add support for ambiguous dispositions


michaelzhang43:
In case M400034-D the charge is for a DUII with unknown disposition. 

The first disambiguation question asks for the disposition. 
But if you select dismissed, displays multiple analyses. 
Instead, it should ask the follow-up question of whether the charge was for Diversion

I haven't found a similar example for other cases with open disposition/follow-up analysis needed yet

### This task is done when:
 -
 -

PR #1175, opened 2020-05-31
Add support for ambiguous dispositions

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/1152


wittejm:
The new recursive rendering of Questions has changed the spacing around the loading spinner, which causes the page to jump around while updating answers (because multiple spinners appear at once)

PR #1187, opened 2020-06-03
Increase http timeout for uwsgi to 5 minutes

KentShikama:
Note this will not take effect until a docker image is pushed.


KentShikama:
TODO: Think about putting a threshold on the length of the ambiguous record (list).

Issue #1176, opened 2020-06-01
Fix hang that is caused by long record (with less than 300 cases)


KentShikama:

wittejm:
I'm not finding examples of this bug. Can you send me one?

KentShikama:
Potentially related: figure out what is causing the 60 second timeout and ensure backend processing is killed along with this timeout.

PR #1193, opened 2020-06-04
Return error to user if expunger will be run over 50 times

KentShikama:
Additionally, limit fetching of OECI record detail page to 30 seconds.

Resolves https://github.com/codeforpdx/recordexpungPDX/issues/1176


wittejm:
This error message sentence "Record with X combinations of ambiguities..." won't mean anything to most users. 

It could be: "The resulting record found was too large to analyze. The record most likely has open cases, and thus does not have any charges eligible to be expunged." 

Here are more thoughts that could warrant a new issue and PR:

The claim that the record may have has open cases is something we can verify without doing the complex analysis, in which case we wouldn't need to say "probably". I think this would be a minimal, if kind of inelegant, step to add here. 

If an open case was found in error (the search matched the wrong person), the user should ideally be aware of this. 

This prompts me to revisit the behavior that we have discussed: in the event there are open cases we don't do the analysis. In this approach we only decline to do the analysis if the record is too complex, and we can still show all of the found cases.

PR #1194, opened 2020-06-05
Remove authentication requirement from app

KentShikama:


wittejm:
The `NotAuthorized` component can also be deleted. Otherwise LGTM

I've never felt this giddy about seeing months of work thrown away :D

Issue #1173, opened 2020-05-31
Copy current fields when adding an alias


KentShikama:
### This task is done when:
 -
 -

PR #1205, opened 2020-06-09
This PR is in reference to: #1173

faranilucero:
This Pull Request:
- populates the 'first name', 'middle name', 'last name', and 'date of birth' fields when adding an Alias


wittejm:
It may be more intuitive to the user to have it populate a new alias with the one immediate prior, rather than always the first one. Otherwise looks fine.

faranilucero:
Thanks for the feedback, let me make some updates.

PR #1208, opened 2020-06-10
Remove notion of accounts from privacy policy

KentShikama:
Here is the proposed update.

----

<section class="mb5"><h2 class="f2 fw9 mb3" id="privacypolicy">Privacy Policy</h2><p class="mb3">Our privacy policy is simple and meant to be read by all of our users. Please email michael@qiu-qiulaw.com if anything is unclear.</p><h3 class="f4 fw7 mb2">What we collect and why</h3><p class="mb3">Our guiding principle is to collect only what we need, and we will not sell your data. Here’s what that means in practice:</p><h4 class="fw7 mb2">Search “pings” to track usage</h4><p class="mb3">The only information we collect are user search “pings,” which tells us when (and only when) a user has made a search. We collect this information to track usage rates. That’s it.</p><h4 class="fw7 mb2">Cookies</h4><p class="mb3">We use persistent first-party cookies to support necessary functions of the application. A cookie is a piece of text stored by your browser to help it remember your login information, site preferences, and more. You can adjust cookie retention settings in your own browser. To learn more about cookies, including how to see which cookies have been set and how to manage and delete them, please visit: <a href="https://www.allaboutcookies.org" class="bb hover-blue">allaboutcookies.org</a>.</p><h4 class="fw7 mb2">Voluntary correspondence</h4><p class="mb3">When you write RecordSponge with a question or to ask for help, we keep that correspondence, including the email address, so that we have a history of past correspondences to reference if you reach out in the future.</p><h3 class="f4 fw7 mb2">What we don't collect</h3><p class="mb3">We care about the privacy of your clients’ criminal records. Indeed, this project’s purpose is to make these records more private. Therefore, RecordSponge does not record or collect search information. If you are using this software for clients, we have no way of identifying who they are.</p><p class="mb3">We do not save your Oregon eCourt Case Information (OECI) login credentials. That’s why we must separately log in to OECI every time you use RecordSponge.</p><h3 class="f4 fw7 mb2">How we secure your data</h3><p class="mb3">All data is encrypted via SSL/TLS when transmitted from our servers to your browser.</p></section>

-----

Most deletions shouldn't require much explanation. The most aggressive edit is the decision to remove the following:

```
### When we access or share your information

Our default practice is to not access your information. The only times we’ll ever access or share your info are:
- To help you troubleshoot or squash a software bug, with your permission. If at any point we need to access your account to help you with a Support case, we will ask for your consent before proceeding.
- When required under applicable law. If the appropriate law enforcement authorities have the necessary warrant, criminal subpoena, or court order requiring we share data, we have to comply. Otherwise, we flat-out reject requests from local and federal law enforcement when they seek data. And unless we’re legally prevented from it, we’ll always inform you when such requests are made.
```

The search "pings" are technically information but they aren't identifiable enough that we can contact any particular person with it unless they explicitly tell us their IP address.

### DIFF
--- a/src/frontend/src/components/Manual/index.tsx
+++ b/src/frontend/src/components/Manual/index.tsx
@@ -795,19 +795,10 @@ class Landing extends React.Component {
               Our guiding principle is to collect only what we need, and we will not sell your data. Here’s what that means in practice:
             </p>
             <h4 className="fw7 mb2">
-              Identity & access
-            </h4>
-            <p className="mb3">
-              When you sign up for RecordSponge, we typically ask for identifying information such as your name, email address, and maybe a company name. We’ll never sell your personal info to third parties, and we won’t use your name or company in marketing statements without your permission either.
-            </p>
-            <p className="mb3">
-              Your username doubles as an email where we can contact you or someone in your organization. We encrypt passwords in a non-reversible way (a hash) so that we can compare for a correct password upon login. As with any online service, we highly recommend making your password different from other passwords you use.
-            </p>
-            <h4 className="fw7 mb2">
               Search “pings” to track usage
             </h4>
             <p className="mb3">
-              The only information we collect are user search “pings,” which tells us when (and only when) a user has made a search. We collect this information to track usage rates - if we see that you are searching more or less than usual, we may contact you to see if we can provide support. That’s it.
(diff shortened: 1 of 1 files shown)


michaelzhang43:
Looks good

PR #1216, opened 2020-06-11
Change charge eligibility label text

KentShikama:
- Use or instead of ⬥ in charge eligibility labels
- Change charge eligibility label prefix for open charges


wittejm:
screenies to show the new appearance:

newest_future_eligible_header
future_eligible_badge

hmarcks:
Regarding "If All Cases Are Closed, Eligible Now", isn't this incorrect since the closed case could involve convictions?

Do we want to do something more like, "[time] if open cases dismissed OR 10 years from conviction of open charges"?

wittejm:
It's shifted around a lot, sorry, meant to provide latest wording: it's now:
eligibility_future

wittejm:
@hmarcks  Is there an instance you're referring to that I'm missing?

Issue #1206, opened 2020-06-10
Remove the icon next to type eligibility


wittejm:
As suggested in this issue: #1190

This is an immediate improvement: Remove the check mark or "X" mark that appear next to charge type.

As discussed in issue #1190, this icon is sometimes confusing. Removing the symbol in all cases is the cleanest solution, and there's no loss of information to the user, I believe. More thoughts on why in the other issue.

Ark20:
pull request #1225

PR #1224, opened 2020-06-11
removed icon

Ark20:
I removed the icon next to type eligibility from issue #1206


wittejm:
Could you revert the space auto-formatting so this is just a few-line change? There might be different sources of auto-format either from your editor or an old git hook that shouldn't be there any more. But the one that @KentShikama just installed a few hours ago is hopefully working in your local repo.

wittejm:
closing in favor of #1225

Issue #1209, opened 2020-06-10
Add frontend validation to check for at least 1 character first name and 2 characters last name


KentShikama:
As a courtesy to OECI, we do not want to overload their systems even though it is permissible to enter double wildcards.

wittejm:
completed with #1234

PR #1234, opened 2020-06-12
Add wildcard validation to search form

sico:
Add validation for the name search fields such that:
* If a wildcard exists, it's at the end of the word
* For first name, the string must be at least 1 character long (not including the `*`)
* For last name, the string must be at least 3 characters long (not including the `*`)

Also adds a `test` dir inside `src` (where create-react-app wants it) in case we want to add more tests there. 

For https://github.com/codeforpdx/recordexpungPDX/issues/1209

Screenshot: 
image

### DIFF
--- a/src/frontend/src/components/RecordSearch/SearchPanel/index.tsx
+++ b/src/frontend/src/components/RecordSearch/SearchPanel/index.tsx
@@ -58,21 +58,18 @@ export default class SearchPanel extends React.Component<Props, State> {
           this.state.aliases[i].last_name.trim().length === 0
         ) {
           missingInputs = true;
-          break;
         }
         if (
           this.state.aliases[i].first_name.indexOf("*") > -1 &&
           !isValidWildcard(this.state.aliases[i].first_name, 2)
         ) {
           invalidFirstNameWildcard = true;
-          break;
         }
         if (
           this.state.aliases[i].last_name.indexOf("*") > -1 &&
           !isValidWildcard(this.state.aliases[i].last_name, 3)
         ) {
           invalidLastNameWildcard = true;
-          break;
         }
       }
       this.setState(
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/RecordSearch/SearchPanel/index.tsx
```diff
@@ -47,10 +55,20 @@ export default class SearchPanel extends React.Component<Props, State> {
...
+            !isValidWildcard(this.state.aliases[i].last_name, 'last'))
+        ) {
+            invalidWildcard = true;
+            break;
+        }
       }
       this.setState(
         {
-          missingInputs: missingInputs,
+          missingInputs,
```
sico, line 70: I wasn't sure if this styes was preferred - I can revert if desired here.

KentShikama:
For the last name it should be at least 2 characters instead of 3. We want to support searches for "Mc*" so we can do "Mc Donald" and "McDonald" without aliases.

KentShikama:
It also looks like you've missed committing a few files.

KentShikama:
Please run

```
pip install pre-commit
pre-commit install
```

This will trigger the `prettier` autoformatter.

src/frontend/src/components/RecordSearch/SearchPanel/validators.ts
```diff
@@ -0,0 +1,8 @@
+export function isValidWildcard(searchTerm: string, firstOrLast: 'first' | 'last'): boolean {
+  const requiredLength = firstOrLast === 'first' ? 2 : 3;
```
KentShikama, line 2: I would prefer just passing in the length.

src/frontend/src/components/RecordSearch/SearchPanel/index.tsx
```diff
@@ -161,6 +179,12 @@ export default class SearchPanel extends React.Component<Props, State> {
                   The date format must be MM/DD/YYYY.
                 </p>
               ) : null}
+
+              {this.state.invalidWildcard && (
+                <p id="name_msg" className="bg-washed-red mv4 pa3 br3 fw6">
+                  Wildcards must be at the end of the search term and contain 3 or more characters.
```
KentShikama, line 185: This error message should depend on first/last length.

wittejm:
Applied @KentShikama 's comments and rebased. Codacy's complaints appear to be incorrect, as far as I can tell.

Issue #1226, opened 2020-06-11
Add a Balance Due by Case section in summary panel


wittejm:
Related to #1191

swervinerv:
@wittejm should this breakdown only display cases that still have a balance or should it display all cases even if they have a $0 balance?

Basing the changes on this mockup that was called out in #1191 
<img width="1151" alt="Screen Shot 2020-06-14 at 9 12 04 AM" src="https://user-images.githubusercontent.com/6025685/84598515-333c5580-ae20-11ea-876d-c23fd1581abe.png">

wittejm:
It shouldn't show cases with zero balance.

wittejm:
completed with #1242

Issue #1191, opened 2020-06-04
Display $ owed in summary


michaelzhang43:
As per recent discussion, we should mark $ owed cases more clearly: 

1. put any $ owed in the summary panel next to each case where $ is owed. Something like:
Charge Name - Arrested m/d/y - $20 owed in case
I realize that for cases with multiple charges, multiple charges will get that tag, making it seem the person owes many times their actual debt. hence the "owed in case" - am open to other phrasings. 

2. Make a separate section for cases with $ owed, below the section for Eligible Now cases

### This task is done when:
 -
 -

wittejm:
Do we still want to indicate in the case detail block, the info "This becomes eligible after the case balance has been paid" ? Something other than the green eligible banner?

michaelzhang43:
Oh yes @wittejm . I had forgotten about that step. Yes, we should put cases with $ owing in a separate section below the Eligible Nows

michaelzhang43:
Let me actually edit the original query

hmarcks:
Here are some thoughts on how to show this:

![balance](https://user-images.githubusercontent.com/6887982/83806033-7ce5ad00-a676-11ea-8511-050d84cfc303.png)

KentShikama:
I realized that if there are multiple charges in a case, it is not clear which charge the balance due is attached to.

michaelzhang43:
Sorry, just seeing the updates now:

I personally like the combination of Summary Option 2 and Case Option 1. I think in both of these, the balance due is attached to the case itself, rather than any of the charges specifically. in Summary Option 2, the charges are implicated as being ineligible until paid off, but the dollar amounts aren't duplicated, creating the illusion of owing multiple times the actual amount

hmarcks:
> I realized that if there are multiple charges in a case, it is not clear which charge the balance due is attached to.

Does it matter what charge in the case it relates to, or do we just need to know the case has a balance.

After our last discussion I also like Summary Option 2 + Case Option 1 because the Summary categories get a bit complex, although it does reveal more detailed information in Summary Option 3.

wittejm:
The current open PR #1218 applies the dollar badge in the summary panel. The other two new features proposed here are separated into new issues #1226 and #1227 and can have independent PRs.

wittejm:
The issues there support @hmarcks and @michaelzhang43 's own endorsements for Summary option 2 and Case option 1.

wittejm:
completed in various recent PRs

PR #1242, opened 2020-06-14
Add Balance by Case Section

swervinerv:
Added a component for displaying balances by case and a shared component for displaying the balance list for both the by case and by county display.

### DIFF
--- a/src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
@@ -20,12 +20,12 @@ export default class CaseBalances extends React.Component<Props> {
 
     const listItems = this.props.cases
       .filter(hasBalanceDue)
-      .map((element: CaseData, i: number) => {
+      .map((element: CaseData) => {
         return (
-          <li className="mb2" key={i}>
+          <li className="mb2" key={element.case_number}>
             <a
               className="link underline hover-blue break-all"
-              href={`#${element.case_number}-1`}
+              href={`#${element.case_number}`}
             >
               {element.case_number}
             </a>{" "}
--- a/src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
@@ -1,0 +1,46 @@
+import React from "react";
+import BalanceList from "./BalanceList";
+import { CaseData } from "../types";
+import currencyFormat from "../../../../service/currency-format";
+
+interface Props {
+  cases?: CaseData[];
+}
+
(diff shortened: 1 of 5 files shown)


swervinerv:
@wittejm I was going to ask if we wanted to add the links so I'll get that added in.

For the `cases` array, I've used a library like [relelect](https://github.com/reduxjs/reselect) on other projects to avoid passing large amounts of data down through a component tree and handle things like the filtering being applied. I can create another branch with that approach quick to show what it would look like.

swervinerv:
@wittejm I have pushed changes for adding the link to the case number and not showing the count balance if the balance due is 0. I also made a change to no longer re-calculate the balance total for the by case balance and to instead use the `total_balance_due` coming back from the api.

wittejm:
Awesome. Kent or I will review it soon

src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    const listItems = this.props.cases
+      .filter(hasBalanceDue)
+      .map((element: CaseData, i: number) => {
+        return (
+          <li className="mb2" key={i}>
+            <a
+              className="link underline hover-blue break-all"
+              href={`#${element.case_number}-1`}
```
KentShikama, line 28: Is there a reason we aren't just linking to the case?

src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
```diff
@@ -0,0 +1,47 @@
...
+  render() {
+    if (!this.props.cases) {
+      return null;
+    }
+
+    const listItems = this.props.cases
+      .filter(hasBalanceDue)
+      .map((element: CaseData, i: number) => {
+        return (
+          <li className="mb2" key={i}>
```
KentShikama, line 25: I think we can use the `case_number` as the `key` here.

swervinerv: Good catch, updated to use the `case_number` as the key

src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
swervinerv: @wittejm's initial comment on the PR mentioned to use the in page link so that is what I went with but I just noticed this would actually be linking to the first charge so I will update that. If we instead wanted to link to the actual case details page I can look at making that change as well.

KentShikama:
Linking https://github.com/codeforpdx/recordexpungPDX/issues/1191 and https://github.com/codeforpdx/recordexpungPDX/issues/1226

PR #1245, opened 2020-06-14
[WIP] issue 1240 parse case for text after vs[.]? to produce defendant name

victorjz:
[WIP] parse case for text after vs[.]? to produce defendant name
new member is called versus under CaseSummary
tests adjusted but still one test failing: test_crawler/test_nonzero_balance_due_on_case
it was suggested I submit as WIP for guidance

### DIFF
--- a/src/backend/expungeservice/crawler/parsers/case_parser.py
+++ b/src/backend/expungeservice/crawler/parsers/case_parser.py
@@ -21,12 +21,14 @@ class CaseParserData:
     hashed_dispo_data: Dict[int, Dict[str, str]]
     balance_due: str
     probation_revoked: Optional[date]
+    versus: str
 
 
 class CaseParser:
     @staticmethod
     def feed(data) -> CaseParserData:
         soup = BeautifulSoup(data, "html.parser")
+        versus = CaseParser.__build_post_vs(soup)
         hashed_charge_data = CaseParser.__build_charge_table_data(soup)
         (
             hashed_dispo_data,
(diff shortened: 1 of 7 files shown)


src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -21,12 +21,14 @@ class CaseParserData:
     hashed_dispo_data: Dict[int, Dict[str, str]]
     balance_due: str
     probation_revoked: Optional[date]
+    versus: str
```
KentShikama, line 24:
```suggestion
    defendant_name: str
```

src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -37,7 +39,20 @@ def feed(data) -> CaseParserData:
             probation_revoked = date.fromdatetime(datetime.strptime(probation_revoked_date_string, "%m/%d/%Y"))
         else:
             probation_revoked = None  # type: ignore
-        return CaseParserData(hashed_charge_data, hashed_dispo_data, balance_due, probation_revoked)
+        return CaseParserData(hashed_charge_data, hashed_dispo_data, balance_due, probation_revoked, versus)
+
+    @staticmethod
+    def __build_post_vs(soup) -> str:
+        case_find = soup.find('b') # bold tag
```
KentShikama, line 46:
```suggestion
        case_find = soup.find('b') # first bold tag on page
```

src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -37,7 +39,20 @@ def feed(data) -> CaseParserData:
...
+        if case_find:
+            case_name = case_find.text
+            versus_pattern = re.compile("State of Oregon[\s]+vs[.]?[\s]+(.*)", re.IGNORECASE)
+            versus_match = versus_pattern.match(case_name)
+            if versus_match:
+                versus = versus_match.group(1)
+                return versus
```
KentShikama, line 53:
```suggestion
                return versus_match.group(1)
```

victorjz:
Adding to what Kent said, `info[0]` is already being loaded into the `name` field in method `CaseCreator`
So you could even drop the issue and add not change the project :D

> class CaseCreator:
    @staticmethod
    def create(
         versus, info, case_number, citation_number, date_location, type_status, case_detail_link, balance="0",
    ) -> CaseSummary:
        name = info[0]

KentShikama:
@victorjz I've just confirmed with @michaelzhang43 that we can use `info[0]`. I apologize about not being able to catch this earlier, but I will say that I don't think I could have caught this unless I attempted to work on it myself.

Issue #1227, opened 2020-06-11
Add Balance Due banner in case detail block


wittejm:
Related to #1191 
 
See mockup for Case Option 1 on that issue.

Ark20:
Where should the paying balances text link to on this? I've added the div & styling

wittejm:
I don't think a good target for that link exists yet. It will need to part of the intended changes to the manual. I'd say for now you can defer that part and just leave the link out of this banner.

PR #1247, opened 2020-06-15
Balance banner

Ark20:
Added Balance Due banner in case detail block. issue: #1227


hmarcks:
Hi @Ark20 , we use Tachyons for our css framework ([more info here](http://dev.huntermarcks.net/design-system/#css-framework)), so could you please use a class combo like `<p className="bg-washed-red fw6 br3 pv2 ph3 ma2">` and remove the css?

The styles you made look good, but this allows us to avoid new css and more easily keep styles consistent when many folks are working on the app. Thanks!

Ark20:
Thanks for the tachyon link and tips @hmarcks, I removed the additional css

hmarcks:
👍 thanks, I'll let @wittejm or @KentShikama approve the js

wittejm:
Code looks good for this, but looking at it again I need to revisit the design: 
@hmarcks @KentShikama @michaelzhang43 

Pulling in comments from the related issue #1191 
> Kent: I realized that if there are multiple charges in a case, it is not clear which charge the balance due is attached to.

> Hunter: Does it matter what charge in the case it relates to, or do we just need to know the case has a balance? 

Now that I'm looking a case with two charges, this is what we get:

balance_banner_two_charges

"Becomes" sounds like it only refers to one thing, but the charges are what become eligible, not the case. This especially should be clear when only some of the charges are eligible.

I think this could be improved by tweaking the language to "Charges become eligible when balance is paid"

Thoughts @hmarcks ?

@Ark20 this PR is coming from a branch on the main repo, but I saw you made your earlier PRs from your own clone. This one will be fine to merge as is, but for next time just remember to PR from your clone.

hmarcks:
Ah, good catch. As long as a case balance causes all charges in a case to be ineligible, then I like your language update.

I guess "Eligible charges become eligible when balance is paid" is more accurate, but awkward.

wittejm:
"Charges are ineligible until balance is paid"?

hmarcks:
I like it

Ark20:
Thanks @hmarcks @wittejm. I must have gotten things mixed up when I rebased but I'll make sure to make PRs from my clone going forward

Issue #1253, opened 2020-06-16
Do not show "Charges are ineligible until balance is paid" if charge is already ineligible


KentShikama:
![Screen Shot 2020-06-16 at 6 31 29 AM](https://user-images.githubusercontent.com/6329898/84780942-0e202200-af9b-11ea-85d7-5aceed71e3dc.png)

Ark20:
@KentShikama I'm not super clear on the backend but want to fix this one. How would I check the charge eligibility in the case file ?

KentShikama:
Since the banner is applied on the case, the logic more precisely is that if all charges on the case are ineligible, then don't show the banner. Everything comes as part of the record summary object from the backend. See the response object for the /api/search endpoint request at `Tools -> Web Developer -> Network`.

Ark20:
Thanks @KentShikama! With that logic I can still see the banner if say one charge is ineligible but others aren't. @hmarcks used the language "Eligible charges become eligible when balance is paid" can we make the banner say that or "Eligible charges are ineligible until balance is paid" while keeping this same logic 
<img width="832" alt="Screen Shot 2020-06-17 at 1 28 39 AM" src="https://user-images.githubusercontent.com/20876942/84875469-dd8fc500-b03a-11ea-9c06-7d1f8b3ef67d.png">

Ark20:
The updated logic ` let ineligibleCharges = 0; 
    let allIneligible = 0
    for (let i = 0; i < charges.length; i++){ 
      let typeEligibility = charges[i].expungement_result.type_eligibility.status
      if (typeEligibility == "Ineligible") {
        ineligibleCharges++
      }
    }
    if (ineligibleCharges == charges.length) {
        allIneligible=1
      }`

PR #1255, opened 2020-06-17
Banner fix

Ark20:
reworded banner and added logic to only display when there are eligible charges that are unpaid issue #1253

### DIFF
--- a/src/frontend/src/components/RecordSearch/Record/Case.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/Case.tsx
@@ -62,8 +62,8 @@ export default class Cases extends React.Component<Props> {
         </div>
         {
           balance_due > 0 && !allIneligible?
-        <div className="bg-washed-red fw6 br3 pv2 ph3 ma2">Eligible charges are ineligible until balance is paid 
-        </div>:""}
+            <div className="bg-washed-red fw6 br3 pv2 ph3 ma2">Eligible charges are ineligible until balance is paid
+        </div> : ""}
         <Charges charges={charges} />
       </div>
     );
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/RecordSearch/Record/Case.tsx
```diff
@@ -19,6 +19,17 @@ export default class Cases extends React.Component<Props> {
...
+    let allIneligible = 0
+    for (let i = 0; i < charges.length; i++){ 
+      let typeEligibility = charges[i].expungement_result.type_eligibility.status
+      if (typeEligibility == "Ineligible") {
+        ineligibleCharges++
+      }
+    }
+    if (ineligibleCharges == charges.length) {
+        allIneligible=1
+      }
```
KentShikama, line 32:
```suggestion
    let allIneligible = charges.every(charge => charge.expungement_result.type_eligibility.status === "Ineligible");
```

Note I have not tested this.

src/frontend/src/components/RecordSearch/Record/Case.tsx
```diff
@@ -19,6 +19,7 @@ export default class Cases extends React.Component<Props> {
       location,
       current_status,
     } = this.props.case;
+    let allIneligible = charges.every(charge => charge.expungement_result.type_eligibility.status === "Ineligible");
```
wittejm, line 22: this can be `const` instead of `let`. Looks good otherwise, so I'm approving and you can just merge after you make that change.

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/1191

PR #1261, opened 2020-06-19
[WIP] - Update CaseBalances component to use reselect library

swervinerv:
Added the reselect library for using a memoized selector and avoid passing the cases array through the component tree via props. This pull request is mainly meant for feedback and as an option to address the following comment from @wittejm on [PR #1242](https://github.com/codeforpdx/recordexpungPDX/pull/1242).

> I think we should avoid passing the whole cases list into the RecordSummary if possible, but that would require a backend change to get case balances into RecordSummaryData . Maybe that would end up being too big a PR, so I'm fine if we leave it with your current approach and follow up with a separate issue.

### DIFF
--- a/src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
@@ -1,47 +1,43 @@
 import React from "react";
+import { useSelector } from "react-redux";
 import BalanceList from "./BalanceList";
 import { CaseData } from "../types";
+import { AppState } from "../../../../redux/store";
+import { selectCasesWithBalanceDue } from "../../../../redux/search/selectors";
 import currencyFormat from "../../../../service/currency-format";
 
 interface Props {
-  cases?: CaseData[];
   totalBalance: number;
 }
 
-const hasBalanceDue = (element: CaseData) => {
-  return element.balance_due > 0;
+const CaseBalances: React.FunctionComponent<Props> = (props) => {
+  const cases = useSelector((state: AppState) => {
+    return selectCasesWithBalanceDue(state);
+  });
+
+  const listItems = cases.map((element: CaseData) => {
+    return (
+      <li className="mb2" key={element.case_number}>
+        <a
+          className="link underline hover-blue break-all"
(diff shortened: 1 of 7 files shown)


KentShikama:
Is it possible to separate this out into 2 commits: one for just updating package-lock.json with all the package version bumps and then updating package-lock.json to add reselect?

swervinerv:
@KentShikama yeah that should be possible. I will let you know when it is updated.

swervinerv:
@KentShikama this should now be two separate commits. Let me know if you have any feedback.

src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
```diff
@@ -1,47 +1,43 @@
...
+  const listItems = cases.map((element: CaseData, i: number) => {
     return (
-      <div className="mb4">
-        <BalanceList
-          listItems={listItems}
-          title="Balance Due by Case"
-          totalBalance={this.props.totalBalance}
-        />
-      </div>
+      <li className="mb2" key={i}>
```
KentShikama, line 20: I think this key should be `case_number`.

src/frontend/src/redux/search/selectors.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const selectCasesWithBalanceDue = createSelector(
+  [getRecordState],
+  (record) => {
+    if (record === undefined) {
+      return [];
+    }
+
+    const { cases } = record;
+    if (cases === undefined) {
```
KentShikama, line 19: Since we're doing style refactors, it would be nice to be able to use the syntax `record?.cases?.filter(hasBalanceDue)` in a future PR.

src/frontend/src/components/RecordSearch/Record/RecordSummary/CaseBalances.tsx
swervinerv: This has been updated

src/frontend/src/redux/search/selectors.ts
swervinerv: I went ahead and made this change now since it cleans up the implementation quite a bit.

PR #1263, opened 2020-06-19
Add and Edit Case/Charge backend

wittejm:
The `edits` data object supplied in the request must include the `case.summary.edit_status` field to specify whether the action is an `ADD`, `UPDATE`, or `DELETE`, and the `charge.edit_status` is also used to specify the edits made on charges. 

applying ADD to case will assume the contained charges data are all ADDs.
DELETE on a case will apply a DELETE status to all the charges in the case, and ignore other contents of edits data for that case.
UPDATE on a case can contain edits for the case data, as well as ADD, UPDATE, and DELETE edits on the charges on that case.

### DIFF
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -30,12 +30,12 @@ class Expunger:
         cases = analyzable_record.cases
         for charge in analyzable_record.charges:
             eligibility_dates: List[Tuple[date, str]] = []
-            other_charges = [
+            other_blocking_charges = [
                 c
                 for c in analyzable_record.charges
                 if c.charge_type.blocks_other_charges and c.id != charge.id and charge.edit_status != EditStatus.DELETE
             ]
-            dismissals, convictions = Expunger._categorize_charges(other_charges)
+            dismissals, convictions = Expunger._categorize_charges(other_blocking_charges)
             most_recent_blocking_dismissal = Expunger._most_recent_different_case_dismissal(charge, dismissals)
             most_recent_blocking_conviction = Expunger._most_recent_convictions(convictions)
             other_convictions_all_traffic = Expunger._is_other_convictions_all_traffic(convictions)
@@ -94,7 +94,7 @@ class Expunger:
                 )
 
             if charge.convicted() and isinstance(charge.charge_type, FelonyClassB):
-                if Expunger._calculate_has_subsequent_charge(charge, other_charges):
+                if Expunger._calculate_has_subsequent_charge(charge, other_blocking_charges):
                     eligibility_dates.append(
                         (
                             date.max(),
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -7,7 +7,7 @@ from dateutil.relativedelta import relativedelta
 from more_itertools import padnone, take
 
 from expungeservice.models.case import Case
-from expungeservice.models.charge import Charge
+from expungeservice.models.charge import Charge, EditStatus
 from expungeservice.models.charge_types.felony_class_b import FelonyClassB
 from expungeservice.models.charge_types.juvenile_charge import JuvenileCharge
 from expungeservice.models.charge_types.marijuana_eligible import MarijuanaUnder21, MarijuanaViolation
@@ -31,7 +31,9 @@ class Expunger:
         for charge in analyzable_record.charges:
             eligibility_dates: List[Tuple[date, str]] = []
             other_charges = [
-                c for c in analyzable_record.charges if c.charge_type.blocks_other_charges and c.id != charge.id
+                c
+                for c in analyzable_record.charges
+                if c.charge_type.blocks_other_charges and c.id != charge.id and charge.edit_status != EditStatus.DELETE
             ]
             dismissals, convictions = Expunger._categorize_charges(other_charges)
             most_recent_blocking_dismissal = Expunger._most_recent_different_case_dismissal(charge, dismissals)
--- a/src/backend/expungeservice/record_editor.py
+++ b/src/backend/expungeservice/record_editor.py
@@ -10,6 +10,14 @@ from expungeservice.models.case import OeciCase, CaseCreator
 from expungeservice.models.charge import Charge, OeciCharge, ChargeType
 from expungeservice.models.charge_types.unclassified_charge import UnclassifiedCharge
 from expungeservice.models.disposition import DispositionCreator
+from enum import Enum
+
+
+class EditStatus(str, Enum):
+    UNCHANGED = "UNCHANGED"
+    EDITED = "EDITED"
+    ADDED = "ADDED"
+    DELETED = "DELETED"
 
 
 class RecordEditor:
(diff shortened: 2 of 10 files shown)


src/backend/expungeservice/record_editor.py
```diff
@@ -6,74 +6,101 @@
...
+                edited_case, new_charges = RecordEditor._edit_case(case, edit)
+                edited_cases.append(edited_case)
+                new_charges_accumulator += new_charges
+            elif edit["summary"]["edit_status"] == EditStatus.UPDATE:
+                for case in search_result_cases:
+                    case_number = case.summary.case_number
+                    if case_number == edit_action_case_number:
+                        edited_case, new_charges = RecordEditor._edit_case(case, edit)
+                        edited_cases.append(edited_case)
+                        new_charges_accumulator += new_charges
```
KentShikama, line 36:
```suggestion
case = next(case for case in search_result_cases if case.summary.case_number == edit_action_case_number, None)
if case:
    edited_case, new_charges = RecordEditor._edit_case(case, edit)
    edited_cases.append(edited_case)
    new_charges_accumulator += new_charges
else:
    raise ValueError("This should never happen")
```

src/backend/expungeservice/record_editor.py
```diff
@@ -6,74 +6,101 @@
...
+            elif edit["summary"]["edit_status"] == EditStatus.DELETE:
+                for case in search_result_cases:
+                    case_number = case.summary.case_number
+                    if case.summary.case_number == edit_action_case_number:
+                        edited_case, new_charges = RecordEditor._edit_case(case, edit)
+                        edited_cases.append(edited_case)
+                        new_charges_accumulator += new_charges
```
KentShikama, line 43: Extract function

src/backend/tests/test_edit_results.py
```diff
@@ -108,20 +122,119 @@ def test_edit_some_fields_on_case():
...
+                "charges": {
+                    "X0001-1": {
+                        "edit_status": "UPDATE",
+                        "charge_type": "FelonyClassC",
+                        "date": "1/1/2001",
+                        "disposition": {"date": "2/1/2020", "ruling": "Convicted"},
+                    },
+                    "X0001-2": {"edit_status": "DELETE"},
+                    "X0001-3": {
+                        # "edit_status": "ADD",
```
KentShikama, line 220: This comment needs clarification

src/backend/expungeservice/record_editor.py
```diff
@@ -6,74 +6,101 @@
...
 from expungeservice.models.case import OeciCase, CaseCreator
-from expungeservice.models.charge import Charge, OeciCharge, ChargeType
 from expungeservice.models.charge_types.unclassified_charge import UnclassifiedCharge
 from expungeservice.models.disposition import DispositionCreator
 
 
 class RecordEditor:
     @staticmethod
     def edit_search_results(search_result_cases: List[OeciCase], edits) -> Tuple[List[OeciCase], List[Charge]]:
+        next_new_case_number = 1
```
KentShikama, line 18:
```suggestion
        next_case_number = 1
```

src/backend/expungeservice/expunger.py
```diff
@@ -31,7 +31,9 @@ def run(record: Record) -> Dict[str, TimeEligibility]:
         for charge in analyzable_record.charges:
             eligibility_dates: List[Tuple[date, str]] = []
             other_charges = [
-                c for c in analyzable_record.charges if c.charge_type.blocks_other_charges and c.id != charge.id
+                c
+                for c in analyzable_record.charges
+                if c.charge_type.blocks_other_charges and c.id != charge.id and charge.edit_status != EditStatus.DELETE
```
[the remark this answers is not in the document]
wittejm: Not sure if I totally follow, but I think what's going on is, in the line:
`other_convictions_all_traffic = Expunger._is_other_convictions_all_traffic(convictions)`
The variable `convictions` already excludes non-blocking convictions, which means it never contains TrafficViolation charges.
so the value of `other_convictions_all_traffic` is true if and only if `convictions` is empty.

So, that sounds like some more cleanup could happen here. It also has nothing to do with EditStatus.DELETE and still seems a bit confusing, so I would suggest that we address this in a different PR.

src/backend/expungeservice/record_editor.py
wittejm: This is awesome.

wittejm:
My new changes bypass your suggested changes @KentShikama , details in the commit message.

wittejm:
Oh and this needs a small frontend change to keep edit disposition working, I think. Forthcoming...

wittejm:
Just made another adjustment to make Answer Disposition work. I'm rather waffling on this, because now a change to Charge data (such as editing the disposition) doesn't need `edit_status` to be specified as an `UPDATE`; this is now treated as an update if the edit_status is not `ADD` or `DELETE`. 

Note also: this means that editing the Disposition (by answering the existing question) doesn't need to update the edit_status on the charge at all. We don't need to add the "edited" badge if the charge data was only altered by the user only through answering a question. This seems nice to me. 

As far as I can tell, with the latest commit the feature is now fully supported in the backend and none of the  existing app disrupted, so it can go into dev before I add any more frontend code for this feature.

src/backend/expungeservice/expunger.py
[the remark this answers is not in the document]
KentShikama: There are non-blocking non-traffic violations like juvenile charges. @michaelzhang43 Do those count under "The defendant has not been convicted of any other offense, excluding motor vehicle violations" in 137.226?

src/backend/expungeservice/record_editor.py
```diff
@@ -55,25 +62,46 @@ def _edit_case(case: OeciCase, case_edits) -> Tuple[OeciCase, List[Charge]]:
...
-            if charge.ambiguous_charge_id in charges_edits.keys():
+    def _edit_charges(
+        case_number: str, charges: Tuple[OeciCharge, ...], charges_edits: Dict[str, Dict[str, str]]
+    ) -> Tuple[Tuple[OeciCharge, ...], List[Charge]]:
+        charges_without_charge_type = [
+            charge for charge in charges if charge.ambiguous_charge_id not in charges_edits.keys()
+        ]
+        charges_with_charge_type = []
+        for edit_action_ambiguous_charge_id, edit in charges_edits.items():
+            if edit.get("edit_status", None) == EditStatus.ADD:
```
KentShikama, line 81: I think this should be extracted into _add_charges like before.

src/backend/expungeservice/record_editor.py
```diff
@@ -55,25 +62,46 @@ def _edit_case(case: OeciCase, case_edits) -> Tuple[OeciCase, List[Charge]]:
...
+                    "name": "N/A",
+                    "statute": "N/A",
+                    "level": "N/A",
+                    "type_name": "N/A",
+                    "balance_due_in_cents": 0,
+                    "edit_status": EditStatus.ADD,
+                }
+                new_charge = from_dict(data_class=Charge, data=charge_edits_with_defaults)
+                charges_with_charge_type.append(new_charge)
+            else:  # edit["edit_status"] is either UPDATE or DELETE
```
KentShikama, line 99: And similarly `_update_or_delete_charges`

wittejm:
Yep sound good. 

I tried out two ways to extract ` _update_and_delete_charge()` , between my latest and next-latest commits. I like the new one better.

PR #1269, opened 2020-06-23
Extract InvalidInput component for validation alert messages.

wittejm:
This removes a bit of repeated code. I'm going to use this component a bunch more times in the Edit Cases code.   

The tachyons styling is not identical in the two places that the alert messages appear now, in OeciLogin and Search. The difference is whether the vertical spacing is `mb3` or `mv4` respectively. @hmarcks let me know if this should be re-tweaked. 

(this PR was supposed to net a negative line count but I was thwarted by the prettier, booo)


src/frontend/src/components/InvalidInput/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface Props {
+  message: any;
+  condition: boolean;
+}
+
+export default class InvalidInput extends React.Component<Props> {
+  render() {
+    return this.props.condition ? (
+      <div role="alert" className="black-70 w-100">
```
hmarcks, line 11: The `role=alert` container should be in the dom before the message is inserted in order for screen readers to announce it: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19

src/frontend/src/components/RecordSearch/SearchPanel/index.tsx
```diff
@@ -127,7 +128,7 @@ export default class SearchPanel extends React.Component<Props, State> {
         <section className="cf mt4 mb3 pa4 bg-white shadow br3">
           <form className="mw7 center" onSubmit={this.handleSubmit} noValidate>
             {aliasComponents}
-            <div className="flex">
+            <div className="flex mb4">
```
hmarcks, line 131: This adds extra margin on the bottom of the search form when an error isn't present.

One thing we could do is remove the bottom padding on the parent `<section>` to make up for it, so `ph4 pt4` instead of `pa4`... 

Actually, that looks like it will give an awkward amount of margin above the error (`mb4`), so I'm thinking a better approach may be adding some top margin to the error component (probably `mv3`) so we don't have to have the spacing on the error containers.

src/frontend/src/components/InvalidInput/index.tsx
wittejm: Ok. I changed it to always render the `div` for the component, but that also means the page always renders the stack of empty `alert div`s before populating any of them. The layout looks fine, but should we avoid this and still wrap all of these `<p>` elements in a single container?

src/frontend/src/components/RecordSearch/SearchPanel/index.tsx
wittejm: I moved the spacing to only appear on the error banner, but it makes the ones on the OECI login to look awkwardly spaced... dang. This refactor might not actually be worth it?

wittejm:
@hmarcks  Yea, so, since this isn't actually super cookie-cutter and getting correct page layout beats out making the code a little simpler, I say screw it.

hmarcks:
Okay, I think we can adjust the OECI styles to accommodate the new spacing, but of course up to you

wittejm:
If you don't mind giving it another look! I have less mastery with margins, and didn't find a nice solution. Sorry to waffle about it.

src/frontend/src/components/OeciLogin/index.tsx
```diff
@@ -142,46 +143,62 @@ export default class OeciLogin extends React.Component<State> {
...
-                </div>
+                      help@recordsponge.com
+                    </a>
+                  </>
+                </InvalidInput>
+                <InvalidInput
+                  message={this.state.expectedFailureMessage}
+                  condition={this.state.expectedFailure}
+                />
                 <p className="lh-copy mt4">
```
hmarcks, line 168: I think we can change this `mt4` to `mt5` and remove the `mb4` from the log in button, and that should make the spacing look decent.

src/frontend/src/components/InvalidInput/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  message: any;
+  condition: boolean;
+}
+
+export default class InvalidInput extends React.Component<Props> {
+  render() {
+    return (
+      <div role="alert" className="black-70 w-100">
+        {this.props.condition && (
+          <p className="bg-washed-red mv4 pa3 br3 fw6">
```
hmarcks, line 13: I recommend changing this `mv4` to `mv3`

src/frontend/src/components/InvalidInput/index.tsx
hmarcks: I think a single container is ideal.

wittejm:
Great. I applied the spacing that you suggested and re-arranged so the `<div role="alert">` container always appears just once. Thanks @hmarcks !

hmarcks:
Nice, looks good to me!

PR #1282, opened 2020-06-27
Landing

Ark20:
moved partner table to it's own component


wittejm:
Is there a commit missing from this? I don't see a Partner component and the Disclosure structure is replicated manually several times.

Ark20:
@wittejm I'm not sure what happened there but I just pushed the changes I intended to before. In this I created a single component to hold the table but did you mean a component for each partner?

wittejm:
Yea. If you define the component just once then you could reuse it like

```
<PartnerInfo
    details=[
        {
            name:"Portland Community College"
            analysisCost:"Free"
            incomeRestrictions:"Under 45k Annual"
            locations:["Portland Metro",...]
            instructions:"Send an email with ..."
            contacts:["email@domain.com, "555-555-5555"]
            courtFees:"Not Included"
        }
</PartnerInfo>
```
Then you don't need to manually repeat the code style for each partner entry (DRY code style -- Don't Repeat Yourself). It will be a lot more readable and easier to maintain.

A lot of the detail rows also have the same style
```
<li className="flex-ns mb3">
    <span className="w4-ns db fw6 mr3">
        {key}
    </span>
    <span>
        value
    </span>
</li> 
```
So you could probably `map()` on those those data rows to generate the styling and repeat less code there too.

src/frontend/src/components/PartnerTable/index.tsx
```diff
@@ -7,6 +7,67 @@ import {
...
+                  <li className="flex-ns mb3">
+                    <span className="w4-ns db fw6 mr3">{detail[0]}</span>
+                    <span>{detail[1]}</span>
+                  </li>
+                ))}
+              </ul>
+              <p className="mw6 bt b--black-10 pv3">{partner.instructions}</p>
+              <ul className="list mb3">
+                <li className="fw6 mb3">{partner.contacts[0]}</li>
+                <li className="fw6 mb2">{partner.contacts[1]}</li>
```
wittejm, line 63: It's likely that partners will have varying number of lines of contact info (no phone number, etc), so you could `map()` these too.

src/frontend/src/components/PartnerTable/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+              </p>
+            </div>
+            <div className="ba bw3 br3 b--lightest-blue1 bg-white mb6">
+              <h3 className="f3 fw9 pa4">Partners</h3>
+              <ul className="list">{partners}</ul>
+            </div>
+
+            <span className="db w4 center bb bw2 b--blue mb3"></span>
+            <p className="tc white mw7 mh4">
+              Over 1,060 analyses delivered as of June 11, 2020
```
wittejm, line 91: I think this text can stay in the Landing page file.

PR #1283, opened 2020-06-28
Update Manual and add Paying Balances Link

swervinerv:
- updated the copy in the manual for Assumption 1 to include more information for paying off balances.
- updated the case and case balances component to include a link to the manual for paying balances


wittejm:
The in-banner link here is in bold but in the #1191 mockup it's not bold. I think having it not bold is helpful to keep it readable.

paying_balanced_bold

swervinerv:
@wittejm good catch, updated the font weight so it should now look like the following

Screen Shot 2020-06-27 at 9 11 54 PM

Issue #1063, opened 2020-04-22
Parse Arresting Agency for case


KentShikama:

michaelzhang43:
The DA Number is straightforward. It's always going to be on the right column right below the Case number in the "Register of Actions". The only complication is that it's not going to be there for many cases, especially cases below the misdemeanor level, or dismissals. 

<img width="190" alt="Screen Shot 2020-04-23 at 9 26 45 AM" src="https://user-images.githubusercontent.com/59623009/80124067-8e4e7c00-8544-11ea-92f1-0c3f24eed937.png">

hmarcks:
We can add it in the results here:

![district-attorney-number](https://user-images.githubusercontent.com/6887982/80560741-c839db00-89a7-11ea-86f0-f463cc3d94fe.png)

KentShikama:
Linking https://github.com/codeforpdx/recordexpungPDX/issues/1181

victorjz:
Parsing the DA number seems clear to me so I'll see if I can contribute that. Claimed assignment.

victorjz:
thanks for review and merge of the piece i could do. the other aspects of this issue i'm not so clear on, so i'll unassign for now. i'll browse for more issues and may even return to this one.

PR #1284, opened 2020-06-28
[WIP] issue #1063 parse DA

victorjz:
- this is strictly for parsing and testing the district attorney number
- unsure whether to go with least invasive additions or fully robust
- case_parser new member as optional so as not to have to add arguments when calling the constructor throughout the project
- you'll see in the new parsing method __parse_distr_atty that there's potential to return all the values from upper right of a given case html file but i just return DA number
- would it be preferred to return more values? or return None instead of empty str?
- also added testing methods to test_case_parser


src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -37,8 +39,19 @@ def feed(data) -> CaseParserData:
...
+    @staticmethod
+    def __parse_distr_atty(soup) -> str:
+        distr_atty_key = "District Attorney Number:"
+        labels = soup.find_all("th", "ssTableHeaderLabel", limit=10)
+        table = {}
+        for tag in labels:
+            table[tag.string] = tag.parent.find("td").string
+        if distr_atty_key in table:
+            return table[distr_atty_key]
+        return ""
```
wittejm, line 53:
```suggestion
        else: 
            return ""
```

wittejm:
Same thing with `CaseSummary` -- you can update all those constructor calls too.

victorjz:
I think I submitted my new changes?  This is my first time modifying code in a pull request.

src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -37,8 +39,20 @@ def feed(data) -> CaseParserData:
...
+    def __parse_district_attorney_number(soup) -> str:
+        district_attorney_key = "District Attorney Number:"
+        labels = soup.find_all("th", "ssTableHeaderLabel", limit=10)
+        table = {}
+        for tag in labels:
+            table[tag.string] = tag.parent.find("td").string
+        if district_attorney_key in table:
+            return table[district_attorney_key]
+        else:
+            return ""
```
KentShikama, line 54:
```suggestion
        DISTRICT_ATTORNEY_KEY = "District Attorney Number:"
        labels = soup.find_all("th", "ssTableHeaderLabel", limit=10)
        table = {tag.string: tag.parent.find("td").string for tag in labels}
        return table.get(DISTRICT_ATTORNEY_KEY, "")
```
Note I have not tested this.

victorjz: I ran the proposed code changes in a local directory and it works fine.

I'm not familiar with how to proceed with changes. Do I click commit suggestion or do I test everything in docker again and do another pull request?

KentShikama: Clicking accept commit suggestion works. You can also just commit locally and push the commit (not a new PR).

KentShikama:
Thanks @victorjz !

PR #1288, opened 2020-06-30
Full frontend functionality for Add/Edit case

wittejm:
Everything works correctly as far as I can tell!
update: located and fixed a few more bugs, detailed in 2nd commit message

I included an "Enable Editing" button, because all the add/edit buttons might unuseful to the user and distracting. 

I'm making this a feature branch in case because there's a fair amount more work to do. Feel free to PR to it, if there's anything bigger than some in-line suggestions.

This is a big feature and my logic doesn't feel amazingly robust, so there's possibly some bugs I missed. I searched pretty hard though. 

Some of the space styling gets a bit hacky like using the occasional empty div e.g. `<div className="pb3"/>` in `Record/index.tsx` line 133. There's probably some cleanup that can happen here.

@hmarcks I'm not really sure about the use of `aria-controls` and `aria-expanded` for the edit and add buttons: when one of the buttons is clicked, it disappears and the edit panel gets rendered. so `aria-controls` is always `false` , and the element that `aria-expanded` could point to never exists at the same time as the button. How should we handle this?

If it's important, I think this could be refactored to use a `Disclosure` as you suggested. I didn't jump at it because I thought it could be a tad complicated to use for both the add and edit actions -- clicking Add Case creates an entire new case element and then also the edit panel within it; whereas Edit just summons the edit panel. It's doable, so it might be worth it.

Comment on backend changes:

Selecting Charge Types by name requires having the full list of types, so I decided to hard-code the list in the frontend. That means we'll have to edit manually as our types change, which is a maintenance hassle. The alternative is to pull the list from the backend. This is something we COULD do, especially as part of the feature in #797 to show a full page of charge types documentation. 

Also I don't think there's a reasonable way to programmatically determine which types can only have one disposition of Convicted/Dismissed, so I hard-coded that info as well.

Comment on redux: It's a mess. The amount of different case-handling is pretty big, and I tried to cover everything I could think of.
Same thing with all the the different possible edit-submit states.

@KentShikama I saw you were doing some `useState()` stuff in the Disposition Questions feature that I didn't really get a handle on, but maybe that's a possble way to clean some stuff up?

### DIFF
--- a/src/frontend/src/components/RecordSearch/Record/Case.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/Case.tsx
@@ -1,13 +1,75 @@
 import React from "react";
-import { CaseData } from "./types";
+import { CaseData, ChargeData } from "./types";
 import Charges from "./Charges";
+import Charge from "./Charge";
+import EditCasePanel from "./EditCasePanel";
+import EditButton from "./EditButton";
+import AddButton from "./AddButton";
+import EditedBadge from "./EditedBadge";
 import currencyFormat from "../../../service/currency-format";
+import { undoEditCase } from "../../../redux/search/actions";
+import store from "../../../redux/store";
 
 interface Props {
   case: CaseData;
+  showEditButtons: boolean;
+  editing: boolean;
+  isNewCase: boolean;
+  whenEditing: Function;
+  whenDoneEditing: Function;
+  customElementId?: string;
 }
 
-export default class Cases extends React.Component<Props> {
+interface State {
+  editing: Boolean;
+  addingNewCharge: boolean;
+  nextNewChargeNum: number;
+  //nextBlankCharge?: ChargeData;
+}
+export default class Case extends React.Component<Props, State> {
+  createNextBlankCharge = (nextNum: number) => {
+    return {
+      case_number: this.props.case.case_number,
+      ambiguous_charge_id:
+        this.props.case.case_number + "-X" + ("00" + nextNum).slice(-2),
+      statute: "",
+      expungement_result: null,
+      name: "",
+      type_name: "",
+      date: "",
+      disposition: {
+        status: "",
+        ruling: "",
+        date: "",
+      },
+      probation_revoked: "",
+      edit_status: "ADD",
+    };
+  };
+  state: State = {
+    editing: this.props.editing,
+    addingNewCharge: false,
+    nextNewChargeNum: 1,
+  };
+
+  handleCaseEditSubmit = () => {
+    this.setState({
+      editing: false,
+      nextNewChargeNum: this.state.nextNewChargeNum + 1,
+    });
+    this.props.whenDoneEditing();
+  };
+
+  handleUndoEditClick = () => {
+    if (
+      !(this.props.case.edit_status === "ADD") ||
+      window.confirm("This data will be lost. Remove anyway?")
+    ) {
+      store.dispatch(undoEditCase(this.props.case.case_number));
+      this.props.whenDoneEditing();
+    }
+  };
+
   render() {
     const {
       name,
@@ -18,6 +80,7 @@ export default class Cases extends React.Component<Props> {
       charges,
       location,
       current_status,
+      edit_status,
     } = this.props.case;
     const allIneligible = charges.every(
       (charge) =>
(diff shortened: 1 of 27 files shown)


wittejm:
Ok an issue I just noticed is that answering any Question will cause the "Edited" badge to appear, and we don't want that. Will fix...

src/frontend/src/redux/search/reducer.ts
```diff
@@ -157,6 +163,154 @@ export function searchReducer(
         loading: action.ambiguous_charge_id,
       };
     }
+
+    case EDIT_CASE: {
+      const edits = JSON.parse(JSON.stringify(state.edits));
+      if (!edits[action.case_number]) {
+        edits[action.case_number] = {};
+      } // This check lets edits merge and not overwrite each other (e.g. AnswerDisposition vs EditCase)
```
KentShikama, line 171: The idomatic pattern seems to be `edits[action.case_number] = edits[action.case_number] || {}`

src/frontend/src/redux/search/reducer.ts
```diff
@@ -157,6 +163,154 @@ export function searchReducer(
...
+    case EDIT_CHARGE: {
+      const edits = JSON.parse(JSON.stringify(state.edits));
+      if (!edits[action.case_number]) {
+        edits[action.case_number] = {
+          summary: { edit_status: "UPDATE" },
+        };
+      }
+      if (!edits[action.case_number]["charges"]) {
+        edits[action.case_number]["charges"] = {};
+      }
```
KentShikama, line 210: Similar patterns to above

src/frontend/src/redux/search/reducer.ts
```diff
@@ -157,6 +163,154 @@ export function searchReducer(
...
+              : [];
+          if (ambiguous_charge_ids_on_case.length === 1) {
+            /*Undo only deleted charge on case means undo the deleted case.*/
+            edits = Object.keys(edits)
+              .filter((key) => action.case_number !== key)
+              .reduce((res: any, key) => ((res[key] = edits[key]), res), {});
+            return { ...state, edits: edits };
+          } else {
+            /*Undo deleted charge on a deleted case that has other charges means
+          we must mark the other charges as deleted.*/
```
KentShikama, line 266: I don't quite understand this comment. Maybe it would help to give an example?

src/frontend/src/redux/search/reducer.ts
```diff
@@ -157,6 +163,154 @@ export function searchReducer(
...
+            edits[action.case_number]["charges"]
+          )
+            .filter((key) => key !== action.ambiguous_charge_id)
+            .reduce(
+              (res: any, key) => (
+                (res[key] = edits[action.case_number]["charges"][key]), res
+              ),
+              {}
+            );
+          //edits[action.case_number] ["summary"]["edit_status"] = "UPDATE";
```
KentShikama, line 308: What is this comment for?

src/frontend/src/components/RecordSearch/Record/Case.tsx
```diff
@@ -1,13 +1,75 @@
...
+  addingNewCharge: boolean;
+  nextNewChargeNum: number;
+  //nextBlankCharge?: ChargeData;
+}
+export default class Case extends React.Component<Props, State> {
+  createNextBlankCharge = (nextNum: number) => {
+    return {
+      case_number: this.props.case.case_number,
+      ambiguous_charge_id:
+        this.props.case.case_number + "-X" + ("00" + nextNum).slice(-2),
```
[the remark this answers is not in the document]
wittejm: Then the user gets to feel clever for breaking the app by hitting "Add case" 100 times.

wittejm:
The new commit fixes a few bugs and applies @KentShikama 's suggestions. 

There's only one remaining behavior issue that I don't love. It's honestly only a minor annoyance IMO:

- If the user answers a Question that is on a charge,
- Then edits the charge,

Then the Question disappears because the user has specified the charge type in the edit. So far so good.

- Then the user reverts the edit, bringing the charge to its original state.

Then the Question re-appears on the charge, with the previous answer's radio button selected. But the selection hasn't actually taken effect to resolve the charge's type. It's mildly confusing. The user can just click the other option which will work fine, and then re-click the first one to properly resolve it as an answer to the Question.  

I haven't figured out a solution. But I've poked at it for a while and don't really want it to hold up the PR. I think if the user would only encounter this if they're really playing around with the editor to "see what would happen if I try this weird combination of actions"

hmarcks:
Wow this is looking great!

I'll take a look at the spacing soon, but here are some initial notes:

1. When you click to create a case, hide the edit and create charge buttons until the case is created. Not critical but I think it's more clear when those options aren't available yet.

2. Regarding aria-controls and aria-expanded: yeah the markup I provided is based off the [disclosure component](https://reacttraining.com/reach-ui/disclosure). Up to you of course, but we could create an issue for it if it doesn't make sense to tackle now.

3. I went to edit a Diverted DUII, and the Charge Type was left blank, and the Diverted DUII option was disabled. I'd expect it to be select when i go to edit it.

4. I see the Charge Name is styled as a <select> but it's actually text field. I assumed we would have a list we could select from, but I guess this isn't standardized enough? Do we just scrape the name from OECI?

5. If you trigger a validation (probation revoked date required), then select an option so that it no longer applies (Dismissed), the validation remains. I'd expect it to be removed. Not critical.

6. When we have open charges the Disposition is "Unkown", but when we provide options to edit a charge, there's an option for "Missing". Should these map?

Agreed about the revert edit of NMA issue you just mentioned - ideally the NMA question clears or it is correctly submitted, but not critical.

hmarcks:
4. I see the Charge Name is styled as a `<select>` but it's actually text field. I assumed we would have a list we could select from, but I guess this isn't standardized enough? Do we just scrape the name from OECI?

5. If you trigger a validation (probation revoked date required), then select an option so that it no longer applies (Dismissed), the validation remains. I'd expect it to be removed. Not critical.

6. When we have open charges the Disposition is "Unknown", but when we provide options to edit a charge, there's an option for "Missing". Should these map?

Agreed about the revert edit of NMA issue you just mentioned - ideally the NMA question clears or it is correctly submitted, but not critical.

wittejm:
> 1.

Ah yep, I introduced that issue with my new commit this morning. dang.
> 2. 

Sounds good. I'll probably be happy to tackle a refactor with Disclosure a few weeks from now :) 

> 3. 

I can't reproduce the issue you're describing. Is this happening after you've already answered the Diverted DUII question? Can you send me the example search you're using?

Although, relatedly I'm noticing that if you try to edit a charge that has an unanswered question, then the existing charge type isn't recognizable (it's just the concatenated string e.g. "DUII OR Diverted DUII") and the pre-selected charge type in the edit panel is "Civil Offense" -- the first valid option. I can fix this so that, if the type isn't determined yet the downdown will start with "Select..." , just like it does with Add New Charge. 

> 4. 

Yea, charge names are all over the place in OECI, so this can't really be a select-menu and is here purely as a convenience for the user. To fix the styling should I change anything besides remove the `<div<i></i></div>` here?  https://github.com/codeforpdx/recordexpungPDX/blob/72fa286149912cbff6e417035bf8b69d34b4d2d7/src/frontend/src/components/RecordSearch/Record/EditChargePanel.tsx#L325

> 5.

Good catch. I'll fix those to hide any validation banners as soon as they don't apply.

> 6. 

Yea it would be better if those match those match. Our existing feature to provide a missing disposition shows the option as Unknown so I will match that here.

wittejm:
Ok, @hmarcks I did fixes from your comments 1, 4, 5, and 6. Also I resolved the issue I described with NMA charges, so now undoing a charge edit causes the question answer to get unselected also. 

I also made the `charge_type` dropdown set to "Select..." if the current charge type is ambiguous. Otherwise, the `charge_type` and `disposition` should both have their current values pre-selected when you open the Edit Charge panel. Let me know if you're still getting that issue you described in point 3, and if so how I can reproduce it.

wittejm:
I think the charge names are helpful to keep untouched when the user selects a charge type because they are likely selecting something related to that original charge. For example if a Possession charge was incorrectly typed by the classifier as a B Felony but the charge name is still human-identifiable as a Possession, so the user can edit the type accordingly. Whereas, if the person is choosing a charge type unrelated to the original charge, they may as well be creating a new charge to achieve this. 

Thanks for spotting those two other bugs. I fixed those, and couldn't find any other issues.

PR #1289, opened 2020-07-01
Add Download PDF Icon and loading indicator

swervinerv:
- added an icon to the Download as PDF button for the search summary
- added a loading indicator that will be displayed while the PDF is being generated
- removed the extra loading button background color that is not needed


hmarcks:
Oh, too slow

KentShikama:
It is always nice to get an extra set of eyes.

PR #1304, opened 2020-07-06
Demo search feature

wittejm:
The demo intro text still needs styling. It looks pretty bland and hard to parse. @hmarcks would you mind tackling this?

Otherwise, the functionality and backend are ready for regular code review. 

This demo option is linked only from the OECI Login page. It could go on the landing page as well. With this plus the Add/Edit feature, we can share the app for anyone to try out and use its full functionality, minus fetching from OECI.

demo_page


hmarcks:
yeah, I'll do some formatting and look at what we might do on the landing page

KentShikama:
Could you post the mypy error you were trying to fix with the last commit?

KentShikama:
I would prefer if you manually dogfooded the records using the add charge feature, injected a print statement on the record, and then pasted those in.

wittejm:
> Could you post the mypy error you were trying to fix with the last commit?

I was initally building CaseSummary OeciCharge and by passing in dictionaries:
 
```
OeciCharge(
    **{
        **shared_charge_data,
        "ambiguous_charge_id": "001-1",
       ...etc.
```

Which ran ok, but was getting this mypy error:

`expungeservice/demo_records.py:141: error: Argument 1 to "OeciCharge" has incompatible type "**Dict[str, object]"; expected "str"`

So I reworked it to use `from_dict`

> injected a print statement on the record, and then pasted those in.

I don't know what you mean. What's the goal of this?

KentShikama:
It will force you to use the add charges feature. Have you ever used print on a record object (before it is serialized and returned to the front end) and checked it out with docker logs for debugging?

wittejm:
Gotcha, sounds reasonable.

wittejm:
Ok, I tried your suggestion for one case (Rosa Parks) and found a few things: 

First: some limitations on trying to produce a record with Add/Edit Case that captures all of our “interesting” data for the purpose of the demo examples.

The issues with this are:
- Can’t provide the person’s `name` for each Case, which is nice to have. 
- `location` can only be selected from Oregon counties, but it’s nice to have others for the demos, for example Montgomery county.
Here’s the result of manually creating a demo case, without touching the backend yet:

result_of_manual

Second: when reusing the generated & printed Cases data to create the demo data, some fiddling was required to get the print output of the Edited cases to have correct syntax and correct data to make it usable as a Search result. 

The main issue is that `RecordEditor.edit_search_results()` give a list of `Charge` objects, not a list of `OeciCharge`. To create a valid result from the mocked `search` function, I needed to convert Charge objects into OeciCharge – no big deal, except that the `charge_type` field is lost as a result. Here’s the output of this process:

creating_from_pasted_print

Whereas this is what we can get by creating the data manually in the backend:

backend_crafted_example

KentShikama:
So there would be two routes.

```
<Route component={RecordSearch} path="/record-search" />
<Route component={DemoRecordSearch} path="/demo-record-search" />
```

And two components would delegate to a shared component that looks like RecordSearch currently.

wittejm:
Ok, I overhauled the examples based on our meeting last night. I also made several examples. Here's a screenshot for @michellelsfong @michaelzhang43 to see the intro content, descriptions, and one of the returned examples (the longest and most interesting one).

I also added the new route as @KentShikama  suggested.

Screenshot_2020-07-08 RecordSponge Oregon

wittejm:
Well it seems at last minute I broke it from running locally? @hmarcks @KentShikama I'll let you know when it can actually run for review.

wittejm:
Maybe it's actually working fine? I started to get a redirect from `/demo-record-search` to the landing page every time, as though the AppRouter had stopped working. But it ended up working after I changed nothing. So lmk if that occurs for y'all.

src/backend/expungeservice/demo_records.py
```diff
@@ -0,0 +1,375 @@
...
+            for alias in aliases:
+                alias_lower = Alias(
+                    alias.first_name.lower(), alias.last_name.lower(), alias.middle_name.lower(), alias.birth_date
+                )
+                try:
+                    alias_search_result = DemoRecords.records.get(alias_lower, [])
+                    search_results += alias_search_result
+                except Exception as e:
+                    errors.append(str(e))
+                    print(e)
```
KentShikama, line 33: Unnecessary print

src/frontend/src/redux/search/actions.ts
```diff
@@ -57,7 +59,7 @@ function buildSearchRequest() {
 
 function buildAndSendSearchRequest(dispatch: any): any {
   return apiService<SearchResponse>(dispatch, {
-    url: "/api/search",
+    url: store.getState().search.demo ? "api/demo" : "/api/search",
```
KentShikama, line 62: /api/demo

hmarcks:
Functionality is working well for me. I pushed some styles and updated the headers to give more info about what type of charges they have instead of repeating the Date of Birth. Made more sense to me but let me know if you don't agree.

wittejm:
Ah nice. Yep, this looks much better.

wittejm:
Replaced with #1317

PR #1305, opened 2020-07-07
Replaced diamond with OR in record_merger and tests.

### LINKED ISSUE #1207: Switch from using ⬥ to OR in charge eligibility labels

wittejm:
Resolves #1207 

Hmm well one example looks like this, which is pretty cluttered. Maybe make these bold-text? 

messy_ors

vs what we have now:

with_diamonds

@KentShikama @hmarcks @michaelzhang43

### DIFF
--- a/src/backend/expungeservice/models/charge_types/duii.py
+++ b/src/backend/expungeservice/models/charge_types/duii.py
@@ -21,7 +21,7 @@ Therefore, to determine whether a dismissal is eligible, ask the client whether 
             raise ValueError("Dismissed criminal charges should have been caught by another class.")
         elif ChargeUtil.convicted(disposition):
             return TypeEligibility(
-                EligibilityStatus.INELIGIBLE, reason="137.225(7)(a) - Traffic offenses are ineligible"
+                EligibilityStatus.INELIGIBLE, reason="Traffic offenses are ineligible under 137.225(7)(a)"
             )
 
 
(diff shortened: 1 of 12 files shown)


hmarcks:
Agreed, could try these:
1. Bold like you mention
2. Each OR on new line
3. Replace all text with a "?" because it's easy to toggle between answers

wittejm:
> Replace all text with a "?" "Needs clarification below" because it's easy to toggle between answers

There are the cases where time eligibility is ambiguous because of some other open charge, like here:

showing-OR

And I think it's useful to keep the Time eligibility reason string since the user will have to go hunting elsewhere in the record to resolve the ambiguity.

I'll make the bold version and we can see how that looks for various examples.

wittejm:
A few attempts:

Just bolding the OR:

bold_ors_no_breaks

a leading OR on newlines:

leading_linebreak_no_border

Adding a border line between Type names and Type Reasons, because these seemed too smooshed together above. 

leading_line_break_and_borderline

hmarcks:
I like each on their own line, and I think the border would be more clear if it was between the ORs

wittejm:
> I think the border would be more clear if it was between the ORs

Could you clarify what you have in mind? Currently we're dealing with:

**Type:** Type1 OR Type2 OR Type3
 Reason 1 (& statute) OR Reason 2 (& statute) OR Reason 3 (& statute)  

if the bar goes between every OR we would have (as a crude markdown  version):

**Type:** Type1 
\--
OR Type2
\--
OR Type3
\--
Reason1 
\--
OR Reason2
\--
OR Reason3

Is this what you mean?

hmarcks:
Ah I was reading it wrong, I think ideally the reason should be paired with the type it corresponds to (similar to Time), so:
**Type:** Type1 - Reason1
--
OR Type2 - Reason2
--
OR Type3 - Reason3

wittejm:
Cool. I'll make that update so we can easily view a range of examples. Once we can test that out it might be good to tweak the widths of the two columns again.

wittejm:
Here's an example with the new style:

ambiguous_with_breaks

@hmarcks  The vertical connector is a little wacky now, there's a tiny segment appearing below the types list

hmarcks:
yeah I fixed it in #1316

hmarcks:
Oh and I think they new style looks good

PR #1307, opened 2020-07-08
Add form filling feature using stock conviction/arrest forms

KentShikama:


KentShikama:
Screen Shot 2020-07-08 at 4 11 42 PM

@hmarcks I'll need some suggestions on how to style this "Fill expungement forms" button.

wittejm:
Some suggested behavior changes:

1. If no charges are eligible, don’t allow clicking the button. Maybe it's grayed out and has a tooltip that is specific to whether its clickable or not, or some other presentation.

2. Windows (7, at least, and earlier versions) can't open .tar files by default. Maybe 8 and 10 can? But let's make this a .zip file. If .zip doesn't work by default on Mac, then that's lame and we could make the download button OS-sensitive or offer manual selection for compression format. 

3. The expungement page can use some explanation for what it's doing: Like with a block of text on the left and the fillable fields on the right. I suggest the following:

> This will fill and download the required pdf forms for the [#] cases that have charges eligible for expungement:
> [Case number 1]
> [Case number 2]
> [Case number 3]
> 
> Please read the complete [instructions in the manual](url) for filing the required forms for expungement. After downloading the pdfs, review their contents to verify that all the required information is present and correct. 
> 
> 
> The following required information is obtained from OECI and will be provided in the form:
> Case number
> Name of charge 
> Date of arrest 
> Date of conviction
> 
> The following information might be missing from OECI. If available, it will be provided in the form. If it is not present in OECI, some of the information may or may not be required in the application; please consult the manual. 
> Arresting Agency
> DA number
> ...etc.

4. Add validation for missing or badly-formatted values for the fillable fields on this page. 

5. Pre-fill the Full Name and Birth Date fields with the contents in the first Alias on the search page.  

6. the folder name created in the archive is generated as `/tmp/tmpcm6oohbw` or such; let's name it `expungement_packet` to match the archive name.

KentShikama:
> Add validation for missing or badly-formatted values for the fillable fields on this page

Which fields were you thinking of? We want all fields to be optional in case the user wants to fill them out in pen/pencil as they may be uncomfortable providing that information into the software.

wittejm:
> We want all fields to be optional in case the user wants to fill them out in pen/pencil as they may be uncomfortable providing that information into the software.

Ah, good point. How about format validation on the phone number, zip code, and birth date; and a selection menu for state with Oregon pre-selected.

Also an explanation of what this form is doing, like "You can also optionally provide your identifying information here to pre-populate the forms. We do not save any of your information."

KentShikama:
@wittejm PTAL. I've addressed a couple issues and made Github issues for the rest.

KentShikama:
Screen Shot 2020-07-09 at 4 29 34 PM

@hmarcks @wittejm This is how the buttons looks now. It only shows if there are charges that are "Eligible Now" under that section.

wittejm:
The button could go next to the "Eligible Now" section header and be a quiet-ish style similar to the Enable Editing button. @hmarcks 's call though. I think the #1312 content should go on this page before it's merged, since the tool is not very self-explanatory.

Code LGTM though.

KentShikama:
Screen Shot 2020-07-09 at 4 52 38 PM

@wittejm How about I put an (Experimental) label on it for now so that users will be cautioned off? https://github.com/codeforpdx/recordexpungPDX/issues/1312 would requiring working on HTML styling and I was hoping @hmarcks could follow up on that in a later PR. I do think the tool is pretty self-explanatory though. You click the button, fill out information, and you get a bunch of PDFs. What the PDFs are for isn't self-explanatory but we are no worse off than if you just randomly linked someone who hasn't read the Manual to the Appendix.

KentShikama:
@wittejm On second thought, the button is a bit too conspicuous for an experimental feature. I've converted the link to be the almost invisible period (after the (5)) as follows:

Screen Shot 2020-07-09 at 5 01 27 PM

PR #1309, opened 2020-07-09
Add state to control up/down icons on landing page

Ark20:


wittejm:
This is using a shared state for all the arrows, so if I click one it looks like this:
shared_arrow

But it should have individual states so that a single arrow click flips only that arrow.

Ark20:
Just realized the latest commit doesn't quite work either. Any suggestions on setting the state for each? @wittejm

wittejm:
Looks like you can set the `open` property on the Disclosure: https://reach.tech/disclosure/
so this could be set as `open={index===this.state.active}`

That way that opening one detail panel would cause the others to collapse, which seems like nice behavior.

You would also need to make sure the the button click doesn't just toggle the `state.open` flag, since clicking a different disclosure should leave the `state.open = true`.

Or, how about tossing the `open` state variable and just setting  `state.active = -1` initially and when no panels are open. Then use:
```
const toggleOpen = (order: any) => {
      this.setState({ active: order === active? -1 : order });
}
```

I think that will work?

Issue #1212, opened 2020-06-10
Gracefully handle when OECI token expires


KentShikama:

wittejm:
completed with #1310

PR #1310, opened 2020-07-09
OECI token expires on closing the browser instead of a time limit.

wittejm:
A solution for #1212 

We used to set the cookie time limit to match OECI, to which is about 5 minutes. We extended it to 2 hours, but there's no reason for it not to last for the entire session. The session cookie now expires on closing the browser, with is a predictable behavior. 

Because our search backend creates a new login session with OECI upon every search, there's no risk of the OECI server session expiring.


hmarcks:
Sounds good to me

PR #1323, opened 2020-07-11
Partner Page with Mailchimp signup

wittejm:
I snafu'd the git history on the previous PR, but this one should have everything from before. The embedded Mailchimp form appears to be working properly. 

I fiddled some with the styling but it doesn't look great, could you polish it off @hmarcks ?

### DIFF
--- a/src/frontend/src/components/RecordSearch/Record/TypeEligibility.tsx
+++ b/src/frontend/src/components/RecordSearch/Record/TypeEligibility.tsx
@@ -13,7 +13,10 @@ export default class RecordType extends React.Component<Props> {
 
     const eligible = (reason: string) => (
       <div className="relative mb3 connect connect-type">
-        <i aria-hidden="true" className="absolute fas fa-circle z-1"></i>
+        <i
+          aria-hidden="true"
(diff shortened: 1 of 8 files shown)


src/frontend/src/components/RecordSearch/Record/TypeEligibility.tsx
```diff
@@ -24,7 +27,7 @@ export default class RecordType extends React.Component<Props> {
       <div className="relative mb3 connect connect-type">
         <i
           aria-hidden="true"
-          className="absolute fas fa-question-circle purple bg-white z-1"
+          className="absolute fas fa-question-circle purple bg-white outline-2-white z-1"
```
wittejm, line 30: @hmarcks I was having trouble figuring out where this style update came from, and not sure if I put it back in there by accident. Do we still want this change to go in? Same thing for Time Eligibility.

hmarcks: Yes please! Confusing because I snuck it in this branch but not related to the issue

Issue #1239, opened 2020-06-13
Create editable forms out of forms in https://recordsponge.com/manual#appendix


KentShikama:
Field names should be the same as https://www.mcda.us/index.php/documents/conviction-motion-and-affidavit-form.pdf/ if applicable.

wittejm:
Making this lower priority than our webapp dev issues. Let's be confident that our software users think this should become a priority.

KentShikama:
Multnomah county is done with https://github.com/codeforpdx/recordexpungPDX/pull/1321; potentially 35 more to go.

PR #1330, opened 2020-07-13
Update custom PDF forms for Clackamas county

KentShikama:
We add a new field called "case_number_with_comments". Since the Clackamas county forms do not require you to write down the charges that are to be expunged, we need to specify which charges are eligible in the case number field as a comment. For example, "Case Number 0123456 (in part, counts 1,2,4)".


src/backend/expungeservice/endpoints/pdf.py
```diff
@@ -214,13 +227,17 @@ def build_pdf_path(case, convictions):
...
+                return path.join(Path(__file__).parent.parent, "files", "clackamas_conviction.pdf")
             else:
                 return path.join(Path(__file__).parent.parent, "files", f"stock_conviction.pdf")
         else:
             if case.summary.location.lower() == "multnomah":
                 return path.join(Path(__file__).parent.parent, "files", f"multnomah_arrest.pdf")
             elif case.summary.location.lower() == "jackson":
                 return path.join(Path(__file__).parent.parent, "files", f"jackson_arrest.pdf")
+            elif case.summary.location.lower() == "clackamas":
+                return path.join(Path(__file__).parent.parent, "files", f"clackamas_arrest.pdf")
```
KentShikama, line 240: With the next county, I'll clean this up to

```python
if case.summary.location.lower() in SUPPORTED_COUNTIES:
    return path.join(Path(__file__).parent.parent, "files", f"{case.summary.location.lower()}_arrest.pdf")
else:
    return path.join(Path(__file__).parent.parent, "files", f"stock_arrest.pdf")
```

KentShikama:
Linking https://github.com/codeforpdx/recordexpungPDX/issues/1239

PR #1332, opened 2020-07-13
Tweaks in Extra Info panel

wittejm:
I went through the extra rules for each type and decided it's actually worth showing it for every type. Along the way I made these tweaks. 

- Some re-wording of rules
- Simplify frontend types
- Add linebreaks (divs) in rules. 
- More Info button isn't bold text. 
- A little less spacing


wittejm:
Looks great. Thanks @hmarcks !

PR #1354, opened 2020-07-21
added da number to case info

Ark20:
Currently if there is a - in the Distr Atty number there is a line break. Is this fine or do we want to avoid this ? cc: @hmarcks


hmarcks:
I think we want to avoid it so I made the column wider, but I see some numbers can get quite long so I am forcing a break rather than having it overlap the next column.

Things are getting a bit crowded in the header especially at smaller screen widths so I'm going to look at a redesign, but I pushed an update that makes it decent for now.

Oh @Ark20 could you insert a "–" if no DA number is returned instead of a blank space? Would like to do that for the other items like DOB as well. Let me know if that should be a separate issue, thanks.

PR #1357, opened 2020-07-21
Generate Forms page documentation

wittejm:
Static language to explain the feature. In the future we could generate a dynamic report of what files will be downloaded.

Also add the Generate Forms button. If there are no eligible charges, this in-line error is shown on click. 

cant_generate_error

The page spacing is bad and I couldn't figure out how to do it in columns. @hmarcks Could you give it a redo?
Screenshot_2020-07-21 RecordSponge Oregon

### DIFF
--- a/src/frontend/src/components/FillForms/UserDataForm.tsx
+++ b/src/frontend/src/components/FillForms/UserDataForm.tsx
@@ -86,13 +86,13 @@ class UserDataForm extends React.Component<Props, State> {
   public render() {
     return (
       <>
-        <main className="mw6 ph2 center">
-          <section className="cf mt4 mb3 pa3 pa4-l bg-white shadow br3">
-            <h1 className="mb4 f4 fw6">User Information</h1>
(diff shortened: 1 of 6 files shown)


hmarcks:
Yeah taking a look now...

src/frontend/src/components/FillForms/UserDataForm.tsx
```diff
@@ -120,7 +120,7 @@ class UserForm extends React.Component<Props, State> {
                 />
               </div>
               <div className="mb4">
-                <label htmlFor="mailingAddress" className="db mb2 fw6">
+                <label htmlFor="mailingAddress" className="db mb1 fw6">
                   Mailing Address
```
KentShikama, line 124:
```suggestion
                  Mailing Street Address
```

KentShikama:
The Phone Number field should probably be moved to the bottom to make it clearer that the City/State/Zip is grouped with the Mailing Street Address.

PR #1366, opened 2020-07-23
Issue 1256: GitHub workflow to push to prod on release.

### LINKED ISSUE #1256: Make Github workflow that triggers a push to prod upon new release tag

erikjasso:
Runs when a release is made via the GitHub interface for making a new release or on pushing a tag of the form "v*" to master.

Tested the trigger on my fork.

Fixes #1256.

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -4,8 +4,8 @@ on:
   release:
     types: [created]
   push:
-    branches:
-      - tags/v*
+    tags:
+      - v*
 
 # Copied from staging.yml.
 # GitHub does not support sharing steps between workflows yet.
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
       - name: Test
-        run: echo "Hello, world!"
+        run: echo "Hello, world!!"
 
     # steps:
     #   - name: Install dependencies
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -5,7 +5,7 @@ on:
     types: [created]
   push:
     branches:
-      - refs/tags/v*
+      - tags/v*
 
 # Copied from staging.yml.
 # GitHub does not support sharing steps between workflows yet.
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -5,9 +5,9 @@ on:
     types: [created]
   push:
     branches:
-      - master
-    tags:
-      - v*
+      - [1256-prod-github-workflow] #master
+      tags:
+        - v*
 
 # Copied from staging.yml.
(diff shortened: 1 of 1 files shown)


.github/workflows/prod.yml
```diff
@@ -0,0 +1,48 @@
+name: deploy to prod
+
+on:
+  release:
+    types: [created]
+  push:
+    branches:
+      - master
```
KentShikama, line 8: Does this not trigger this workflow when a PR is merged to master? Or is this just the syntax that Github actions has instead of something like the following:

```
on:
    push:
        tags:
            - v*
```

erikjasso: Good catch. You're right. That will run when pushed to master or a tag is pushed. On the other hand we don't want it to run when a tag is pushed to any branch other than master.

loganballard: Is there a way to do some sort of "and" logic here?  I've been combing the github actions documentation and haven't found anything.  Right now it seems like we are running on 

(release created) OR (push to master OR tag pushed)

What we want is 

(release created) OR (push to master AND tag pushed)

or ideally just:

(tag pushed to master)

loganballard: Based on the documentation on the [create workflow trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#create), it looks like github actions are triggered using the [webhook for create](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#create), which will have the information in the "ref_type" key/value pair whether it is a branch or a tag that is created.  Is it possible to leverage that from github actions as a trigger?  Not sure.

erikjasso: YML syntax doesn't like mixing types in dicts, so this, which to me would mean AND of "push to master" and "tag like v*", doesn't work:

    on:
      push:
        branches:
          - master
          tags:
            - v*

erikjasso: I think @loganballard is on to something w/using refs. Isn't there a path using refs that means "tags on master"?

Or maybe we want to use release tags as the filter?

.github/workflows/prod.yml
```diff
@@ -5,7 +5,7 @@ on:
     types: [created]
   push:
     branches:
-      - refs/releases/tag/v*
+      - refs/tags/v*
```
KentShikama, line 8: This seems fine to me.

erikjasso:
This is what works (meaning I push a tag name v* to my branch and see on GitHub that it ran the action):

    on:
      push:
        tags:
          - v*

erikjasso:
I can't tell -- does the web interface allow the merger to squash commits when merging? If not I'll squash on my branch and push a single commit to be merged.

KentShikama:
You can squash merge in the web interface indeed from the drop down. The main logic is still commented out on this PR though.

KentShikama:
Looks good!

PR #1376, opened 2020-07-27
Exclude removed charges when generating forms.

### LINKED ISSUE #1360: Cases that are edited to be removed still appear in Generate Forms
This seems to be a side-effect of having removed charges still have a label of "Eligible Now" if that is the case.

wittejm:
Resolves #1360 

I considered an alternate approach of checking the EditStatus when the charge eligibility gets originally set to `ELIGIBLE_NOW` at 
https://github.com/codeforpdx/recordexpungPDX/blob/d69322e86181ef3c4bbbc3d70e03ddbc2bc63a26/src/backend/expungeservice/record_merger.py#L181

But it feels weirdly out of place to stick Edit functionality info into the merger. Regardless of where we check Edit status, it feels like we're patching it in in an inelegant fashion, so I chose to avoid touching the already-really-complex merger code.

### DIFF
--- a/src/backend/expungeservice/endpoints/pdf.py
+++ b/src/backend/expungeservice/endpoints/pdf.py
@@ -134,10 +134,9 @@ class FormFilling(MethodView):
 
     @staticmethod
     def _build_pdf_for_case(case, user_information):
+        charges = [c for c in case.charges if c.edit_status != EditStatus.DELETE]
         ineligible_charges_generator, eligible_charges_generator = partition(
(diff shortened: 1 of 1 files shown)


KentShikama:
Hmm, this does mean that deleted charges will still appear in the forms under charges_all.

KentShikama:
I use case on line 157 for charges. We should probably pass down charges and case summary such that we are guaranteed to never use case.charges again.

wittejm:
Rather than separate out summary and charges, I decided to move the edit up further. PTAL @KentShikama

Issue #1377, opened 2020-07-27
Documention page for charge types rules


wittejm:
This feature is in the second half of this mockup from @hmarcks 
![charge_type_doc_page](https://user-images.githubusercontent.com/2104990/88580266-39694880-d000-11ea-9461-85e54b8c4c39.png)

wittejm:
completed with #1391

Issue #1308, opened 2020-07-09
Update Manual with Add/Edit feature


hmarcks:
We want documentation for #1079

wittejm:
Here's a rough draft on this section. These in-line images should be the html actual visuals. The charge-type select feature kind of requires the documentation page of #797 #1277 to be useful to a new user.

**Add & Edit Cases**

Sometimes the result of a Search doesn’t completely match a person’s true record. This can happen for a few reasons:

- The search returns the wrong person’s record: In the rare event that a different person’s Oregon reccord shares a birth date and name with the person of interest, these records may appear in the search result.
- One or more of a person’s cases cannot be found: If the record has a typo in the name or the incorrect birth date, it can be difficult to locate by searching OECI.
- A charge date or other piece of information for a particular case is missing or incorrect in OECI.
- The expungement analyzer could not classify the charge: Many of the records in OECI are older and have statutes that have been renumbered or repealed; or a charge has an irregular name not recognized by our analyzer.

If any of these issues appear in the record, they can be corrected with the Editing feature. You can also use the Editing feature to build a record from scratch and run an analysis without relying on OECI at all. 

Click the Enable Editing button after running a search:

![enable_editing_row](https://user-images.githubusercontent.com/2104990/87096469-c5edcb80-c1f7-11ea-8c11-5dbc7376d638.png)

You can now perform the following actions to change the contents of the record. These changes exist only temporarily in your session and will disappear if you run another search. The edits will persist if you navigate between Search and the Manual or other pages on recordsponge.com, but they will disappear if you leave the website.

Add Case: ![add_case_button](https://user-images.githubusercontent.com/2104990/87096362-7c04e580-c1f7-11ea-83ff-0f8097a7e5cc.png)

Create a new case from scratch that can be populated with new charges. You must provide the case Current Status, County, Balance, and Birth Year, all of which are used to provide complete analysis of the record.

Edit Case: 
You can edit any of these fields on an existing or newly created case by clicking the edit button ![img](https://user-images.githubusercontent.com/2104990/87096668-2c72e980-c1f8-11ea-8f10-64cc0510101f.png) in the case's header.  

Remove Case: 
Remove a case from the analysis by clicking the edit button and then Remove Case in the editing panel. With will display a [removed badge] badge on the case as well as all of the charges in the case. Any charges have have been removed will no longer affect the time eligibility of other charges.

Add Charge:
[...]

PR #1391, opened 2020-07-30
Editing guide in Manual and Charge Types page

### LINKED ISSUE #1384: Diverted DUII "More Info" entry says "not added yet"
The rules are defined on the Class, so investigate this bug.

wittejm:
Resolves #1384, #1377 , #1308 
Charge Types are given in a separate page that is linked in the Editing Results section in the Manual, and in the Edit Charge Panel. This means a user will only come across it after they start digging into Editing. 
   
I finally abandoned the idea of a Complete Document of Time Eligibility. The list of charge types is still essential for use of the Editing feature though. 

This PR changes the `expungement_rules`of many charge types. They're no longer being formatted with markdown and, because I wanted to retain the bullet-pointed lists, I chose to define the content from the backend with some nested tuples that get used to build HTML in the frontend. 

There are two small spacing issues I couldn't figure out. @hmarcks Could you tackle these? 

Overhanging icon:
overhang_editing_icon

Off-center:
off-center-type-icon


src/frontend/src/components/Manual/index.tsx
```diff
@@ -539,20 +547,19 @@ class Manual extends React.Component {
...
+                be directed to input identifying information. Inputting
+                identifying information is optional but can help save you time.
               </p>
               <p className="mb4">
-                After you input the information, RecordSponge will generate 
-                a .zip file with PDFs of the expungement paperwork for all 
-                of the charges.
+                After you input the information, RecordSponge will generate a
+                .zip file with PDFs of the expungement paperwork for all of the
+                charges.
```
KentShikama, line 562: One PDF per case with any eligible charge

src/frontend/src/components/Manual/index.tsx
```diff
@@ -598,64 +605,66 @@ class Manual extends React.Component {
                     </p>
                     <h4 className="fw7 mb2">Conviction versus Arrest</h4>
                     <p className="mb3">
-                      If the Disposition of a case is “Dismissed,” use an Arrest Form.
+                      If the Disposition of a case is “Dismissed,” use an Arrest
```
KentShikama, line 608: If there are no convicted charges in the case

src/frontend/src/components/Rules/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  };
+
+  fetchRules = () => {
+    axios
+      .request({
+        url: "/api/rules",
+        method: "get",
+      })
+      .catch((error) => {
+        return Promise.reject(error);
```
KentShikama, line 35: Is this necessary?

wittejm: It is not!

wittejm:
@KentShikama LMK if this looks good to go.

Issue #1268, opened 2020-06-23
Fix implemenation of `Expunger._is_other_convictions_all_traffic(convictions)`


KentShikama:
There are non-blocking non-traffic violations like juvenile charges.

PR #1401, opened 2020-08-02
#1268: Expunger checks if all other convictions are traffic related

0x6362:
Previously, the `_are_other_convictions_all_traffic` function was passed in
`convictions` that included **only convictions from blocking charges**. This
could lead to improper handling of records with juvenile cases.

The desired behaviour was that we want to capture **all** charges that aren't
traffic, not all **blocking** charges that aren't traffic.


src/backend/expungeservice/expunger.py
```diff
@@ -30,15 +30,20 @@ def run(record: Record) -> Dict[str, TimeEligibility]:
...
-            most_recent_blocking_conviction = Expunger._most_recent_convictions(convictions)
-            other_convictions_all_traffic = Expunger._is_other_convictions_all_traffic(convictions)
+
+            other_charges = [c for c in analyzable_record.charges
+                             if c.id != charge.id and c.edit_status != EditStatus.DELETE]
+
+            other_blocking_charges = [c for c in other_charges
+                                      if c.charge_type.blocks_other_charges]
+
+            dismisals_all, convictions_all = Expunger._categorize_charges(other_charges)
```
wittejm, line 40:
```suggestion
            _, convictions_all = Expunger._categorize_charges(other_charges)
```
Since this variable isn't used.

src/backend/tests/test_crawler_expunger.py
```diff
@@ -254,3 +254,24 @@ def test_expunger_for_record_with_mj_over_21(record_with_mj_over_21):
             status=EligibilityStatus.ELIGIBLE, reason="Eligible now", date_will_be_eligible=date(2001, 3, 3)
         )
     }
+
+@pytest.fixture
+def record_with_juvenille_and_traffic_charges():
```
wittejm, line 259:
```suggestion
def record_with_juvenile_and_traffic_charges():
```

src/backend/expungeservice/expunger.py
```diff
@@ -30,15 +30,20 @@ def run(record: Record) -> Dict[str, TimeEligibility]:
...
+            dismissals_blocking, convictions_blocking = Expunger._categorize_charges(other_blocking_charges)
+
+            most_recent_blocking_dismissal = Expunger._most_recent_different_case_dismissal(charge, dismissals_blocking)
+            most_recent_blocking_conviction = Expunger._most_recent_convictions(convictions_blocking)
+
+            other_convictions_all_traffic = Expunger._are_other_convictions_all_traffic(convictions_all)
```
KentShikama, line 46: Nitpick. I would rename as follows:

```
dismisals_all, convictions_all -> dismissals, convictions
dismissals_blocking, convictions_blocking -> blocking_dismissals, blocking_convictions
```

src/backend/expungeservice/expunger.py
```diff
@@ -194,7 +199,7 @@ def _most_recent_convictions(recent_convictions):
             return newer
 
     @staticmethod
-    def _is_other_convictions_all_traffic(convictions):
+    def _are_other_convictions_all_traffic(convictions):
```
KentShikama, line 202: While using `are` might be grammatically correct, I've always seen `is` used for any boolean return type method name where applicable.

src/backend/tests/test_crawler_expunger.py
```diff
@@ -254,3 +254,24 @@ def test_expunger_for_record_with_mj_over_21(record_with_mj_over_21):
...
+        {
+            "X0001": CaseDetails.CASE_PARKING_VIOLATION,
+            "CASEJD1": CaseDetails.CASE_MJ_CONVICTION,
+        },
+    )
+
+# FIXME: Unclear to me what the best way to create a minimal test case
+# representing a record that would pass under the old logic but fail under this
+# revised logic
+def test_expunger_for_record_with_non_blocking_other_charges(record_with_mj_over_21):
```
KentShikama, line 271: Use `YoungDoe.SINGLE_CASE_RECORD` with a convicted CivilOffense charge. See `record_with_mj_under_21`.

KentShikama:
Will resolve https://github.com/codeforpdx/recordexpungPDX/issues/1268

src/backend/expungeservice/expunger.py
0x6362: Will revert this in the next revision.

src/backend/tests/test_crawler_expunger.py
0x6362: @wittejm had pointed me towards using the factories that are used in some of the other test suites; is there a preference one way or the other? 

I had started working on their suggestion, but was having a bear of a time getting an `MarijuanaUnder21` charge to be produced (even when injecting a sufficiently young birth year into the Case/CaseSummary) - should it only be tagged as such after passing through the Expunger, since the factories don't seem to actually run through the charge analyzer?

wittejm: Oh you're right, sorry, that may have been a bad suggestion. If all the required HTML fixtures are there, then that makes sense to use those.

KentShikama: I don't have a preference. Using the CrawlerFactory will require you to create a new `CaseDetails.CASE_MJ_CONVICTION_WITH_CIVIL_OFFENSE` charge. Or you could just modify the existing `CaseDetails.CASE_MJ_CONVICTION` to include a civil offense charge.

0x6362: The right way to do either of those would be to create a new HTML file, correct?

src/backend/tests/test_time_analyzer.py
```diff
@@ -634,3 +635,37 @@ def test_marijuana_violation_eligible_with_prior_conviction():
...
+        date=mj_violation_date,
+        disposition=DispositionCreator.create(ruling="Convicted", date=mj_violation_date),
+    )
+
+    case_1 = CaseFactory.create(
+        info=juvenile_doe_info,
+        type_status=["Juvenile Delinquency: Misdemeanor", "Closed"],
+        case_number="1",
+        charges=[marijuana_violation]
+    )
```
KentShikama, line 656: I would just make a test with one Case and two charges. You may have to include the case number in the two charges like follows:
```
ChargeFactory.create(
        case_number="1",
```

src/backend/tests/test_crawler_expunger.py
KentShikama: Indeed. Make sure to look at the rendered HTML of existing examples. It should just involve copy/paste, add a new line for the charge, and then a new line for the disposition. The events part is irrelevant in this context.

src/backend/tests/test_crawler_expunger.py
```diff
@@ -254,3 +258,69 @@ def test_expunger_for_record_with_mj_over_21(record_with_mj_over_21):
...
+
+def test_expunger_for_record_with_mj_under_21_with_non_traffic_charge_not_subject_to_137226():
+    record = CrawlerFactory.create(
+        record=YoungDoe.SINGLE_CASE_RECORD,
+        cases={"CASEJD1": CaseDetails.CASE_MJ_AND_JUVENILE_CONVICTION})
+
+    expunger_result = Expunger.run(record)
+
+    assert isinstance(record.charges[0].charge_type, MarijuanaUnder21)
+    assert isinstance(record.charges[1].charge_type, JuvenileCharge)
```
0x6362, line 305: I can not trace back what is meant to be modified in the fixture to produce a JuvenileCharge; it seems like it should be controlled by the case type or violation level, but there were no examples and I spent enough time on it to call for help -- see `./case_html/mj_with_juvenile_offense.html`

KentShikama: Just do a `CivilOffense` charge by putting "fugitive" in the charge name.

0x6362: This worked 👍

0x6362:
OK, final question: I have a case with an MJ <21 conviction + a class A Felony committed this year.

1. Is the MJ Charge still dismissable
2. Should the dismiss date be: 3yrs or 10 years on?

KentShikama:
@0x6362 Indeed, the MJ is blocked by the class A felony for 10 years. It will then become eligible.

src/backend/tests/test_crawler_expunger.py
```diff
@@ -254,3 +257,72 @@ def test_expunger_for_record_with_mj_over_21(record_with_mj_over_21):
...
+
+def test_expunger_for_record_with_mj_under_21_not_blocked_by_traffic(record_with_mj_under_21_and_traffic_violation):
+    expunger_result = Expunger.run(record_with_mj_under_21_and_traffic_violation)
+
+    assert isinstance(record_with_mj_under_21_and_traffic_violation.charges[0].charge_type, MarijuanaUnder21)
+    assert isinstance(record_with_mj_under_21_and_traffic_violation.charges[1].charge_type, TrafficViolation)
+
+    assert expunger_result["CASEJD1-1"] == TimeEligibility(status=EligibilityStatus.ELIGIBLE,
+                                                           reason="Eligible now",
+                                                           date_will_be_eligible=date(1999, 3, 3)) #one year after conviction
```
KentShikama, line 275:
```suggestion
                                                           date(1998, 3, 3)+relativedelta(years=1)
```

To be consistent with what you did below

KentShikama: Even better would be to just use `charge.disposition.date + relativedelta(years=1)`

src/backend/tests/test_crawler_expunger.py
```diff
@@ -254,3 +257,72 @@ def test_expunger_for_record_with_mj_over_21(record_with_mj_over_21):
...
+    expunger_result = Expunger.run(record)
+
+    mj_charge = record.charges[0]
+    felony_charge = record.charges[1]
+
+    assert isinstance(mj_charge.charge_type, MarijuanaUnder21)
+
+    assert expunger_result["CASEJD1-2"] == TimeEligibility(status=EligibilityStatus.INELIGIBLE, reason="Never. Type ineligible charges are always time ineligible.",
+                                                           date_will_be_eligible=date.max())
+    assert expunger_result["CASEJD1-1"] == TimeEligibility(status=EligibilityStatus.ELIGIBLE,
```
KentShikama, line 326: Nitpick but I prefer to be consistent about the ordering: assert for CASEJD1-1 before CASEJD1-2.

src/backend/tests/test_crawler_expunger.py
```diff
@@ -254,3 +257,72 @@ def test_expunger_for_record_with_mj_over_21(record_with_mj_over_21):
...
+        record=YoungDoe.SINGLE_CASE_RECORD,
+        cases={"CASEJD1": CaseDetails.CASE_MJ_AND_CIVIL_CONVICTION})
+
+    expunger_result = Expunger.run(record)
+
+    assert isinstance(record.charges[1].charge_type, CivilOffense)
+
+    assert expunger_result["CASEJD1-1"] == TimeEligibility(status=EligibilityStatus.ELIGIBLE,
+                                                           reason="Eligible now",
+                                                           date_will_be_eligible=date(1998, 3, 3)+relativedelta(years=1))
```
KentShikama, line 291: This is interesting. I would expect +3 years just like the test below.

After more digging, it looks like the CivilOffense is being picked up as a dismissal instead of a conviction. Or more specifically, since the parser doesn't recognize the
```
2. Civil Offense
Convicted
```
that you wrote it in the fixture, by omission, it becomes dismissed. In any case this tests the same thing as `test_expunger_for_record_with_mj_under_21_with_non_traffic_charge_not_subject_to_137226` so I would just delete this test (while keeping the naming).

KentShikama: And if you're curious as to why it doesn't pick up, it looks like you missed a semicolon after the `&nbsp`. As you can see, the case parser is fragile but we've received assurances that we can rely on the HTML structure not changing, so I am not concerned.

PR #1402, opened 2020-08-02
#1223: Only create an error message for a missing disposition if the case is closed.

### LINKED ISSUE #1223: Remove redundant error banner (missing disposition) when the case is Open
This is a behavior fix that we made some time ago but the issue re-appeared due to other recent changes:

I'm pretty sure this is undesired behavior, but pinging @hmarcks to verify.

In this example, the same case number appears twice.

repeated_error_banner

The "missing disposition" error should only show if the cause of the error is for something other than the case being open. If the case is open, the disposition is expected to be missing. 

These errors are generated in the backend, so the logic check for this should appear there.

0x6362:
* Only create an error message for a missing disposition if the case is
  closed.

  The "missing disposition" error should only show if the cause of the
  error is for something other than the case being open. If the case is
  open, the disposition is expected to be missing.

* Add `open` method to `Case` class to make certain pieces of logic more
  obvious (removing double negation in `if` statement)

### DIFF
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -276,7 +276,7 @@ class ErrorChecker:
         for charge in charges:
             if charge.charge_type.blocks_other_charges:
                 case_number = f"[{charge.case(cases).summary.case_number}]"
-                if charge.disposition.status == DispositionStatus.UNKNOWN:
+                if charge.disposition.status == DispositionStatus.UNKNOWN and charge.case(cases).summary.closed():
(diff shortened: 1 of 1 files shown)


wittejm:
I’m not sure that adding the open() method improves readability here. Why not just have:

`if charge.disposition.status == DispositionStatus.UNKNOWN and charge.case(cases).summary.closed()`?

0x6362:
Good point! Addressed in 7c5759b

wittejm:
Awesome. Feel free to merge!

Issue #874, opened 2020-02-21
Reference the blocking charge in time eligibility reason if there is one


wittejm:
When a charge is blocked by some other charge, it may be useful to the user, and make results more intuitive, if the time reason provides more than just "blocked by most recent other conviction" and actually names the blocking conviction that's doing the blocking. Or more specifically, it should name the case number which is the unique identifier on the record, rather than just the charge name.   
 
The case number and charge in the blocking conviction/arrest are close at hand, where the time elig reason string is being constructed: 
https://github.com/codeforpdx/recordexpungPDX/blob/b3bd39a5f4b601d9a09bda1d7b97eced006523fc/src/backend/expungeservice/expunger.py#L47

wittejm:
copying @KentShikama  comment on this over from #872 :

> I guess the issue is that we don't really have an ID for charges. The closest identification is something like "2nd charge of case XYZ". Maybe something like

```
eligibility_dates.append(
                    (
                        most_recent_blocking_conviction.disposition.date + relativedelta(years=10),
                        "Time-ineligible under 137.225(7)(b)",
                        f"Ten years from charge in case {most_recent_blocking_conviction.case()().case_number} (most recent {conviction}) (137.225(7)(b))",
                    )
                )
```

KentShikama:
Note if the blocking charge is an open charge then the message should say something like `"Ten years from charge in case  {most_recent_blocking_conviction.case.case_number} (hypothetical most recent {conviction}) (137.225(7)(b))"`.

PR #1403, opened 2020-08-02
#874 Reference Blocking Charge in Eligbility

0x6362:
Reference the blocking charge in time eligibility reason if there is
one.

> When a charge is blocked by some other charge, it may be useful to the
user, and make results more intuitive, if the time reason provides more
than just "blocked by most recent other conviction" and actually names
the blocking conviction that's doing the blocking. Or more specifically,
it should name the case number which is the unique identifier on the
record, rather than just the charge name.

- [ ] Review logic and confirm language is correct in cases where most
  recent blocking charge is:
    - [ ] Open
    - [ ] Closed
- [ ] Update test suite once language and logic is reviewed by team

--- 
After Change:
image

Before Change:
image

### DIFF
--- a/src/backend/expungeservice/expunger.py
+++ b/src/backend/expungeservice/expunger.py
@@ -3,7 +3,7 @@ from dataclasses import replace
 from expungeservice.models.charge_types.unclassified_charge import UnclassifiedCharge
 from expungeservice.util import DateWithFuture as date
 from functools import lru_cache
-from typing import Set, List, Tuple, Dict
+from typing import Set, List, Tuple, Dict, Optional
 
 from dateutil.relativedelta import relativedelta
 from more_itertools import padnone, take
@@ -85,10 +85,16 @@ class Expunger:
 
             if most_recent_blocking_conviction:
                 conviction_string = "other conviction" if charge.convicted() else "conviction"
+                recency = "most recent"
+                summary = most_recent_blocking_conviction.case(cases).summary
+
+                if not summary.closed():
+                    recency += "hypothetical"
+
                 eligibility_dates.append(
                     (
                         most_recent_blocking_conviction.disposition.date + relativedelta(years=10),
-                        f"Ten years from most recent {conviction_string} (137.225(7)(b))",
+                        f"Ten years from charge in case {summary.case_number} ({recency} {conviction_string}) (137.225(7)(b))",
                     )
                 )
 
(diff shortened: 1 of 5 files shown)


wittejm:
I think when the most recent case is closed it can say 

"Ten years from most recent conviction (137.225(b)): case 1000000"

to provide the "legal justification" first and followed by a "hint".

If a case has an open record, I don't think we should change the language at all. My one thought was that it could point to all of the open cases, but I don't think this adds any useful information.

wittejm:
I keep trying to tweak wordings of the new hint language and I don't love any of it:

"Ten years from most recent conviction (137.225(b)): case 1000000"
"Ten years from most recent conviction (137.225(b)) in case 1000000"
"Ten years from most recent conviction (137.225(b)) from case 1000000"
"Ten years from most recent conviction (137.225(b)) of case 1000000"

Changing my mind again, maybe "from" is better than "in"? @hmarcks What do you think?

hmarcks:
I think I prefer the first option with a ":", but I like "from" better than "in".

I also think it makes it easier to understand if the statute is removed. Perhaps it could be in the "More Info"?

KentShikama:
I like "Ten years from most recent conviction (137.225(b)) from case 1234".

KentShikama:
> If a case has an open record, I don't think we should change the language at all. 

We probably should since the open charge is not guaranteed to become a conviction. The word "potential" might be better than "hypothetical". For example, "Ten years from most recent potential conviction (137.225(b)) from open case 1234". I think it is fine to either reference one, all, or none of the open cases as any of them have the possibility to be the latest conviction. A mathematician would say "without loss of generality".

KentShikama:
Will resolve https://github.com/codeforpdx/recordexpungPDX/issues/874

0x6362:
Since there's not yet consensus on the language I'm going to refactor the test suite tonight to be more DRY, and then once the final variant is settled it should be easier to make the changes reflected everywhere.

wittejm:
I realized there would rarely be more than one open case but if there are then I think referencing all of them is better than just one, which would feel weird and potentially confusing. So just having the one linked case in almost all instances does make sense. If there are 2+ open cases, then the benefit of listing all of them is that we’re more explicit and the cost is that there is slightly more clutter to show slightly redundant information.

I like the idea of pulling the statute citation out to the front. Every other time & type reason has the citation in-line above More Info, and I think we should stick with that if we can.  Also "potential" is cool. 

All together (including the "Eligible now OR" that comes in from the merger) it would look like
```
Eligible now 
OR 137.225(7)(b) – Ten years from most recent potential conviction from open cases 1000000, 2000000.
```

KentShikama:
@wittejm Sounds good to me.

0x6362:
Thanks for clarifying. I implemented most of the test improvements and I will hopefully get this change out for review tomorrow.

KentShikama:
@0x6362 Any updates? Do you need help here by any chance?

KentShikama:
@0x6362 I have gone ahead and pushed the language update. @wittejm PTAL

PR #1405, opened 2020-08-04
Enable fillable forms in demo

KentShikama:


0x6362:
This fixes the bug caused by missing Oeci search credentials in Demo mode. 

Tested by:
1. New private window
2. enable demo mode
3. Search for "Single Conviction" user
4. Click "Generate Paperwork"
5. Validated Generated Paperwork was a valid ZIP file and Document had all the proper data (which should not be affected by this change)

LGTM.

PR #1411, opened 2020-08-11
Show loading animation when downloading Expungement Paperwork

### LINKED ISSUE #1313: Add loading spinner for the form filling download as it takes a while for large charges

wittejm:
Unrelatedly, remove some unused imports from the FillForms/index file.

Closes #1313


wittejm:
@hmarcks I used the same effect as the download pdf, but on the blue button it's fairly subtle. I think it's probably fine, but let me know if you think it should be something more noticeable.

hmarcks:
Agree it's fine for now

PR #1448, opened 2020-10-19
Handle negative case balance.

wittejm:
Parse parens in case balance if they appear in OECI and interpret as a negative value.
Show negative balance due in the frontend using parens.

### DIFF
--- a/src/backend/expungeservice/models/case.py
+++ b/src/backend/expungeservice/models/case.py
@@ -105,6 +105,10 @@ class CaseCreator:
 
     @staticmethod
     def compute_balance_due_in_cents(balance_due_dollar_amount: str):
-        is_negative = len(balance_due_dollar_amount) > 0 and balance_due_dollar_amount[0] == "("
-        balance_due_dollar_amount_float = float(re.sub("[,()]", "", balance_due_dollar_amount))
-        return (-1 if is_negative else 1) * int(balance_due_dollar_amount_float * 100)
+        return int(CaseCreator._balance_to_float(balance_due_dollar_amount) * 100)
+
+    @staticmethod
+    def _balance_to_float(balance: str) -> float:
+        commas_removed = balance.replace(",","")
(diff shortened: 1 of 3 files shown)


src/backend/expungeservice/models/case.py
```diff
@@ -97,5 +105,6 @@ def create(
 
     @staticmethod
     def compute_balance_due_in_cents(balance_due_dollar_amount: str):
-        balance_due_dollar_amount_float = float(balance_due_dollar_amount.replace(",", ""))
-        return int(balance_due_dollar_amount_float * 100)
+        is_negative = len(balance_due_dollar_amount) > 0 and balance_due_dollar_amount[0] == "("
+        balance_due_dollar_amount_float = float(re.sub("[,()]", "", balance_due_dollar_amount))
+        return (-1 if is_negative else 1) * int(balance_due_dollar_amount_float * 100)
```
KentShikama, line 110:
```suggestion
        return int(CaseCreator._balance_to_float(balance_due_dollar_amount) * 100)

    @staticmethod
    def _balance_to_float(balance: str) -> float:
        commas_removed = balance.replace(",","")
        normalized_negative = re.sub("\((?P<balance>.*)\)", "-\g<balance>", commas_removed)
        return float(normalized_negative)
```

KentShikama (approved):

src/backend/expungeservice/models/case.py
KentShikama: This feels more "proper" than `balance_due_dollar_amount[0] == "("` but it might be harder to read as no where else in the codebase do we use match groups in regex. Feel free to ignore.

wittejm: Mmmmm. How does one learn this black magic?

wittejm: Though it does feel cleaner and not crazy to figure out what's going on.

KentShikama (approved):

Issue #1456, opened 2020-11-10
Add note of (in part) to all partial cases in PDF


KentShikama:
### This task is done when:
 -
 -

wittejm:
Completed with #1459

PR #1459, opened 2020-11-10
Always use case_number_with_comments for case_number

KentShikama:
This ensures that the case number field in the expungement PDF forms always have `(in part)` for partial cases regardless of whether each charge is individually listed.


wittejm (approved):

KentShikama:
Resolves https://github.com/codeforpdx/recordexpungPDX/issues/1456

PR #1462, opened 2020-11-11
Bugfix: Update the 137.225(6)(f) type listed in Edit Charge 

wittejm:
names listed now match the charge name in backend.

I think I renamed these types to "Young Offender Sex Crime" some time ago but didn't sync their names in the frontend.


wittejm:
Giving these charge types two different readable names `type_name` in the backend was required for making them both selectable in the list. 

For some reason, when the two charge types had the same `type_name` then the charge classifier doesn't construct an ambiguous type. Now that the names are different (this is the only change that could have caused this, though it does not make sense to me), the classifier creates the OR ambiguous type which is the desired behavior anyway. 

It's tough to spot because we don't really test the classifier against real OECI data unless it shows up in regular usage, and the R&J type never does.

KentShikama (approved):

Issue #1457, opened 2020-11-10
Add filling fee total to summary panel


KentShikama:
Add filing fees and update balance section so they are both organized by county since that is how they are paid.

![filing-fees-and-balance](https://user-images.githubusercontent.com/6887982/98873722-80a80780-243e-11eb-987c-84bc02337483.png)

[filing-fees-and-balance.html.zip](https://github.com/codeforpdx/recordexpungPDX/files/5527218/filing-fees-and-balance.html.zip)

hmarcks:
@wittejm @KentShikama I added the mockup and attached some markup, thanks

wittejm:
Awesome.

PR #1464, opened 2020-11-16
Fines and fees

wittejm:
- Add filing fees to record summary panel
- Also reorg fines to group by county
- Some additional data is provided by the backend


wittejm:
Issue #1457

Only thing left todo is the extra note for why filing fees are "None" .

src/backend/expungeservice/record_summarizer.py
```diff
@@ -13,27 +15,30 @@ class RecordSummarizer:
...
         eligible_charges_by_date = RecordSummarizer._build_eligible_charges_by_date(record)
+        eligible_nonconvictions_only = RecordSummarizer._build_eligible_nonconvictions_only(record.charges)
         return RecordSummary(
             record=record,
             questions=questions,
             eligible_charges_by_date=eligible_charges_by_date,
             total_charges=len(record.charges),
             county_fines=county_fines,
             county_filing_fees=county_filing_fees,
+            eligible_nonconvictions_only=eligible_nonconvictions_only,
```
wittejm, line 26: I feel sad about this variable name because to be fully accurate it needs to be true when the number of eligible convictions is zero. A more accurate name is `has_eligible_nonconvictions_and_no_eligible_convictions`. But that is probably too silly.

src/backend/expungeservice/record_summarizer.py
```diff
@@ -86,3 +91,20 @@ def get_label(charge: Charge):
...
             eligible_charges_by_date[label] = charges_tuples
         return eligible_charges_by_date
+
+    @staticmethod
+    def _build_eligible_nonconvictions_only(charges_eligible_now):
+        return (
+            charges_eligible_now
+            and len(
+                list(
+                    filter(
```
wittejm, line 101: Why doesn't a generator have a builtin `next` alternate that returns None instead of an exception? Blar.

src/frontend/src/components/RecordSearch/Record/RecordSummary/CountyFilingFees.tsx
```diff
@@ -0,0 +1,69 @@
...
+          </a>
+        </h3>
+        {this.props.county_filing_fees.map(
+          (case_filing_fee: CountyFilingFeeData) => {
+            return (
+              <>
+                <h4 className="fw6 mb2">{case_filing_fee.county_name}</h4>
+                <ul className="mw5 list mb3">
+                  <li className="mb2">
+                    <span className="link hover-blue">{`Case Conviction x${case_filing_fee.cases_with_eligible_convictions}`}</span>{" "}
```
hmarcks, line 35: I accidentally put hover style classes here, should be removed

src/frontend/src/components/RecordSearch/Record/RecordSummary/CountyFilingFees.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    <span className="fr">
+                      {currencyFormat(
+                        perCaseFilingFee *
+                          case_filing_fee.cases_with_eligible_convictions
+                      )}
+                    </span>
+                  </li>
+                  <li className="mb2">
+                    <span className="link hover-blue">Fingerprint Fee</span>{" "}
```
hmarcks, line 44: I accidentally put hover style classes here, should be removed here too

hmarcks:
Functionality and styles look good to me (besides the css mistakes I made)! One thing I noticed is that the search state is lost if you navigate away then back (for example if you run a search then click one of the new ? icon links). Not sure if that happened in this PR or earlier though.

src/backend/expungeservice/record_summarizer.py
KentShikama:
```
g = iter([0,1])
next(g, None)
```

wittejm: Oh right, We've done this before. I was looking around SO and saw many answers of "yep can't do that" for some reason.

wittejm:
Thanks for catching this @hmarcks, I noticed the losing-state issue then forgot to actually fix it. 

Btw @KentShikama I think we usually kind of disregard the Codacy notes anyway, but the things it suggests here are all bad suggestions.

src/frontend/src/components/RecordSearch/Record/RecordSummary/CountyFilingFees.tsx
```diff
@@ -0,0 +1,70 @@
...
+          return county.cases_with_eligible_convictions > 0;
+        }).length > 0 ? (
+          <div className="mw5 bt b--light-gray pt2">
+            <span className="fw6">Total</span>
+            <span className="fr">
+              {currencyFormat(this.props.total_filing_fees_due)}
+            </span>
+          </div>
+        ) : this.props.eligible_nonconvictions_only ? (
+          <span>None (no convctions)</span>
```
KentShikama, line 63:
```suggestion
          <span>None (no convictions)</span>
```

src/backend/expungeservice/record_summarizer.py
```diff
@@ -13,27 +15,30 @@ class RecordSummarizer:
     def summarize(record, questions: Dict[str, QuestionSummary]) -> RecordSummary:
         county_fines, county_filing_fees = RecordSummarizer._build_county_balances(record)
         eligible_charges_by_date = RecordSummarizer._build_eligible_charges_by_date(record)
+        eligible_nonconvictions_only = RecordSummarizer._build_eligible_nonconvictions_only(record.charges)
```
KentShikama, line 18: We need to pass in eligible charges here right?

src/backend/expungeservice/record_summarizer.py
KentShikama: What about `is_all_cases_feeless`

src/backend/expungeservice/record_summarizer.py
```diff
@@ -86,3 +91,15 @@ def get_label(charge: Charge):
             ]
             eligible_charges_by_date[label] = charges_tuples
         return eligible_charges_by_date
+
+    @staticmethod
+    def _build_eligible_nonconvictions_only(charges_eligible_now):
+        return charges_eligible_now and next(
+            filter(
+                lambda charge: charge.expungement_result.charge_eligibility.status
+                == ChargeEligibilityStatus.ELIGIBLE_NOW
```
KentShikama, line 100: `charges_eligible_now`: once this becomes true you do not need to check for `ELIGIBLE_NOW`

wittejm: Woops, that variable name was inaccurate and based on earlier attempted logic. Now it's just all charges on the record.

src/backend/expungeservice/record_summarizer.py
```diff
@@ -86,3 +91,18 @@ def get_label(charge: Charge):
             ]
             eligible_charges_by_date[label] = charges_tuples
         return eligible_charges_by_date
+
+    @staticmethod
+    def _build_is_all_cases_feeless(charges):
+        return bool(
+            charges
```
KentShikama, line 98: Should this be charges or eligible charges?

wittejm: The method checks if there are eligible charges without fees, which is the circumstance that the "no convictions" message is shown in the fees section.

wittejm:
@hmarcks , Kent pointed out the use case where there is an eligible dismissal and a future-eligible conviction, which deserves its own reason for having no fees, so now the three possible reasons are:
* **no convictions eligible now** (but there are future eligible convictions or dismissals)
* **no eligible convictions** (but there are eligible dismissals)
* **no eligible cases** (nothing is or will become eligible)

hmarcks:
👍

src/backend/expungeservice/record_summarizer.py
```diff
@@ -93,16 +93,29 @@ def get_label(charge: Charge):
...
+            None,
+        )
+        if has_future_eligible_convictions:
+            reason = "no convictions eligible now"
+        else:
+            if has_eligible_nonconvictions:
+                reason = "no eligible convictions"
+            else:
+                reason = "no eligible cases"
+        return reason
```
KentShikama: I like the switch to reason strings; it makes it much clearer at glance what this will be used for in the front-end compared to `is_all_cases_feeless`.

In general, [list comp should be used over filter](https://www.artima.com/weblogs/viewpost.jsp?thread=98196) and I think that applies here too (especially given the rest of the codebase using list comps). The following is not tested.
```python3
reason = "no eligible cases" # reason only applies if feeless
if charges:
    charges_will_be_eligible = [c for c in charges if c.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.WILL_BE_ELIGIBLE]
    nonconvictions_eligible_now = [c for c in charges if c.expungement_result.charge_eligibility.status == ChargeEligibilityStatus.ELIGIBLE_NOW and c.disposition.status != DispositionStatus.CONVICTED]
    if charges_will_be_eligible:
        reason = "no convictions eligible now"
    elif nonconvictions_eligible_now:
        reason = "no eligible convictions"
return reason
```

KentShikama (approved):

PR #1528, opened 2021-07-03
Allow expungement date to be shifted to the future

KentShikama:
Sometimes we want to run our Expunger as if we were running it tomorrow or some arbitrary date in the future. We therefore add a new field internally called "today" to the SearchPanel to be able to change what day "today" is. By default it is the date as of now in PST/PDT. This is a really sweet feature addition from a technical perspective since we only rely on date.today() in one location in expunger.py, so we can do this change in relatively few lines and safely.

Here is a screenshot of the frontend changes:

Screen Shot 2021-07-03 at 1 02 31 PM

### DIFF
--- a/src/backend/expungeservice/endpoints/search.py
+++ b/src/backend/expungeservice/endpoints/search.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from dacite import from_dict
 from expungeservice.models.record_summary import RecordSummary
 from flask.views import MethodView
@@ -10,7 +11,7 @@ from expungeservice.request import error
 from expungeservice.serializer import ExpungeModelEncoder
 from expungeservice.crypto import DataCipher
 from expungeservice.record_summarizer import RecordSummarizer
-from expungeservice.util import LRUCache
+from expungeservice.util import LRUCache, DateWithFuture as date_class
 
 
 class Search(MethodView):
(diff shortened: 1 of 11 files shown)


michaelzhang43:
What an elegant solution!

michaelzhang43 (approved):
This sounds like a really elegant way to do it. Much better than the 'print all future eligible cases' idea I had. I especially like that this is the 'general solution' to my problem of being able to generate paperwork for one type of date (arbitrarily far in the future). Great idea

src/backend/expungeservice/endpoints/search.py
```diff
@@ -1,3 +1,5 @@
+from datetime import datetime
+
```
wittejm, line 2: space.

src/backend/expungeservice/endpoints/search.py
```diff
@@ -10,7 +12,7 @@
 from expungeservice.serializer import ExpungeModelEncoder
 from expungeservice.crypto import DataCipher
 from expungeservice.record_summarizer import RecordSummarizer
-from expungeservice.util import LRUCache
+from expungeservice.util import LRUCache, DateWithFuture as date_class
```
wittejm, line 15: I don't think this needs an alias.

wittejm (approved):
Two nitpicks, otherwise LGTM

src/backend/expungeservice/endpoints/search.py
KentShikama: We're pretending like DateWithFuture is the standard `date` library, which we've called `date_class` across most of our codebase for legacy reasons. I'm not sure if I like this decision in retrospect but I'll stick with this for now until we feel like we should change it everywhere to not have an alias.

michaelzhang43:
Bug: When I run it in dev, I can't generate paperwork if I change the expunge date. Had an intern try it and got the same result. The "Generate Paperwork" button still works, but does not generate any records

KentShikama:
Ah yes, it is still generating paperwork as if "today" was now in PST. I've put a update here: https://github.com/codeforpdx/recordexpungPDX/pull/1529.

NickSchimek:
This is sweet!

PR #1533, opened 2021-07-19
Expunger law overhaul

wittejm:
Need to figure out branch control scheme before merging to master.


wittejm:
Still Todo:
- Probation revoked to 3 years. I didn't do this yet because I had trouble updating the test to match.
- Make the editor a bit smarter. Right now you can select charge type Felony Class A and level as Misdemeanor Class A, etc. This works, but is confusing and could be re-imagined. Solutions could be to restrict level options given the type, or add an explanatory note e.g. in a modal.

src/backend/expungeservice/expunger.py
```diff
@@ -39,10 +39,9 @@ def run(record: Record, today: date = date.today()) -> Dict[str, TimeEligibility
 
             other_blocking_charges = [c for c in other_charges if c.charge_type.blocks_other_charges]
 
-            _, convictions = Case.categorize_charges(other_charges)
-            blocking_dismissals, blocking_convictions = Case.categorize_charges(other_blocking_charges)
+            convictions = [c for c in other_charges if c.convicted()]
```
KentShikama, line 42: I know it was like this before but we should call this "other_convictions"

src/backend/expungeservice/expunger.py
```diff
@@ -131,6 +112,45 @@ def run(record: Record, today: date = date.today()) -> Dict[str, TimeEligibility
...
+            return (
+                relative_charge_date + relativedelta(years=3),
+                "Three years from date of conviction (137.225(1)(a))",
+            )
+        elif any([x in target_charge_level for x in ["Misdemeanor Class B", "Misdemeanor Class C", "Violation"]]):
+            return (
+                relative_charge_date + relativedelta(years=1),
+                "One year from date of conviction (137.225(1)(a))",
+            )
+        elif "Misdemeanor" in target_charge_level: # TODO: Is an unspecified Misdemeanor always a Class A?
```
KentShikama: @michaelzhang43 Is an unspecified Misdemeanor always a Class A?

src/backend/expungeservice/expunger.py
```diff
@@ -83,15 +83,10 @@ def run(record: Record, today: date = date.today()) -> Dict[str, TimeEligibility
...
-                potential = "potential " if not summary.closed() else ""
-                eligibility_dates.append(
-                    (
-                        most_recent_blocking_conviction.disposition.date + relativedelta(years=10),
-                        f"137.225(7)(b) – Ten years from most recent {potential}{conviction_string} from case [{summary.case_number}].",
-                    )
-                )
+                relative_case_summary = most_recent_blocking_conviction.case(cases).summary
+                blocking_convictions_time_eligibility = Expunger._other_blocking_conviction_years_by_level(charge.level, most_recent_blocking_conviction.disposition.date, relative_case_summary)
+                if blocking_convictions_time_eligibility:
```
KentShikama, line 88: When will `blocking_convictions_time_eligibility` resolve to `False`/`None` here?

src/backend/expungeservice/models/charge_types/misdemeanor_class_bc.py
```diff
@@ -0,0 +1,24 @@
...
+
+from expungeservice.models.charge import ChargeType
+from expungeservice.models.charge import ChargeUtil
+from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+
+
+@dataclass(frozen=True)
+class MisdemeanorClassBC(ChargeType):
+    type_name: str = "Misdemeanor Class B or C"
+    expungement_rules: str = """Convictions for misdemeanors are generally eligible under ORS 137.225(5)(b).
```
KentShikama, line 11: Did you check if "ORS 137.225(5)(b)" is the right category still?

src/backend/expungeservice/charge_classifier.py
```diff
@@ -316,6 +318,11 @@ def _traffic_crime(statute, name, level, disposition):
         if "infraction" in level:
             return AmbiguousChargeTypeWithQuestion([TrafficViolation()])
 
+    @staticmethod
+    def _contempt_of_court(name):
+        if "contempt of court" in name:
```
KentShikama, line 323: @michaelzhang43 Is it sometimes just "contempt" or is it always fully written out in OECI as "contempt of court"? Or to put it differently, is there any way that "contempt" can be in the name without it being a "contempt of court"?

src/backend/expungeservice/models/charge_types/contempt_of_court.py
```diff
@@ -0,0 +1,18 @@
...
+
+from expungeservice.models.charge import ChargeType
+from expungeservice.models.expungement_result import TypeEligibility, EligibilityStatus
+
+
+@dataclass(frozen=True)
+class ContemptOfCourt(ChargeType):
+    type_name: str = "Contempt of Court"
+    expungement_rules: Any = (
+        """The statute was updated as of Jan 1, 2022 to name Contempt of court as eligible under subsection 137.225(5)(e)."""
```
KentShikama, line 12: Capitalization of "Contempt of court"

src/backend/expungeservice/record_editor.py
```diff
@@ -102,7 +103,7 @@ def _add_charge(case_number, ambiguous_charge_id, edit) -> Charge:
             "case_number": case_number,
             "id": f"{ambiguous_charge_id}-0",
             "statute": "",
-            "level": "",
```
KentShikama, line 105: What happens if you just remove `"level": "",` instead of doing the change you made here? Would "level" then not be defined sometimes? In which case what would it default to?

KentShikama:
Btw I love the commit by commit. For a PR with a total of a 1000 line diff, I found it really easy to follow.

src/backend/expungeservice/expunger.py
michaelzhang43: Let's assume that it's Class A for practical purposes

src/backend/expungeservice/charge_classifier.py
michaelzhang43: It's usually Contempt of Court, but I strongly believe there are no other criminal or civil charges with the word "Contempt", so Ok to assume that as well.

src/backend/expungeservice/expunger.py
michaelzhang43: Not to be confused with the common situation in which a C Misdemeanor is listed as "C/Unclassified", in which case it should be considered a C Misdemeanor

src/backend/expungeservice/models/charge_types/misdemeanor_class_bc.py
wittejm: I hadn't checked, but yes this is unchanged.

src/backend/expungeservice/expunger.py
wittejm: @michaelzhang43 that sounds surprising. Just to check, you mean the level would say "C/Unclassified" or would it say "Misdemeanor Class C/Unclassified"? (I mean, if it says "C/Unclassified" that just sounds awful if it's not specifying misdemeanor vs felony)

src/backend/expungeservice/expunger.py
wittejm: Ah, it's always non-empty because of the check above on `most_recent_blocking_conviction`, so I'll remove the conditional here.

src/backend/expungeservice/record_editor.py
wittejm: Ah yes, it's required in the frontend like everything else so I've removed the default here and added a value to all the unit tests.

wittejm:
My thoughts on publishing these changes: 

We should ideally have both the current version and this new version ("3.0") both live until Jan 1st. Merging this to master does that automatically by publishing to dev.recordsponge.com. But what if we want to push independent changes to Prod without including Version 3? E.g. if we want to add another Partner to the front page. We could have a feature branch that is just Version 2.5 with whatever new commits, and we tag that with Version 2.5.2(,3,4...) -- and, tagging that commit will trigger the publish to recordsponge.com.  Is that right?

My only question here is: does the `deploy to prod` Github action work on any branch and not just master? Reading through the yml and makefile it looks like `master` is never referred to and it might just work on any new version tag.... yes? In which case everything is awesome and we can safely maintain Version 2 and Version 3 to be published simultaneously.

wittejm:
Ready for another review @KentShikama

KentShikama (approved):
LGTM!

KentShikama:
> But what if we want to push independent changes to Prod without including Version 3?

We can think about it when that need comes to be. It seems unlikely we'll need to change anything with only Verison 2.

PR #1537, opened 2021-07-29
Auto-add slashes to DOB field and disable inputs that don't look like dates

wittejm:
Sarah Kolb asked for this feature. 

I've seen web forms out there that have nice text styling too that makes this feel more intuitive, like with a light-colored suggestion text where the slash "solidifies" at the right time. Other solutions have 3 fields and the cursor moves between them as the user types. Those are probably harder to implement than the solution I did here.

@KentShikama @hmarcks what do you think of putting this change in?

### DIFF
--- a/src/frontend/src/components/RecordSearch/SearchPanel/Alias.tsx
+++ b/src/frontend/src/components/RecordSearch/SearchPanel/Alias.tsx
@@ -69,7 +69,19 @@ export default class Alias extends React.Component<Props, State> {
           content={this.props.aliasData.birth_date}
           coda="mm/dd/yyyy"
           onChange={(fieldValue: string) => {
-            this.handleFieldChange("birth_date", fieldValue);
+            let newValue = fieldValue;
+            const match1 = new RegExp("^[0-9][0-9](\/[0-9][0-9])?$").exec(fieldValue);
+            const match2 = new RegExp("^[0-9]?([0-9](\/([0-9]([0-9](\/([0-9]([0-9]([0-9][0-9]?)?)?)?)?)?)?)?)?$").exec(fieldValue);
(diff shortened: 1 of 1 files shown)


wittejm:
A bit more of what this does currently so you don't have to pull changes to try it out if you don't want to:
- After entering two digits, a slash appears. It doesn't check whether those digits are in 1-12.
- If you try to type a slash yourself at that point, nothing happens.
- If you backspace to delete the slash, it gets deleted as expected.
- Same thing after you've typed a month, slash, and then two more digits.
- Ignoring invalid inputs works by checking if the complete input so far is a "work in progress date string" and ignoring the change from any inputs that produce any other string. For example, letters and special characters don't work, and trying to type two slashes doesn't work.

hmarcks:
I'm running your branch locally but can't seem to get it working - after I type two numerals, a slash doesn't appear. Perhaps I need to do something else to enable it?

KentShikama:
> I've seen web forms out there that have nice text styling too that makes this feel more intuitive, like with a light-colored suggestion text where the slash "solidifies" at the right time.

Why not just use a date picker library for React?

wittejm:
Why not? No particular reason except I wanted to try this approach. But this does feel pretty clunky and finding a suitable version out there sounds good.

hmarcks:
I think most date pickers are inaccessible. [Here's an article about that with a potential example we could use](https://adrianroselli.com/2019/07/maybe-you-dont-need-a-date-picker.html). I don't think a picker is that important since the date should be known.

wittejm:
Thanks for that reference @hmarcks . I don't know  why it wouldn't work for you locally; I think I'm going to ditch this feature as unimportant.

PR #1547, opened 2021-08-05
MJ Manu/Del 'Felony Unclassified' is a Felony Class C

wittejm:
Revert most of the change that adds a follow-up question; instead just use this `MarijuanaManufactureDelivery` for the "felony unclassified" case, following Michael's new info.


KentShikama (approved):
Glad to see it simplified!

PR #1550, opened 2021-09-01
Prob revoked 3 years

wittejm:


src/backend/tests/fixtures/case_html/with_revoked_probation.html
```diff
@@ -372,26 +372,26 @@
...
-            Created: 11/09/2010 12:00 AM</div></td></tr><tr><th class="ssTableHeaderLabel" valign="top" id="RCDER73">11/09/2010</th><td style="border-right: 1px solid black">&nbsp;</td><td>&nbsp;</td><td headers="COtherEventsAndHearings RCDER73"><b>Sentence</b><div style="padding-left: 10px;font-size:8pt;">
-            Created: 11/18/2010 12:00 AM</div></td></tr><tr><th class="ssTableHeaderLabel" valign="top" id="RCDER74">11/09/2010</th><td style="border-right: 1px solid black">&nbsp;</td><td>&nbsp;</td><td headers="COtherEventsAndHearings RCDER74"><b>Judgment - Sentence</b>
+            Created: 11/09/2010 12:00 AM</div></td></tr><tr><th class="ssTableHeaderLabel" valign="top" id="RCDER72">11/09/2017</th><td style="border-right: 1px solid black">&nbsp;</td><td>&nbsp;</td><td headers="COtherEventsAndHearings RCDER72"><b>Notice - Advise Appeal Rights</b><div style="padding-left: 10px;font-size:8pt;">
```
KentShikama, line 375: I find it odd that you're modifying the test source instead of the python tests as the source OECI should never change in theory. But I think this has the same effect so 🤷

KentShikama (approved):

src/backend/tests/fixtures/case_html/with_revoked_probation.html
wittejm: I know. That was my first idea but, because of other time restrictions on the record the probation revoked date no longer was the limiting date on eligibility. So some edits to the source data would be necessary regardless.

PR #1552, opened 2021-09-18
Update Summary PDFs to contain balance due

KentShikama:
Based on feedback from @michaelzhang43.

We want to mimic the design of the HTML summary panel as much as possible for the PDF. This includes printing out information about the balances due for each case.

Also suppresses the balance due/case grouping if the charges are Eligible now, Ineligible, etc.

Screenshots examples

Screen Shot 2021-09-18 at 4 09 35 AM
Screen Shot 2021-09-18 at 4 10 31 AM


michaelzhang43:
Looks fantastic!

michaelzhang43:
I really like the summary panel design as it is, so hewing close to that design in the summary pdf, as here, would be ideal

wittejm (approved):

PR #1557, opened 2021-11-11
Dismissed traffic violations are ineligible.

wittejm:
* This is at Michael's request because he expects courts to reject petitions for these charges.

### DIFF
--- a/src/backend/expungeservice/models/charge_types/parking_ticket.py
+++ b/src/backend/expungeservice/models/charge_types/parking_ticket.py
@@ -23,4 +23,4 @@ class ParkingTicket(ChargeType):
             )
 
     def hidden_in_record_summary(self, disposition):
-        return ChargeUtil.convicted(disposition)
+        return True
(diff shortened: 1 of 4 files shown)


KentShikama (commented):
Should we not just exclude these charges from being listed?

wittejm:
Ah yep, done. You didn't mean to suggest we take out their detail entries as well, did you?

KentShikama (commented):
Oh that is what I was thinking.

KentShikama (commented):
This is a bit sad. I think the proper solution is to have a third state of "N/A" that shows up Grey but that might be a lot of work. Let's just re-add it to the summary list and keep it as you had it originally.

src/backend/expungeservice/models/charge_types/parking_ticket.py
```diff
@@ -18,8 +18,9 @@ def type_eligibility(self, disposition):
             return TypeEligibility(EligibilityStatus.INELIGIBLE, reason="Ineligible under 137.225(7)(a)")
         elif ChargeUtil.dismissed(disposition):
             return TypeEligibility(
-                EligibilityStatus.ELIGIBLE, reason="Dismissed violations are eligible under 137.225(1)(b)."
+                EligibilityStatus.INELIGIBLE,
+                reason="Dismissed parking violations are effectively ineligible at the discretion of the court.",
```
KentShikama, line 22: I think we can just expand a bit here instead. Basically say everything that @michaelzhang43 mentioned. "They are eligible but because courts have been rejecting these for the past years, it will be a hassle to pass...we want you to have a smooth experience...etc"

PR #1559, opened 2021-11-14
Updated rules string to for dismissed traffic violations

wittejm:
screenshot to demonstrate formatting: 
dismissed_traffic_rules_example

### DIFF
--- a/src/backend/expungeservice/models/charge_types/parking_ticket.py
+++ b/src/backend/expungeservice/models/charge_types/parking_ticket.py
@@ -27,4 +27,4 @@ As such, petitioners filing to expunge dismissed traffic tickets may receive pus
             )
 
     def hidden_in_record_summary(self, disposition):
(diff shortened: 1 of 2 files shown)


KentShikama (approved):

src/backend/expungeservice/models/charge_types/parking_ticket.py
```diff
@@ -23,4 +27,4 @@ def type_eligibility(self, disposition):
             )
 
     def hidden_in_record_summary(self, disposition):
-        return ChargeUtil.convicted(disposition)
+        return True
```
KentShikama, line 30: Oh do we want to do this for traffic violations?

wittejm: We do not. Thanks. Some sort of copy-paste error...

PR #1594, opened 2022-02-06
Parse SID and pass into PDF generation

KentShikama:


src/backend/expungeservice/crawler/parsers/case_parser.py
```diff
@@ -39,15 +41,37 @@ def feed(data) -> CaseParserData:
...
+            return ""
+
+    @staticmethod
+    def __attempt_parse_sid(soup) -> str:
+        try:
+            party_information = soup.find("div", class_=SECTION_TITLE_CLASS, string="Party Information")
+            for information_section in party_information.parent.next_siblings:
+                sid_text = information_section.find(text=re.compile("SID"))
+                if sid_text:
+                    return sid_text.parent.next_sibling
```
KentShikama, line 70: I would love to find a better way to reliably parse out the SID. It seems to work for all the test cases. We'll see how it works in the wild.

PR #1617, opened 2022-06-11
Update "About Us" for Kent Shikama

KentShikama:
The line "currently volunteers full-time at Code for PDX" is no longer true. I decided to remove the "description" part all together as I don't want to have to maintain it for now. The reformatting was automatic: I have touched no one else's bio.

### DIFF
--- a/src/frontend/src/components/About/index.tsx
+++ b/src/frontend/src/components/About/index.tsx
@@ -61,24 +61,33 @@ export default class About extends React.Component {
             <h2 className="f3 f2-ns fw9 mb5">Our Team</h2>
             <div className="flex flex-wrap justify-between">
               <div className="mw5 mb5 mr3">
-                <img className="img w5 mb3" src="/img/mz-2x.jpg" alt="Photo of Michael Zhang" />
+                <img
+                  className="img w5 mb3"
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/About/index.tsx
```diff
@@ -94,48 +103,57 @@ export default class About extends React.Component {
...
-                  <dd className="lh-copy">
-                    Hunter studied graphic design at the University of Minnesota.
-                    He joined Code for PDX in 2018 to contribute to civic
-                    technology and has primarily worked on RecordSponge. He has
-                    been working on digital products for over ten years and
-                    currently designs software at Higg.
-                  </dd>
+                  <dt className="fw8 mb2">Kent Shikama</dt>
+                  <dt className="fw6 mb3">Developer</dt>
+                  <dd className="lh-copy"></dd>
```
hmarcks, line 156: Maybe delete this empty `<dd>`?

hmarcks (approved):

src/frontend/src/components/About/index.tsx
KentShikama: Yup makes sense!

PR #1625, opened 2022-09-22
Requested changes: form filenames

wittejm:


src/backend/expungeservice/form_filling.py
```diff
@@ -364,3 +365,14 @@ def _build_pdf_path(location: str, convictions: List[Charge]) -> str:
                 return path.join(Path(__file__).parent, "files", "oregon_with_arrest_order.pdf")
```
KentShikama, line 365: What's the difference here?

wittejm: What do you mean? The difference between `_build_pdf_path` and `_build_base_file_name`? This is to decouple internal source file names from the generated file names. But I'm not sure what you're asking about.

KentShikama: Why do they need to be different?

KentShikama: Oh I see. We are renaming the files that are output.

KentShikama (approved):

Issue #1634, opened 2022-12-20
Upgrade frontend npm dependencies


monocle:
Per the Slack discussion, I'll start working on this.

- [x] switch from node-sass (deprecated) to dart-sass (current primary implementation)
- [ ] upgrade all dependencies to current
- [x] switch out of moment.js per [their recommendation](https://momentjs.com/docs/#/-project-status/)

Notes:
[Reach UI is not currrently maintained](https://github.com/reach/reach-ui/issues/972). This is blocking from updating React to 18x. The Reach UI team mentions alternative:

- [Radix UI](https://www.radix-ui.com/)
- [Ariakit](https://github.com/ariakit/ariakit)
- [Reac Aria](https://react-spectrum.adobe.com/react-aria/)

It would probably be better for long term maintenance just to use plain HTML/CSS.

KentShikama:
@monocle Nice! Feel free to submit PRs one bullet at a time (or even in more granular chunks). My guess is swapping moment.js to dayjs would be a nice chunk.

monocle:
In trying to upgrade the dependencies, I've run into a lot of little fixes. [Here is my fork](https://github.com/monocle/recordexpungPDX/commits/update-frontend-deps) with the commits so far. The commits are pretty atomic. I could rebase if that would be better.

Questions:

- I'm using Prettier through VS Code and some of the commits have whitespace changes due to that. Is there some configuration that I should set on my local machine to avoid those whitespace changes?
- The changes are building up and other things still need to be upgraded. Should I just keep going in the same branch? Or should we try to merge the other non-upgrade related commits separately?
- I've noticed some [minor content issues](https://github.com/monocle/recordexpungPDX/issues/3) that I have not fixed yet. Should we fix those and move them into production before the other things?
-  To avoid spamming you and the Slack channel, I'm planning to tee up things, including issues, in my fork. Let me know if you'd like to be more in the loop with anything and, if so, what the prefer communication method would be.

KentShikama:
I realized I missed this post. There should be a husky hook in this repository that you can run or make into a git commit hook. In general, smaller chunks are preferable at first. Feel free to reach out wherever. Tagging me will make it easier for me to follow @monocle.

monocle:
[...]

PR #1645, opened 2023-01-15
Initial dependency upgrades and associated fixes

monocle:
@KentShikama, this is the beginning of #1634. The first batch is focused on getting to a baseline before future changes.

## Commit 1
* The first package upgrade is changing node-sass to sass. This is what's currently recommended by Sass and also avoids potential compilations issues during installation.
* Prettier was only upgraded to the version that's in the `.pre-commit-config.yaml`. We could upgrade that in a future commit, if you think it would be okay.
* I added snapshot tests for the components that have routes to help with future refactoring.

## Commits 2 and 3
* Upgraded TypeScript and react-scripts.
* There were a lot of TypeScript, ESLint and React warnings. So there are a lot of changes, but they are small and should not affect the app's behavior.

## Commit 4
There were some problems with using newer versions of TS and react-scripts with the older node version we had. Upgrading to `node:18.13-bullseye` fixed those.

With these commits, the frontend dev environment should be generally good to go. However, on my Macbook, I've been having problems with the `docker-compose.yml` binding to port 5000. I've been using `"5001:5000"` port binding instead. If you don't think that port binding would cause other problems (I'm still having problems getting the backend container to work), then I could add that change to this PR. Please let me know.

Thanks.


src/frontend/src/components/About/index.tsx
```diff
@@ -148,7 +148,7 @@ export default class About extends React.Component {
                 <img
                   className="img w5 mb3"
                   src="/img/ks-2x.jpg"
-                  alt="Photo of Kent Shikama"
+                  alt="Kent Shikama"
                 />
                 <dl>
                   <dt className="fw8 mb2">Kent Shikama</dt>
```
KentShikama, line 154: Feel free to add your name btw

KentShikama (approved):
Looks safe to me! Let's push this to dev and test a bit.

PR #1657, opened 2023-01-22
Add logout

trezp:


KentShikama (commented):
@trezp Neat! I'll take a closer look tomorrow.

@hmarcks I feel like the logout button belongs more in the header-ish area but what do you think?

KentShikama:
One question I have right now is how do I logout from the form generation page?

trezp:
@KentShikama I didn't know there was another page. 😬 I'll set this PR back to draft take another look. 

I also thought it should go in the header, but it's a little more complex to conditionally display a log out button, since we're not keeping track of logged in/out state anywhere.  If you all have feedback/suggestions on how this should be approached I'm super open to hearing it.

monocle:
@trezp I think just deleting the cookie on the browser side (not sure that a server request is needed?) and then redirecting would be good since RecordSearch always checks for the oeci_token:

https://github.com/codeforpdx/recordexpungPDX/blob/77cccb450e7ce11b80922c950ea441846891c127/src/frontend/src/components/RecordSearch/index.tsx#L17-L22

Besides deleting the cookie, though, the redux store should get cleared out because someone could log out, but then another person could come and visit the demo and see the results of the last search by the previously logged in person.

trezp:
Yeah the server request isn’t necessary for thud particular thing, but there’s a Todo in the login endpoint to create a logout endpoint for something else, so I thought might as well 🤷🏻‍♀️I’ll look into clearing the redux store, thanks On Jan 22, 2023, at 6:46 PM, monocle ***@***.***> wrote:﻿
@trezp I think just deleting the cookie on the browser side (not sure that a server request is needed?) and then redirecting would be good since RecordSearch always checks for the oeci_token:
https://github.com/codeforpdx/recordexpungPDX/blob/77cccb450e7ce11b80922c950ea441846891c127/src/frontend/src/components/RecordSearch/index.tsx#L17-L22
Besides deleting the cookie, though, the redux store should get cleared out because someone could log out, but then another person could come and visit the demo and see the results of the last search by the previously logged in person.

hmarcks:
> @hmarcks I feel like the logout button belongs more in the header-ish area but what do you think?

If it can work, yeah I think it ideally goes to the right of Search, thanks

KentShikama:
>  If you all have feedback/suggestions on how this should be approached I'm super open to hearing it.
> Should there be a dedicated log in button as well? Basically toggle between the two?

I don't really have a good answer off the top of my head but let me just say that we have an less common system where we aren't logging in to our own site but the OECI site. In short, I'm not sure if it makes sense to have a "Login" button explicitly as that might give the impression that a user can login to our site. The "Logout" button feels better to me as by the time the user has logged in, they know that they've logged in to "OECI" and the only logical extension is that the "Logout" is for that. But maybe I'm overthinking it.

hmarcks:
> Should there be a dedicated log in button as well? Basically toggle between the two?

Since the Search button in the header brings you to the OECI login view, I don't think we need one.

trezp:
Closing as I've made a new pull request.

PR #1659, opened 2023-01-25
Provide alternative search summary display. Closes #1652

### LINKED ISSUE #1652: Provide alternative search summary display
On visiting last Friday, @michaelzhang43 thought it would be helpful to have a search summary display that was sorted by case. It would be a condensed form of the displayed cases.

## Current Search Summary
image

## More info needed
@michaelzhang43 

* Do we want to have the heading to be clickable to toggle the displays? Ex:
image

image

* What fields do we want displayed? Something like this?

image

The top line has: case date, case number. Clicking on the case number scrolls down to the case details.

The changes can be seen here
https://monocle.github.io/recordexpungPDX/

Just navigate to the demo.

(Note some things, like the icons, are broken due to how I'm hosting it. But at least you can see and click things as a webpage.)

monocle:
@KentShikama ,
I'm submitting this now. If you're able to review and put it into the dev server, I can go to the CLEAR clinic on Friday and get some user feedback. Otherwise, we can wait longer for more discussion in case there are changes that should be made.

Either way, I think it'd be good to keep it in dev until I see users use it with real data.

---

This PR creates the `CasesList` component to provide an alternative view to `ChargesList` that is used in `RecordSummary`.

This PR includes refactorings for the following components:

* `RecordSearch`
* `RecordSummary`
* `ChargesList`
* `Demo` (can use demo data files in order to help visualize changes during dev with running the server)


src/frontend/src/components/Demo/index.tsx
```diff
@@ -1,42 +1,36 @@
...
-  componentDidMount() {
-    document.title = "Demo - RecordSponge";
-    store.dispatch(this.props.startDemo());
-  }
+export default function Demo() {
+  const dispatch = useAppDispatch();
+  let record = useAppSelector((state) => state.search.record);
+  // record = multipleCharges;
+  // record = johnCommonRecord;
+  // record = blankRecord;
```
KentShikama, line 19: Probably good to clean up

KentShikama (approved):
Sounds good!

src/frontend/src/components/Demo/index.tsx
monocle: Sorry, left that in there since I'm still not using the backend to provide record data during dev. I figure using Demo is a safe place to visualize changes in the browser. Not ideal. Will try to keep that out of the PRs.

Issue #1656, opened 2023-01-20
Create new Eligibility types: "Eligible but on case with [Ineligible/Future Eligible] charge"


michaelzhang43:
Currently all eligible charges are grouped together and classified as "Eligible." However, we want to distinguish between (i) eligible charges on cases that are completely eligible (and therefore the entire case is eligible) and (ii) eligible charges on cases where there is at least one non-eligible charge. 

We further want to distinguish between the types of non-eligibility, i.e.  eligible charges on cases where is
(i) at least one Ineligible charge
(ii) at least one Future Eligible charge
(if a case has both Ineligible and Future Eligible charges, then only state the Ineligible charge)

**Eligible on case with Ineligible charge**
**Eligible on case with charge eligible [date of latest eligibility]**

These should be listed in the Summary Panel after Eligible charges. **Eligible on case with charge eligible [date of latest eligibility]** should be listed in closest -> furthest date of eligibility
so that the order is

Possibly Eligible
Ineligible
Eligible
Eligible on case with Ineligible charge
Eligible on case with charge eligible [date (earliest)]
...
Eligible on case with charge eligible [date (latest)]
etc.

monocle:
#1298 seems to be related. That ticket was tagged as `Backend`, but if all of the information is already sent to the frontend via the `expungement_result` JSON data field, then the display logic can be handled without new backend code. 

Here's an example from one of the John Common charges:

<img width="802" alt="image" src="https://user-images.githubusercontent.com/120387970/214603863-238982a5-5bc2-4c50-a511-3cda801a80ca.png">

wittejm:
@michaelzhang43 a few questions:

1. Should the expanded view be affected? Since charges are grouped by case it's easy to see when an eligible charge is on a case with a future eligible or ineligible charge, so it seems redundant to say so.

2. Should the charge detail display be affected?

3. Should Generate Paperwork no longer generate the paperwork for these charges?

wittejm:
Another way to phrase the same question 3: do we generate paperwork for a case only if every charge on that case is eligible?

michaelzhang43:
1./2. Indeed would be redundant, but I'd like to have both the expanded view and the case detail panels reflect the new status - I did a couple "test runs" and found myself wanting to look at both of these to evaluate whether expunging those particular charges were worth it. Having the charges at issue in a different color than the eligible charges on other cases would help to distinguish them 

[...]

PR #1678, opened 2023-03-10
More ineligibility categories

wittejm:
this covers part of #1656 

This uses a demo record to test, which I expanded with two cases that provide the new ineligibility categories.
Screenshot 2023-03-09 at 12 33 57 PM


wittejm:
Ready for review @KentShikama . I was stuck for a while on a bug from `groupby` using a different key from the sort that's done on the list immediately prior. So it now reuses the same key function.

It's not clear to me what this sort is even for (the ordered numbers in the key function) since the grouped charges don't get rendered in the same order.

KentShikama (approved):
Feel free to handle the deploy once you confirm on dev.recordsponge.com that it's working

PR #1679, opened 2023-03-13
Finish refactoring FormFilling

### LINKED ISSUE #1626: Delete unused pdf forms and form-filling logic
all counties except multnomah now use the shared state form. douglas and umatilla still use "with arrest order" and "with conviction order" forms.

Paperwork and form-filling for all other counties can be removed.

### LINKED ISSUE #1661: Add Arrest Date information to all expungement forms
From Michael Zhang: I received word from our partners in Douglas County that the court there is requiring the Arrest Information to be filled out on ALL expungement paperwork. 

Currently, Arrest Information for a case is not filled out if there is at least one Conviction on the case (which is correct, if one actually reads the forms). However, the Douglas County isn't accepting the forms unless the Arrest information is also filled out. Adding this information on ALL forms should not present any issues for other counties since the information is extraneous but doesn't detract from the request to also expunge the Convictions. 

Here is the ask:
For ALL case paperwork, provide the Date(s) of Arrest in the "Date of arrest" line, and check the appropriate box corresponding to "no accusatory instrument" or "accusatory instrument was filed" - basically, the information that goes here (Arresting Agency doesn't need to be provided). 

Screen Shot 2023-01-27 at 5 40 36 PM

Thank you

### This task is done when:
 -
 -

monocle:
* Create new CaseResults and UserInfo dataclasses.
  - Remove intermediate dict mappings.

* AcroFormMapper
  - Rename to PDFFieldMapper.
  - Move direct mapping logic (a PDF field named "Full Name" gets mapped
to `full_name`) to here.
  - Allow multiple PDF fields to directly map to CaseResults/UserInfo.
Ex. if `full_name` is needed for a second field, "Full Name---" can
be used.

* PDF
  - Move PDF writing logic to here.
  - Returns a `shrunk_fields dict` of annotation keys and values instead
of generating specific text.

* FormFilling
  - Move all warning handling logic to `_generate_warnings_text`
  - Change PDF file naming convention so that PDFs with arrest or
conviction orders have "_arrest" or "_conviction" suffix.

* Update tests
  - Add John Common record to test `FormFilling.build_zip`.
  - Add tests for PDF and download path generation.
  - Add tests for warnings text generation.
  - Add tests to verify form fields are correct for
    * oregon.pdf
    * oregon_arrest.pdf
    * oregon_conviction.pdf
    * multnomah_arrest.pdf
    * multnomah_conviction.pdf

* Use all `arrest_dates` instead of `dismissed_arrest_dates`. Closes #1661.
* Remove unused PDFs. Closes #1626

* Add back time analyzer tests.


KentShikama (approved):
Thanks for the clean up here!

KentShikama:
@monocle Let me know when you've done a few checks on https://dev.recordsponge.com/ especially around the form filling.

monocle:
@KentShikama I found something with the font shrinking that's not working right. Will start digging into it.

PR #1674, opened 2023-02-23
Fill forms soft requirement

Matth5050:
Refactored UserDataForm.tsx, added EmptyFieldsModa.tsx for a modal pop-up when the user doesn't enter fields for expungement packer download, and updated FillForms.test.tsx

Issue #1644, opened 2023-01-14
Create soft requirments for FillForms


monocle:
<img width="1135" alt="image" src="https://user-images.githubusercontent.com/120387970/212484710-44644d89-f4d1-472f-b181-1dbb0c71852c.png">

Per yesterday's conversation, all the fields should have a soft requirement, but there should be a way that the user can still bypass filling the forms in rare situations. 

@michaelzhang43 Should the heading "User Information" be "Client Information"? It seems like our primary "user" is the clinic staff that helps the client.

@KentShikama Thoughts on this https://github.com/monocle/recordexpungPDX/issues/14

Matth5050:
Can someone let me know what a soft requirement is? Like making sure the user actually inputs something?

Matth5050:
I have two proposed methods for this and I would appreciate anyone's input. 

1. The first is a checkbox that the user would have to check in order to download without filling in the forms. 
2. The second is a popup that alerts them that they haven't filled out the form but that they can continue to download if they want.

Let me know which you'd like or if we should go in another direction is better!

![Screenshot 2023-02-06 at 20 13 03](https://user-images.githubusercontent.com/95751146/217155861-4a805ffc-e0c9-4974-8d92-4e5328409eed.png)
![Screenshot 2023-02-06 at 21 15 39](https://user-images.githubusercontent.com/95751146/217155869-965e599f-f6b5-413f-9c1b-b863d0c00c10.png)

michaelzhang43:
I like the second format better, but I think the first sentence should be phrased as "You are about to download an incomplete form." The forms even without the demographic info would still have the case and charge information, etc.

monocle:
@Matth5050 I like the modal as well. Nice work with that!

Matth5050:
@michaelzhang43 @monocle Thanks!

PR #1680, opened 2023-03-14
Fill forms requirements v2

Matth5050:
This branch is for [issue 1644](https://github.com/codeforpdx/recordexpungPDX/issues/1644) and is taking the place of my [past PR](https://github.com/codeforpdx/recordexpungPDX/pull/1674) for this issue. 

It solves the problems with the past PR and improves on most other areas as well.

Thanks to @monocle for the help!


src/frontend/src/components/FillForms/UserDataForm.tsx
```diff
@@ -1,337 +1,293 @@
...
+    fullName += first_name ? first_name + " " : "";
+    fullName += middle_name ? middle_name + " " : "";
+    fullName += last_name;
+
+    return fullName.includes("*") ? "" : fullName || " ";
+  }
+
+  function allInputsValid() {
+    const errorMessages = {
+      "Date Birth format must be mm/dd/yyyy": isPresent(dob) && !isDate(dob),
```
KentShikama, line 54: Is this format something that the courts have asked for? I could see someone wanting to just type out "Jan 1, 2023" - after all we aren't parsing this value but writing this out onto a form that would be read by a human.

KentShikama (commented):
Neat to see you pushing on this! Any chance you are willing to show a (silent) screen demo of this working? Particularly I'm interested to see what the error messages look like and how it looks like after the error is corrected. Forms are notoriously difficult to handle elegantly around this area.

src/frontend/src/components/FillForms/UserDataForm.tsx
Matth5050: I would love to demo but I'm not sure what a silent screen demo is. Is it screenshots or video? 

As far as the date format, that's how it was before I touched it and I'm not sure if that's the required format for the courts. Anytime I've ever dealt with dates I've always used something like <input type="date"> to force the user into a uniform format...

I'm happy to demo and/or change anything 😊

KentShikama: > Is it screenshots or video?
video

> that's how it was before I touched it and I'm not sure if that's the required format for the courts
Let's leave it out then. It gets auto-filled from the previous screen and the previous screen requires that format but that isn't for the courts.

Matth5050: alright I made the vid. I'm not sure how to attach it here so I put it on my google drive and I'll link it [here](https://drive.google.com/file/d/1t-OQYYK8NI__k8AJq-LATWypODducfK4/view?usp=sharing). I can always share it on discord/slack if that's better.

P.S. I noticed that the error messages were lingering even after a successful submit as well as the error message for the phone number being a bit odd. I'm going to push those two small changes now

KentShikama:
Thanks for the video. In the future you can probably just literally drag and drop it into Github in the PR description.

Ah I like how you put the errors below the button. It takes away some classic reactivity problems while not having to do more state management. Let's just remove the Date of Birth formatting since I could reasonably see someone wanting to just input something like "Jan 1, 2023" as I mentioned above.

Also for the zip code, I think we want to allow for any extensions like 12345-1234.

src/frontend/src/components/FillForms/UserDataForm.tsx
```diff
@@ -1,337 +1,295 @@
...
-      } else {
-        return "";
-      }
-    } else {
-      return "";
-    }
-  };
+const isPhoneNumber = (str: string) => {
+  //the phone RegEx accepts the following formats (123) 456-7890, 123-456-7890, 123.456.7890, 1234567890, (123)-456-7890
+  return /^(\(?\d{3}\)?[-.\s]?){2}\d{4}$/.test(str); // Check for empty string included
```
KentShikama, line 17: Oh does this account for international phone numbers?

Matth5050: No, just a standard 10 digit US phone number

KentShikama:
> I'm curious if the courts would even attempt to call an international phone number, they wouldn't be able to send texts either without some extra steps.
My guess is that they almost never call this number but I would ask Micheal for more details. Maybe a user wants to put down a VoIP phone number from a foreign country because that is cheaper to obtain, and they can't afford a standard US phone plan. I'm just afraid in this context to commit without knowing more because it sounds like we are _preventing_ the user from download if they don't get past this error.

> I'd like to know why the current birth format is questionable. Currently, the user can see the format before they type so I don't think there would be any confusion.

Remember, this is just to fill a form. We aren't actually parsing the birth date and doing logic on it at this stage. A lawyer might want to print the birth date in a certain format because they think that would look better with the DA or judge.

Matth5050:
@michaelzhang43 Hello Micheal! 

Can you weigh in on this form? Kent brought up a good point about the VoIP number. Kent also mentioned that a lawyer might want to fill in a form differently than our current requirements. If this is the case should we even have checks for formatting (ex. date of birth must be in mm/dd/yyyy)? We can remove any warnings and only have the modal pop up if a field is empty. 

Thanks!

michaelzhang43:
I love it overall! Notes on specifics:
- Zip code starting with 5 digits sounds good; no one ever puts more than that but it's fine
- ten digit phone number is good; no need to account for international phone/VoIP
- No need to request formatting for Date of Birth
- In general, important to preserve the function of being able to fill out an incomplete form (only popping up the "are you sure" banner with clickthrough works)

Matth5050:
@michaelzhang43 I appreciate the input!

Matth5050:
@KentShikama I have removed the date of birth requirements and altered the phone number error message to say "Phone number must be a valid 10 digit number"

Let me know if any further changes should be made

src/frontend/src/components/FillForms/UserDataForm.tsx
```diff
@@ -1,337 +1,294 @@
...
-            moment(this.state.dob, "M/D/YYYY", true).isValid() === false,
-        },
+  function buildName() {
+    const { first_name, middle_name, last_name } = aliases[0];
+    let fullName = "";
+    fullName += first_name ? first_name + " " : "";
+    fullName += middle_name ? middle_name + " " : "";
+    fullName += last_name;
+
+    return fullName.includes("*") ? "" : fullName || " ";
```
KentShikama, line 49:
```suggestion
    return fullName.includes("*") ? "" : fullName;
```

src/frontend/src/components/FillForms/UserDataForm.tsx
```diff
@@ -1,337 +1,294 @@
...
-  mailingAddress: string;
-  phoneNumber: string;
-  city: string;
-  state: string;
-  zipCode: string;
-  invalidZipCode: boolean;
-  invalidPhone: boolean;
-  invalidBirthDate: boolean;
-}
+const errorMessagesObjectToArray = (obj: object) => objectToArray(obj);
```
KentShikama, line 13: This is more of a preference but `objectToArray` is currently only used once in this entire project in this file, so I would rather just keep the function in here. In the future, when it is used again, we can extract it as you did. It is a tough judgement call but I find that if I "prematurely" extract it whenever I can, it ends up never being used again half of the time, making it just harder to read/review half of the time. Of course if it is really generic then it makes sense but this `objectToArray` is only for specific types of objects.

src/frontend/src/components/FillForms/UserDataForm.tsx
```diff
@@ -1,337 +1,294 @@
...
-    });
-  };
 
-  public componentDidMount() {
-    if (!(this.props.aliases.length > 0)) {
-      history.push("/record-search");
-    }
+    if (!allInputsValid()) return;
+
+    setErrorMessages([]);
```
KentShikama, line 85: I see that `setErrorMessages` is called in `allInputsValid`. Are you thinking we call it again here just to make it explicit that we are setting it to `[]`?

src/frontend/src/components/FillForms/UserDataForm.tsx
```diff
@@ -1,337 +1,294 @@
...
+        isPresent(phoneNumber) && !isPhoneNumber(phoneNumber),
+    };
 
-        resolve
-      );
-    });
-  };
+    if (errorMessagesObjectToArray(errorMessages).length === 0) return true;
+
+    setErrorMessages(errorMessagesObjectToArray(errorMessages));
```
KentShikama, line 62: The function name `allInputsValid` suggests to me that there are no "side-effects" in the function - just a simple evaluation of variables which resolve to true/false. I would prefer a name that suggests the side-effect. Or figure out how to move the side-effect out (while still keeping things simple).

KentShikama (commented):
Thanks for the product updates! Just a few code nits.

src/frontend/src/components/FillForms/UserDataForm.tsx
Matth5050: The `setErrorMessages([]);` on line 85 is to clear the error messages after successful submission. Without it, error messages persist even if you successfully download the packet.

src/frontend/src/components/FillForms/UserDataForm.tsx
Matth5050: would something like `inputsMeetCriteria` work?

src/frontend/src/components/FillForms/UserDataForm.tsx
Matth5050: ok I changed it back

src/frontend/src/components/FillForms/UserDataForm.tsx
KentShikama: Hmm I don't think that implies a side-effect. Maybe inputsMeetCriteriaAndSetErrors . Kind of a mouthful but at least it is clear.

Matth5050: I have updated that function name along with completing the other recommendations you made above

KentShikama (approved):
Thanks for all the followup!

PR #1681, opened 2023-03-14
Increase heights of PDF fields

monocle:
1. I noticed yesterday in the Oregon conviction order, that the text was being cut off without being shrunk, so there was no warning being generated. Unfortunately, the different PDF renderers seem to be inconsistent on how many characters can fit in a given field width. 

So instead of changing the shrinking function, for the Oregon conviction and arrest orders, I have increased the field heights for the charge name and date fields and made those fields multi-line so that they will wrap and at least two lines of text can fit.

2. Small fix to not generate commas in the address field if there are no field values.


src/backend/expungeservice/form_filling.py
```diff
@@ -61,6 +61,13 @@
     "yamhill": "District Attorney - 535 NE 5th St #117 - McMinnville, OR 97128",
 }
 
+T = TypeVar("T")
+
+
+def join_dates_or_strings(arr=List[T], connector="; ", date_format="%b %-d, %Y") -> str:
+    get_attr = lambda elem: elem.strftime(date_format) if isinstance(elem, DateWithFuture) else elem
```
KentShikama, line 68: I prefer to use nested functions over lambda's for things that you are naming (i.e., get_attr). And also not use the name `get_attr` even though this is really local in scope as `get_attr` as it feels too close to `getattr` that I have to think for a second whether or not it is intentional or just a mistake.

monocle: Yes, let me clean that up. I clobbered some built in functions too that I should rename.

monocle: Should be good to go now.

src/backend/expungeservice/form_filling.py
```diff
@@ -61,6 +61,15 @@
     "yamhill": "District Attorney - 535 NE 5th St #117 - McMinnville, OR 97128",
 }
 
+T = TypeVar("T")
+
+
+def join_dates_or_strings(arr=List[T], connector="; ", date_format="%b %-d, %Y") -> str:
```
KentShikama, line 67:
```suggestion
def join_dates_or_strings(arr: List[T], connector="; ", date_format="%b %-d, %Y") -> str:
```

Are you sure you don't mean this? Is it a default value of List[T]?

KentShikama: Btw are you aware of https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments? Probably not relevant here but was reminded of it.

KentShikama: Ok on a closer look, I actually think not having default arguments is the way to go here. Just pass it in. Explicit is better than implicit and since you only use it twice anyways, I don't think the default arguments are worth it - yet.
```
def join_dates_or_strings(arr, connector, date_format) -> str:
```

monocle: Good catch. 👍🏽  Definitely unintended. I'm surprised the checkers didn't flag that.

I just fixed that and removed the defaults. Also, I replaced the generic type with something more specific since the function isn't that generic.

Really appreciate the time you take on the code review. I'm going to grab a copy of that book. Looks like good stuff.

KentShikama (approved):

PR #1685, opened 2023-03-21
Create Clackamas specific PDF. Closes  #1665

### LINKED ISSUE #1665: Clackamas County
After a Motion to expungement a case is accepted, the court signs an Order that has the expungement take legal effect. Clackamas County is apparently requiring people to provide their own Orders using the template below. 

For all Clackamas County cases (and only Clackamas County cases), include as part of each Motion paperwork, a corresponding filled out Order. 

Only these slots need to be filled out:
- Replace NAME with Defendant Name as it appears in the name of the case. 
- Full Name
- Date of Birth
- SID Number
- Case No
- Charge(s)
- Arrest Date
- Disposition

In particular, Agency and FBI number do not need to be filled out. SID Number is OK if not filled out

[Expungement Order Template.docx](https://github.com/codeforpdx/recordexpungPDX/files/10677030/Expungement.Order.Template.docx)

monocle:
This creates a new PDF that contains the oregon.pdf plus the
Clackamas "ORDER SETTING ASIDE CONVICTION AND SEAL RECORD OF ARREST"
form.

* The charges being used in the order part are all eligible charges.
* Prints the first two charges in the form boxes.
* Prints the remaining charges on page 2 as lines of strings in the
whitespace.


KentShikama (approved):

KentShikama:
@monocle Tell me when you think it's good to release to production.

PR #1686, opened 2023-03-22
Add logout button - closes #1653

### LINKED ISSUE #1653: Create feature to log out of OECI credentials
Users must input OECI credentials in order to access the search tool. However, there is no feature to "log out" of these credentials, so accessing the webpage again with cookies enabled will allow the user of the same machine to log back in without inputting OECI credentials. This creates an issue when users use RecordSponge on public computers and the only way to remove OECI credentials is to clear cookies. 

As such, we need a button to log out of OECI credentials.

trezp:
Closes #1653 

- Adds log out button to header
- Conditionally renders based on "logged in" status
- Empties records search form on log out

### DIFF
--- a/src/frontend/src/components/Header/index.tsx
+++ b/src/frontend/src/components/Header/index.tsx
@@ -1,36 +1,51 @@
-import React from "react";
-import { Link } from "react-router-dom";
+import React, {useState, useEffect} from "react";
+import { Link, useLocation } from "react-router-dom";
 import Logo from "../Logo";
+import { isLoggedIn, oeciLogout } from "../../service/cookie-service";
 
-export default class Header extends React.Component {
-  public render() {
-    return (
-      <div className="bg-white shadow">
-        <nav 
-          className="mw8 relative center flex flex-wrap justify-between pa3"
-          aria-label="Primary"
-        >
-          <div className="logo mb4 mb0-ns">
-            <Link to="/" aria-label="Home">
-              <Logo />
-            </Link>
-          </div>
-          <div className="mt5 mt2-ns">
-            <Link
-              to="/manual"
-              className="link hover-blue f5 fw6 pv2 ph0 ph3-ns mr4-ns"
+
+export default function Header (){
+  const [loggedIn, setLoggedIn] = useState(isLoggedIn());
+  const location = useLocation();
+
+  useEffect(() => {
+    setLoggedIn(isLoggedIn());
+  }, [location]);
+  
+  return (
+    <div className="bg-white shadow">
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/Header/index.tsx
```diff
@@ -1,36 +1,51 @@
...
-              to="/manual"
-              className="link hover-blue f5 fw6 pv2 ph0 ph3-ns mr4-ns"
-            >
-              Manual
-            </Link>
-            <Link
-              to="/record-search"
-              className="absolute top-1 right-1 static-ns bg-blue white bg-animate hover-bg-dark-blue f5 fw6 br2 pv2 ph3"
+
+export default function Header (){
```
KentShikama, line 7: Could we please give the prettier code formatter a run here? It should reformat this file.

src/frontend/src/components/Header/index.tsx
```diff
@@ -1,36 +1,51 @@
...
+  const [loggedIn, setLoggedIn] = useState(isLoggedIn());
+  const location = useLocation();
+
+  useEffect(() => {
+    setLoggedIn(isLoggedIn());
+  }, [location]);
+
```
KentShikama, line 14: Any chance we could refactor this to use something like https://www.npmjs.com/package/react-cookie? It feels a bit odd to hook into location for this kind of thing.

KentShikama (approved):
Thanks for giving this another shot!

KentShikama (changes requested):
Pardon. Didn't mean to approve.

src/frontend/src/components/Header/index.tsx
trezp: @KentShikama Can you tell me more about why this bothers you and what you would prefer to be done with this package?

trezp: @KentShikama react-cookie wasn't going to work because it doesn't trigger a re-render when the cookies change, but I pushed another solution. Looks like I need to update so front end tests are passing though. Will re-request a review once I button that up.

KentShikama: @trezp Pardon the delay in reply. It is just that the use of the `location` is "indirect". Ideally, we would be reacting to when the cookie is added/removed/expired as that is what the status of the login button should reflect.

KentShikama: @trezp 

> react-cookie wasn't going to work because it doesn't trigger a re-render when the cookies change
Ah ok. Pardon for taking you down that path then.

PR #1687, opened 2023-03-24
Update multnomah_conviction.pdf

monocle:
There was a strange bug where Chrome would be missing almost all the values from the Multnomah conviction PDF. The form was properly filled out when viewed in Apple Preview, Adobe Reader and Firefox.

After regenerating the Case Number field in the multnomah_conviction.pdf, all fields seem to be displaying in Chrome.

Otherwise, the new Clackamas PDF seemed to be working as intended.


KentShikama (approved):
Yup I've faced this before too. For some reason order matters.

PR #1692, opened 2023-04-21
Fixes for Eligible charge on case with ineligible charge

wittejm:
A few change requests from users:

1. Paperwork should be generated if this is the only type of eligible charge
2. The header for this type in the summary should be green


KentShikama (dismissed):

wittejm:
redo because of required rebase.

KentShikama (approved):

PR #1697, opened 2023-05-03
Generate forms won't get an empty array error

wittejm:


monocle (approved):
Pulled this locally and it's working for me.

PR #1699, opened 2023-05-10
Remove clackamas form

wittejm:
First commit is a rather timid commenting out the part that calls separate logic for clackamas. Second commit I'm actually deleting the code and counting on us digging into commit history if we ever need it again.


src/backend/expungeservice/form_filling.py
```diff
@@ -550,72 +545,9 @@ def validate_initial_state(self):
...
-        else:  # List
-            font_size = self.FONT_SIZE_MEDIUM
-
-        annotation.DA = PdfString.encode(f"/{self.FONT_FAMILY} {font_size} Tf 0 g")
-
-
 class FormFilling:
     OREGON_PDF_NAME = "oregon"
-    NON_OREGON_PDF_COUNTIES = ["multnomah", "clackamas"]
+    NON_OREGON_PDF_COUNTIES = ["multnomah"]
```
KentShikama, line 550: This is an interesting name now that I think about it.

src/backend/expungeservice/form_filling.py
```diff
@@ -447,13 +447,8 @@ class PDF:
...
     @classmethod
     def fill_form(cls, mapper: PDFFieldMapper, should_validate=False):
-        klass = cls
-        county = mapper.get("(County)")
 
-        if county and county.lower() == "clackamas":
-            klass = ClackamasPDF
-
-        pdf = klass(mapper)
+        pdf = cls(mapper)
```
KentShikama, line 451: Can this just be 
```
pdf = PDF(mapper)
```
now with a `@staticmethod`?

KentShikama (approved):

KentShikama (approved):

Issue #1711, opened 2023-08-31
Group Owe$ charges with Future Eligible charges on same case


michaelzhang43:
**Background.** 
Previously we made a change to the Summary Panel so that for a case that has (i) Future Eligible convictions and (ii) charges that are Eligible Now, we group the Eligible Now charges with the Future Eligible convictions. We also made it such that these charges in (ii) above don't get printed. This is for practical purposes - generally we don't want our users to file Partial Expungements (i.e. where only some of the charges on a case are eligible), if they will be filing a further expungement on the same case later. 
This has been a generally popular change. However, it has only been implemented for cases/charges where there is no money owed. 

**Request**
I would like for this behavior to be consistent for charges on cases where there is also money owed (Note that money owed is a characteristic of a case, not a charge.) That is, in a case with money owed, and (i) Future Eligible convictions and (ii) charges that are Eligible Now, change the Summary Panel results so that the charges under (ii) are grouped under (i) - they should all be listed as "Future Eligible but Owes $"

PR #1712, opened 2023-09-07
Label Refactor

JeremyC3:
Refactored label grouping algorithm as per #1711. As a result, one backend test case was changed to list charge grouping as "Eligible NOW If Balance Paid on case with Ineligible charge".

Changed grouping algorithm to sort by label and case number only, removed secondary date sort to ensure that charges with the same case number are properly grouped. Will re-introduce if users find new sorting less useful.

### DIFF
--- a/src/backend/expungeservice/charges_summarizer.py
+++ b/src/backend/expungeservice/charges_summarizer.py
@@ -5,7 +5,6 @@ from expungeservice.models.record import Record
 from expungeservice.models.record_summary import ChargesForSummaryPanel
 from expungeservice.util import DateWithFuture as date
 from typing import Dict, List, Tuple
-from enum import Enum
 
 class ChargesSummarizer:
     @staticmethod
--- a/src/backend/expungeservice/charges_summarizer.py
+++ b/src/backend/expungeservice/charges_summarizer.py
@@ -38,49 +38,53 @@ class ChargesSummarizer:
     def _primary_sort(charge: Charge, record: Record):
         charge_eligibility = charge.expungement_result.charge_eligibility
 
-        class ChargeList(str, Enum):
-            NEEDS_MORE_ANALYSIS = 0
-            INELIGIBLE_NOW = 1
-            NOWELIGIBLE_NOCHARGE_NOINELIGIBLE = 2
-            NOWELIGIBLE_NOCHARGE_HASINELIGIBLE = 3
-            NOWELIGIBLE_HASCHARGE_NOINELIGIBLE = 4
-            NOWELIGIBLE_HASCHARGE_HASINELIGIBLE = 5
-            FUTUREELIGIBLE_NOCHARGE_NOINELIGIBLE = 6
-            FUTUREELIGIBLE_HASCHARGE_NOINELIGIBLE = 7
-            FUTUREELIGIBLE_NOCHARGE_HASINELIGIBLE = 8
-            FUTUREELIGIBLE_HASCHARGE_HASINELIGIBLE = 9
+        '''
+            Order is:
+            0 Needs More Analysis
+            1 Ineligible
+            2 Eligible Now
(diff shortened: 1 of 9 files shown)


src/backend/expungeservice/charges_summarizer.py
```diff
@@ -38,6 +38,8 @@ def build_charges_for_summary_panel(record: Record) -> ChargesForSummaryPanel:
     @staticmethod
     def _primary_sort(charge: Charge, record: Record):
         charge_eligibility = charge.expungement_result.charge_eligibility
+
+        charge_dict = {"nma":0,"ine":1,"nnn":2, "nni":3, "ncn":4, "nci":5, "fnn":6, "fcn":7, "fni":8, "fci":9 }
```
KentShikama, line 42: Could we use an str, Enum and spell these out?

src/backend/expungeservice/charges_summarizer.py
```diff
@@ -15,10 +15,10 @@ def build_charges_for_summary_panel(record: Record) -> ChargesForSummaryPanel:
         ]
         eligible_charges_by_date: List[Tuple[str, List[Tuple[str, List[Tuple[str, str]]]]]] = []
         sorted_charges = sorted(
-            sorted(visible_charges, key=ChargesSummarizer._secondary_sort, reverse=True),
+            visible_charges,
+            #sorted(visible_charges, key=ChargesSummarizer._secondary_sort, reverse=True),
```
KentShikama, line 19: I prefer to delete all comments as the git history keeps them anyways

src/backend/expungeservice/charges_summarizer.py
JeremyC3: I can spell out the different charges, but could you elaborate a bit on the benefits of enum vs. a regular dict for this? Thanks!

JeremyC3:
@KentShikama changes made to utilize an str,Enum with written-out classifications instead of a dict. Removed pre-existing comment about order now each classification is written out. Manually cast Enum back into an int on line 83 due to typecasting errors occurring when using IntEnum

KentShikama (commented):
You should be able to do something like
```
ChargeList.NEEDS_MORE_ANALYSIS
```

It effectively just acts as a labelled integer. I realized since you just want an integer here you can just use `class Foo(Enum):`

src/backend/expungeservice/charges_summarizer.py
KentShikama: It isn't that different - just a bit more robust. I also just prefer not to abbreviate. For example, I don't know what `fni` means.

KentShikama (commented):
Ignore what I said earlier. I finally understand what you were trying to do with the +=. Basically a clever way to append on a prefix. I would prefer if we just built up boolean variables and then had a function at the end instead of `int(ChargeList[classification])`.

```
def function(foo, bar, baz):
    if foo and bar and baz:
        return 7
    if foo and bar and not baz:
        return 8
```

JeremyC3:
Reverted the sorting logic back to if statements, changed `no_balance` label to `has_balance` to match the other variable `case_has_ineligible_charge`

src/backend/expungeservice/charges_summarizer.py
```diff
@@ -5,7 +5,7 @@
 from expungeservice.models.record_summary import ChargesForSummaryPanel
 from expungeservice.util import DateWithFuture as date
 from typing import Dict, List, Tuple
-
+from enum import Enum
```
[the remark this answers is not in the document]
JeremyC3: Nope, thanks for the catch! removing now.

KentShikama (approved):
Nice!

KentShikama:
You know about https://dev.recordsponge.com/ right?

PR #1714, opened 2023-11-09
bugfix: section header elif case 

wittejm:
The search was crashing on a record that contained a future eligible case with an ineligible charge. Apparently this hasn't shown up (or been reported, at least) since I updated this section header stuff in the spring. Fixed by removing the typo `not` to cover the missing case.

I also added an `else: return` after all the `elif` cases so that if there's another bug it won't crash the search with a 500, it'll just print the unsorted charge in a section with a blank section header.

### DIFF
--- a/src/backend/expungeservice/charges_summarizer.py
+++ b/src/backend/expungeservice/charges_summarizer.py
@@ -1,5 +1,4 @@
 from itertools import groupby
-
 from expungeservice.models.charge import Charge, EditStatus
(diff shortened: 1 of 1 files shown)


KentShikama (approved):
I'm guessing it came from the recent refactor here: https://github.com/codeforpdx/recordexpungPDX/pull/1712.

wittejm:
Oh interesting. Yep.

PR #1723, opened 2024-04-23
fix pdfs with bugged date of arrest field

wittejm:
I used Acrobat to delete and replace the fillable field with the same name and now it works. Thanks for the guidance!


wittejm:
I'm keeping the old pdfs here for now just in case there are issues with the new ones. they seem to work fine but I wanted Michael to check over it too.

KentShikama (approved):
Yeah I wish the PDF format was easier to parse.

PR #1727, opened 2024-07-22
Move disclaimer

emko-274:


wittejm:
btw @emko-274 You can run tests locally with `make test`, so long as the docker stack is running.

emko-274:
Oh, right. Thank you :)

PR #1733, opened 2024-07-26
changes to downloadables and relevant tests

emko-274:
- moved disclaimers to end of summary PDF
- added balances summary to end of summary PDF
- updated tests to reflect changes in PDF file
- updated Needs More Analysis sex crime prompt
- updated tests to reflect change in Needs More Analysis sex crime prompt
- removed county autofill in OSP form
- updated test to reflect removal of county autofill from OSP form

PR #1734, opened 2024-07-26
Remove autofill 1

emko-274:
Removed OSP court autofill and updated tests to reflect changes. This was erroneously listed as a fix in the description of pull request #1733.

### DIFF
--- a/src/backend/expungeservice/form_filling.py
+++ b/src/backend/expungeservice/form_filling.py
@@ -396,7 +396,8 @@ class PDFFieldMapper(UserDict):
             # Is the OSP Form
             osp_fields: Dict[str, object] = {}
             for i in range(10):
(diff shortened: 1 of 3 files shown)


src/backend/expungeservice/form_filling.py
```diff
@@ -705,4 +706,5 @@ def counties_with_cases_to_expunge(all_case_results: List[CaseResults]):
             if case_result.case.summary.balance_due_in_cents != 0:
                 counties_with_balances.append(case_result.case.summary.location)
         counties_with_expungements = [county for county in counties_with_eligible_charge if county not in counties_with_balances]
-        return counties_with_expungements
+        return []
```
wittejm, line 709: Not a big deal since the result of this function isn't used anymore, but I think this change in particular is confusing since it breaks the functionality. I would either leave it alone or comment out the function.

wittejm (approved):
needs a few changes, but after that looks good to merge.

PR #1754, opened 2024-08-23
Fix compiled

emko-274:


ekong5 (approved):

wittejm:
What is this pdf.css file for? what is the rest of this PR doing?

emko-274:
The pdf.css file isn't doing anything and shouldn't be included. I didn't realize it was still there. I'll remove it. 

Previously, COMPILED.pdf contained duplicates or missed cases. The rest of the PR is just compiling the motion to set aside files and OSP forms differently than it did before to include the correct forms. I also realize now that I forgot to remove the get_pages method from the PDF class which I wrote but didn't end up using.

PR #1755, opened 2024-08-23
Fixed COMPILED.pdf and updated PDF layout

emko-274:
- added 50px margin to PDF forms
- moved "Assumptions" block under "Disclaimer" in summary PDF

to-do: 
- update build_zip form filling tests
- add clear button test to search panel test file


wittejm:
Why did you have to modify the stylesheet?

emko-274:
Michael requested that we add a bigger margin to the PDF files, which I did by modifying the stylesheet.

wittejm:
And you're customizing the stylesheet that we were downloading from
somewhere?

wittejm:
It was surprising to see such a large new file. I think that comments in
the file mentioning where it came from, and noting the change you made from
the original, would be helpful

emko-274:
Yes, the new file just contains the text from the source that was originally being requested from (https://raw.githubusercontent.com/sindresorhus/github-markdown-css/gh-pages/github-markdown.css), with a one-line edit. Is it alright if I add the comment in the file where the new one is read from? I'm not sure how to write comments in the txt file so that they aren't read into the string that's supposed to contain only CSS, other than by skipping lines when reading, which I think might be awkward.

wittejm:
Ah, that's why it's a txt file. Yea, explaining it and showing the URL in
the comment when you call the file sounds good

emko-274:
Okay, I added the comment

wittejm (approved):

